PR #12, opened 2018-08-24
No longer need dbus_launch

AHumanFromCA:
With a little jenkins knowhow and some pretty impressive docker image modifications, I have been able to replace the dbus_launch shell in the integration test phase with:

```groovy
sh "echo 'jenkins' | gnome-keyring-daemon --unlock"
sh "npm run test:integration"
```

**NOTE:** In testing I have seen that the `gnome-keyring-daemon --unlock` is not needed in every pipeline. I am still trying to determine why that is the case and how I can get that working in this pipeline.

For example, If this pipeline had run with the image `mikebauerca/docker-brightside` and I tried to create a bright profile, it would work without the need for unlocking. In our pipeline, however, a similar operation that happens in one of our test clis that are installed during pipeline execution will yield:

```
** Message: Remote error from secret service: org.freedesktop.DBus.Error.UnknownMethod: No such interface 'org.freedesktop.Secret.Collection' on object at path /org/freedesktop/secrets/collection/login
Command Error:
Unable to store the secure field \"username\" associated with the profile \"profile-name\" of type \"username-password\".
Error Details:
No such interface 'org.freedesktop.Secret.Collection' on object at path /org/freedesktop/secrets/collection/login
```
without the unlock line.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -410,20 +410,22 @@ pipeline {
                 JEST_JUNIT_TITLE="{title}"
                 JEST_HTML_REPORTER_OUTPUT_PATH = "${INTEGRATION_RESULTS}/index.html"
                 JEST_HTML_REPORTER_PAGE_TITLE = "${BRANCH_NAME} - Integration Test"
-                TEST_SCRIPT = "./jenkins/integration_tests.sh"
             }
             steps {
                 timeout(time: 30, unit: 'MINUTES') {
                     echo 'Integration Test'
 
                     /**************************************************************************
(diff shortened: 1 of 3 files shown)


AHumanFromCA:
I think this PR is documentation enough. We should be able to see this first if a search is done on dbus. I can just link what was done to remove the need for dbus launch. See these 2 commits for specific details.
- https://github.com/AHumanFromCA/jenkins-npm-keytar/commit/07ef4e1d1e1d1151d9262c7489ea26e5fbc01704
- https://github.com/AHumanFromCA/jenkins-npm-keytar/commit/26d23a8e22b460f3a23554227294a7385c23c3ab

PR #29, opened 2018-09-27
Enhancement/21

### LINKED ISSUE #21: Allow plugins to use the OverridesLoader
Plugins to a base cli should be able to modify certain aspects of the CLI. They should make use of the current overrides loader so that not much custom work has to be done. Plugins that perform such an action will be referred to as **Infrastructure Plugins**

For now we will be talking about the CredentialManager section as that is the only overrides available right now.

## Install Behavior
- Installing an infrastructure plugin will modify a **settings.json** file (to be discussed later)
    - Any overrides provided by the plugin will now be saved in this file, overwriting any previous settings that were there before.
- Any migration that must occur due to the new plugin will not happen automatically
- Overrides are provided by the plugin in the imperative config overrides object.

## Settings file
- A settings.json file will be needed to reliably keep track of which infrastructure plugin to use for each overrides.
    - This allows the consumer to choose which plugin gets used, easily, in the field.
    - We are preparing for a possible config group later
- On each run of the cli this file is loaded so that we can determine what settings are available
    - If it does not exist, it is created with all the defaults specified.
- It will be stored in the CLI home directory

### Proposed Layout of File

Since we are expecting this file to be used for settings in the future and we want to handle any new overrides, I am proposing the following initial file structure.

```TypeScript
export interface ISettingsFile {
  overrides: {
    /**
     * If null, use the default provided by the CLI
     * If string, use the option provided by the plugin named in the string
     */
    [K in keyof IImperativeOverrides]: null | string
  };
}
```

So for now with no plugins the generated file will look like this:

```JSON
{
  "overrides": {
    "CredentialManager": null
  }
}
```

This allows for future overrides to be created as needed without any changes to the load/save actions.

## Plugin Load Behavior
- Infrastructure Plugins and Command Plugins (plugins that add a command group) will be the managed by the PMF.
- The PMF is now the first thing that gets control in `Imperative.init`
    - Now divided into 2 parts: Loading and Initializing
    - Loading will loop through all plugins in plugin.json.
        - Commands/Groups/Profiles are saved in memory for later injection
        - Imperative config object is returned with the new overrides provided by any plugin. The facility will look in settings.json to determine which one will be used for each overrides.
    - Initializing takes the preloaded Commands/Groups/Profiles and defines them at their current spot in Imperative
- After plugins are loaded, the config object will contain the correct overrides object. This ties nicely into the current operations of the overrides loader. I don't think much will need to change with that class.

See design docs for the before and after in the next comment:

AHumanFromCA:
## Changelog

- **[Minor]** Plugins can now override any overridable feature in a CLI. Currently the only overridable field is the CredentialManager (Closes #21)
- **[Minor]** Logger will now log items to memory before being initialized. This allows for aspects of imperative required for logging to be initialized first. If the CLI crashes before logging is initialized, then all messages will be dumped to the console.
- **[Minor]** A settings manager for imperative has been added to help out with overrides.
- **[Patch]** File logs now include millisecond times, this will be useful for benchmark tests in the future.

### Overriding a Credential Manager

To override the default CredentialManager with a plug-in, you perform the following steps.

1. A plug-in provides a TypeScript module that adheres to the following guidelines:

    * It provides a class that extends Imperative's [AbstractCredentialManager](https://github.com/zowe/imperative/blob/d70fc215ed8d28c73752b5a4807166ea842c740f/packages/security/src/abstract/AbstractCredentialManager.ts).

    * Details can be found in the link above, but as an overview, your class will implement the following methods.

        * initialize - An optional method to perform initialization after the class is constructed.
        * save - Save credentials for an account managed by the credential manager.
        * load - Load credentials for an account managed by the credential manager.
        * delete -  Delete credentials for an account managed by the credential manager.

    Within these methods, the plug-in will implement a means to securely encrypt, and decrypt credentials. The plugin will also devise a scheme for identifying which credentials are being accessed and where the credentials are stored. None of these decisions are dictated by Imperative. The plug-in can implement any scheme that meets the requirements of your organization.

2. Within the `overrides.CredentialManager` property of your plug-in's Imperative configuration, specify the location of the module that implements your CredentialManager override class.

    * Snippet of a plugin's imperative configuration
    ```javascript
        {
           ...,
            "overrides": {
                "CredentialManager": "./overrides/CredentialManager.override"
            },
           ...
        }
    ```

    * The string value of the `overrides.CredentialManager` property can either be a relative path or an absolute path (which enables the use of `__dirname`).

        If a relative path is used, it must be relative to your plug-in's 'require' entry point. For the example above,
        - Assume that `/YourProject` is the root of your project
        - `/YourProject/lib/index.js` is the compiled entry point of your project.
        - `/YourProject/lib/overrides/CredentialManager.override.js` is the compiled location of the CredentialManager implemented by the plug-in

3. Specify the name of which plug-in Imperative should use as its CredentialManager.

   The mere existence of a plug-in which contains an override is insufficient for Imperative to know whether that plug-in should become the CredentialManager override for your system. For example, you may want to use Imperative's default CredentialManager. Alternatively, you may have installed multiple plug-ins that can override Imperative's default CredentialManager.

   You select which plug-in should become the CredentialManager, by specifying the plug-in's name in a settings file named `$ZOWE_CLI_HOME/settings/imperative.json`.

   The settings file currently only contains an `overrides` property, which currently only contains a `CredentialManager` property. The value of the `CredentialManager` property is the name of the plug-in that will be used as the CredentialManager on your system.

```javascript
    {
        "overrides": {
            "CredentialManager": "@brightside/zowe-cli-sample-plugin"
        },
    }
```

When the value of the `overrides.CredentialManager` property in `$ZOWE_CLI_HOME/settings/imperative.json` is `false`, Imperative's default Credential Manager will be used.

Despite their similar names, the `overrides.CredentialManager` property in `$ZOWE_CLI_HOME/settings/imperative.json` is different than the property in the plug-in's Imperative configuration.

| Configuration Object                          | Property                        | Meaning                                                             |
| --------------------------------------------- | ------------------------------- | ------------------------------------------------------------------- |
| Plug-in's Imperative <br>configuration object | overrides.<br>CredentialManager | Path to a program that implements a CredentialManager               |
| $ZOWE_CLI_HOME/settings/ <br>imperative.json  | same property                   | Name of the plug-in whose CredentialManager should be used on this system |

When Imperative encounters an error while trying to override the default CredentialManager with one from a plug-in, Imperative will instead use a special `InvalidCredentialManager`. The `InvalidCredentialManager` will purposely fail and display an error on the console each time the application attempts to access secured credentials, until you reconfigure `$ZOWE_CLI_HOME/settings/imperative.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@brightside/imperative",
-  "version": "1.1.0-next.201809051706",
+  "version": "1.1.0-next.201809281807",
   "description": "framework for building configurable CLIs",
   "license": "EPL 2.0",
   "repository": "https://github.com/gizafoundation/imperative",


AHumanFromCA:
Any particular reason this was closed?

gejohnston:
I did not explicitly close it. After a disastrous merge of master into 21, push to remote, and attempts to fix it, maybe some git server logic closed it? Until proper merges of master in the 21 branch are complete we should not merge 21 to master.

PR #35, opened 2018-10-04
Enhancement/28

tucker01:
- Removes the requirement for Keytar in a CLI.
- Keytar is moved to a dev dependency in imperative.
- A CLI implementation can enable securing credentials by including keytar in their dependencies. 
- Secured profile fields in the yaml file will read "managed by <secure implementation>" if credentials are managed by keytar or a plugin


tucker01:
I don't think that line should have effect on credential managers and profiles. It's all done via the CliProfileManager and OverridesLoader/CredentialManagerFactory... I refactored some code to set the managed by name and I must have missed some stuff that was done previously.

Thanks for pointing this out. Let me take a peek and see what's what.

tucker01:
Sorry for the confusion here. I removed those changes. I should have looked more closely at what was already done... I offered to complete the changes when the team member working on it was asked to help elsewhere. 

I refactored where the display name gets supplied (to decouple some internal packages). 

@ChrisBoehmCA is making a PR to update the sample zowe-cli plugin with the changes required.

Issue #34, opened 2018-10-04
Implement "search order" precedence for CLI options ("--", ENV var, profile, default)


tucker01:
Imperative should support a "search order" for option/argument values that includes:

- Values specified on CLI options
- Environment variables
- Profiles
- Default values

# Tasks
 - [x] Create utility to take an option or profile key and return camel case and dash version
Status: @AHumanFromCA
Assigned: None
  - [x] Pass envVariablePrefix into CommandProcessor constructor
Status: Complete
Assigned: @gejohnston 
  - [x] Update command processor to perform "search order" for arguments - just create a stub for ENV var (for now)
Status: Not Started
Assigned: @tucker01 and @ChrisBoehmCA
 - [x]  (Optional) Implement function to get environment value for an option name
Status: Complete
Assigned: @gejohnston 
 - [x]  Write integration test - update/add an imperative test CLI to have a command with an option profile
Status: Complete
Assigned: @ChrisBoehmCA 
 - [x]  Document in Imperative Wiki
Status: Not Started
Assigned: None
 - [x]  Manually validate that existing plugins & CLIs don't break - zowe-cli, sample plugin, CICS, DB2, Endevor, FileMaster
Status: In Progress
Assigned: @gejohnston, @AHumanFromCA , @ChrisBoehmCA 
  - [x] Populate default values for options that were not already populated at the end of "prepare"
Status: Complete
Assigned: @ChrisBoehmCA 
  - [x] Add the aliases to the args object along with the kebab and camel case of the option
Status: Complete
Assigned: @gejohnston

8 point story

JTonda:
Comments from @tucker01  (original imperative repo issue)

Imperative should support a "search order" for option/argument values that includes:

Values specified on CLI options
Environment variables
Profiles
Design
Imperative CLIs use ICommandOptionDefinition to define options to commands. Yargs supplies command handlers with an "Argv" object in the form:

{
  "_":  ["cmd_segment_1", "cmd_segment_2"],
  "$0": "node ...",
  "option_one": "value_one",
  "optionOne": "value_one",
  "option_two": "value_two",
  "optionTwo": "value_two"
}
We will introduce an imperative Argv and deprecate the existing handler parms Argv. This change will allow handlers/CLIs/plugins to remove the @types/yargs and will insulate them from changes to the object made by Yargs.

We will inject options and their values into the imperative Argv as follows:

Yargs will pass the command processor the parsed 'Argv'
[...]

PR #83, opened 2018-10-19
Enhancement/34

tucker01:
- Introduces changes outlined in #34. 
- Fixes a recursion issue with command preparation and the `passOn` feature (was doubling up options "passed down" the tree)


tucker01:
I think a good portion of the team has commits in this branch, so I marked @zFernand0 to review as a neutral party 😄

packages/utilities/src/CliUtils.ts
```diff
@@ -95,6 +99,218 @@ export class CliUtils {
...
+     *
+     * @param {...any[]} args - variadic set of objects to be merged
+     *
+     * @returns {*} - the merged object
+     *
+     */
+    public static mergeArguments(...args: any[]): any {
+        let merged = {};
+        args.forEach((obj) => {
+            merged = {...merged, ...obj};
```
zFernand0: :warning: I hope you guys are aware that this is not a deep copy of the objects.

packages/utilities/src/CliUtils.ts
```diff
@@ -126,7 +342,204 @@ export class CliUtils {
...
+     *
+     * @param {Arguments} args - Yargs argument object
+     *
+     * @returns {ICommandArguments} - Imperative argument object
+     *
+     */
+    public static buildBaseArgs(args: Arguments): ICommandArguments {
+        const impArgs: ICommandArguments = {...args};
+        Object.keys(impArgs).forEach((key) => {
+            if (key !== "_" && key !== "$0" && impArgs[key] === undefined) {
```
zFernand0: Would you want to remove `null` and `''` (empty strings) as well?
Maybe use `!impArgs[key]` if that's the case :smile:
What about nested null properties? (would you want to recursively remove them too?)
Here is a [stackoverflow comment](https://stackoverflow.com/a/38340374/9937473) than might be interesting :yum:

packages/cmd/src/syntax/SyntaxValidator.ts
```diff
@@ -96,13 +97,13 @@ export class SyntaxValidator {
...
-     * @param {yargs.Arguments} commandArguments
+     * @param {ICommandArguments} commandArguments
      * @return {Promise<ICommandResponse>}
      */
-    public validate(responseObject: CommandResponse, commandArguments: Arguments): Promise<ICommandValidatorResponse> {
+    public validate(responseObject: CommandResponse, commandArguments: ICommandArguments): Promise<ICommandValidatorResponse> {
         return new Promise<ICommandValidatorResponse>((validationComplete) => {
             const syntaxValid: boolean = this.validateSyntax(commandArguments, responseObject);
-            validationComplete({ valid: syntaxValid });
+            validationComplete({valid: syntaxValid});
```
zFernand0: Again with the spaces :rofl:

packages/cmd/src/syntax/SyntaxValidator.ts
```diff
@@ -803,7 +859,7 @@ export class SyntaxValidator {
                     "The value must be a number",
```
zFernand0: Fun fact, this `validateNumeric` allows for things like `"1e-2"` which is technically equal to 0.01 :smile:
I'm grateful for simple implementations :+1:

(I'm putting the comment in here because GitHub wouldn't let me place on the `isNaN` line)

packages/cmd/src/utils/CommandUtils.ts
```diff
@@ -74,28 +73,25 @@ export class CommandUtils {
...
+    public static optionWasSpecified(optionName: string, commandDefinition: ICommandDefinition, args: ICommandArguments["args"]): boolean {
         const optionDef = CommandUtils.getOptionDefinitionFromName(optionName, commandDefinition);
-
         if (isNullOrUndefined(optionDef)) {
             // if it's not an option, it's not specified
             return false;
         }
         if ((optionDef.type as string) === "boolean") {
-            return commandArguments[optionName] === true;
+            return args[optionName] !== undefined;
```
zFernand0: Why the special treatment for the boolean types?
Is there a reason why you would want it to be `null`?
Why not just call the `isNullOrUndefined` function with `args[optionsName]`?

packages/cmd/src/utils/CommandUtils.ts
```diff
@@ -74,28 +73,25 @@ export class CommandUtils {
...
         }
         else {
-            return !isNullOrUndefined(commandArguments[optionName]);
+            return !isNullOrUndefined(args[optionName]);
```
zFernand0: I see this function is deprecated. Maybe we should think about "reimplementing" it, or using what [NodeJS suggests](https://nodejs.org/api/util.html#util_util_isnullorundefined_object)
`prop === undefined || prop === null`
I'm worried that they could take it out without any warning. :worried:

packages/utilities/src/CliUtils.ts
```diff
@@ -126,7 +342,204 @@ export class CliUtils {
...
+     * CliUtils.getOptionFormat("hello-World-");
+     *
+     * // returns
+     * const return4 = {
+     *     key: "hello-World-",
+     *     camelCase: "helloWorld",
+     *     kebabCase: "hello-world"
+     * }
+     */
+    public static getOptionFormat(key: string): IOptionFormat {
```
zFernand0: cool function :+1:

packages/cmd/src/CommandProcessor.ts
```diff
@@ -553,7 +617,7 @@ export class CommandProcessor {
             response.data.setMessage(handlerErr.message);
         } else if (handlerErr instanceof Error) {
             this.log.error(`Handler for ${this.mDefinition.name} rejected by unhandled exception.`);
-            response.setError({ msg: handlerErr.message, stack: handlerErr.stack });
+            response.setError({msg: handlerErr.message, stack: handlerErr.stack});
```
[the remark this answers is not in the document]
tucker01: I think we use `.editorconfig` which is supposed to help with this kind of stuff.

packages/utilities/src/CliUtils.ts
tucker01: Yes this is OK. It's really only keyword/value pairs.

packages/cmd/src/utils/CommandUtils.ts
tucker01: Yeah, we should be replacing these.

packages/utilities/src/CliUtils.ts
zFernand0: Thanks for clarifying

packages/cmd/src/CommandProcessor.ts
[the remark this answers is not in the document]
zFernand0: Cool stuff! :+1:

packages/cmd/src/utils/CommandUtils.ts
zFernand0: It's good for now though. No worries there.

Issue #87, opened 2018-10-24
Hardcoded use of @brightside in PMFConstants


AHumanFromCA:
The following lines exist in PMFConstants:

https://github.com/zowe/imperative/blob/09afc7e9275ecd0fffec889b7f3b5cfb5887bbd1/packages/imperative/src/plugins/utilities/PMFConstants.ts#L89-L91

This suggests that the `@brightside` namespace is hardcoded into our code when an imperative based cli or plugin for a host imperative application does not have to be part of the brightside namespace.

I have mitigated this a bit in my work for #33 but that still leaves us with messages like this lying around: (from the install plugins command)

```
Imperative's plugin installation program handles peer dependencies for modules
in the @brightside namespace, so you can safely ignore NPM warnings about
missing peer dependencies related to @brightside modules.
```

This text should be made more generic to not lock it down to an `@brightside` module.

zFernand0:
This issue can probably be closed.
We couls still use it to remove any brightside references in the code?
https://github.com/zowe/imperative/search?q=brightside&type=code

gejohnston:
The term "brightside" is no longer used within the plugins logic. So, the original issue has been addressed.

The remaining "brightside" references are in comments (not displayed to a user), are in test scripts (not displayed to end users), are used for backward-compatible queries into the secure credential store, and is used as an adjective for a profile. This issue has had no activity for 12 months. No action to clean up these remaining references is targeted for the next 2 calendar quarters. Therefore, this issue is being closed.

PR #94, opened 2018-10-26
Module Loader for Plugins

### LINKED ISSUE #33: Using pkg to create a binary breaks plugins
If you use the `pkg` tool to create a binary executable for an imperative based CLI, plugins will no longer function correctly. They will install, but cannot be executed. 

In summary, a symbolic link to the imperative package is created with `../snapshot/...` and the filesystem/node cannot resolve the link. 

The plugins will install fine, but any command will fail with "cannot find module imperative`.

Some investigation was done regarding module resolution in plugins. It turns out, if you use the absolute path (as dictated within the binary) in your plugins requires/imports, node will resolve imperative just fine and the plugin will function properly. In other words, if you change the import statements in the plugin to use an absolute path (at least on windows) of something like `C:\\snapshot\node_modules\...` it "works". 

A possible solution (which could eliminate the need for peerDependencies, the symlink, and admin privileges during plugin install) is to use a `require` hook to inject the correct path when imperative (and/or the base CLI) is required by a plugin. VSCode does something [similar](https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/api/node/extHost.api.impl.ts). 

- [x] Remove any symlink logic in the PluginManagementFacility
- [x] Finish creating the hook class to include methods for removing the hooks
- [x] Properly initialize the hook mechanism only when plugins are needed. If a CLI doesn't have plugins available, don't waste the processing power.
- [x] Confirmed that `@brightside/imperative` can be provided but we should also integrate requires to the base package as well. For example, `@brightside/core` for the zowe-cli should be properly intercepted for plugins.
  - [x] Also allow for submodule path imports just in case. E.G `require('@brightside/imperative/rest')` should work.
- [x] Properly unit / integration test
- [x] ~Port to 1.0.0?~ It has been decided that this will not be ported to 1.0.0

AHumanFromCA:
## Changelog

- **[Major]** Added a module loader that will inject Imperative and the Host Package whenever required anywhere during runtime. It is initialized by the Plugin Management Facility. (Resolves #33)
- **[Major]** Removed symlink creation logic in Plugin Management Facility, the injection of modules is now done at runtime by the new Module Loader.
- **[Minor]** Added `hostPackageName` and `imperativePackageName` as public properties for the `ImperativeConfig.instance` object.
    - `hostPackageName` is the name of the package imperative was called from
    - `imperativePackageName` is the name of Imperative's package.json name (in case we decide to change that at some point)
- **[Minor]** Remapped `PMFConstants.CLI_CORE_PKG_NAME` and `PMFConstants.IMPERATIVE_PKG_NAME` to point to the new values created in ImperativeConfig.
    - This partially addresses issue #87.
- **[Internal]** Added some trace log information to the validate plugin method.
- **[Internal]** Added a `MockWrapper` and `getMockWrapper()` for unit tests. See more below
- **[Internal]** Added a `MockMethod` decorator for unit tests. See more below.

### Using MockWrapper in Tests

```TypeScript
// In your test, there should be some jest mocks
jest.mock("./path/goes/here/Module1");
jest.mock("./path/goes/here/Module2");

import {Module1} from "./path/goes/here/Module1";
import {Module2} from "./path/goes/here/Module2";
import {Module3} from "./path/goes/here/Module3";

const mockObject = getMockWrapper({
   module1Function1: Module1.function1,
   module1Function2: Module1.function2,
   module2: Module2
});

// Typescript will now accept the following syntax.
mockObject.module2.mockImplementation(() => {
    console.log("Implementation can be mocked");
});
mockObject.module1Function1.mocks.clear();

Module1.function2("hooray proper types");
expect(module1Function2.mocks.calls[0][0]).toBe("hooray proper types");
```

### Using MockMethod Decorator in tests
```TypeScript
export class TestClass {
    // This method will be wrapped in a jest.fn() call with the implementation
    // being the contents of the below function.
    @MockMethod()
    public test() {
        return "string";
    }
}

// In some other file
const myClass = new TestClass();
console.log(myClass.test()); // Prints "string" to the console
expect(myClass.test).toHaveBeenCalledTimes(2);
// Fails with message
// Expected mock function to have been called two times, but it was called one time
```


packages/imperative/src/plugins/utilities/PMFConstants.ts
```diff
@@ -87,8 +87,8 @@ export class PMFConstants {
 
     constructor() {
         this.NPM_NAMESPACE = "@brightside";
-        this.CLI_CORE_PKG_NAME = `${this.NPM_NAMESPACE}/core`;
-        this.IMPERATIVE_PKG_NAME = `${this.NPM_NAMESPACE}/imperative`;
+        this.CLI_CORE_PKG_NAME = ImperativeConfig.instance.hostPackageName;
```
tucker01: Just curious, might be a "convenience" thing, but any reason to replicate the constant from imperative config in PMF constants?

AHumanFromCA: I wasn't too sure what things might still be using these constants in the PMF. So I went with consistency until we maybe can come back and resolve this as well when doing #87.

tucker01:
We should probably document in the Imperative wiki and/or the README.md that we are now hooking `require`. What would the implication be for projects that pull in imperative as a dependency (for a building a CLI or other purpose)? It looks like the hook is injected during `init`, so no impact if you don't `init`, but I'm thinking about the potential use of Imperative in our VS code extension (or other projects that already hook require). It appears that you are adjusting the path to `require` the "correct" Imperative and just passing thru to the original require. This should allow "other" hooks to function correctly? Anyway to confirm (and document if there is any "gotchas")?

AHumanFromCA:
> We should probably document in the Imperative wiki and/or the README.md that we are now hooking `require`.

I can work with @JamesBauman and @BrandonJenkins14 to get some doc in the Imperative wiki. I think that's where it should go.

> What would the implication be for projects that pull in imperative as a dependency (for a building a CLI or other purpose)? It looks like the hook is injected during init, so no impact if you don't init, but I'm thinking about the potential use of Imperative in our VS code extension (or other projects that already hook require).

It will hook Imperative and the host package if running directly from a node package. There might need to be more testing done to ensure that the module loader works as intended when a package includes both imperative and the host package. I suspect that this loader currently wouldn't handle that too well. This is a step in the right direction and that functionality could be built on top of this code later. I don't think symlinks would have worked much better (maybe even worse).

> It appears that you are adjusting the path to require the "correct" Imperative and just passing thru to the original require. This should allow "other" hooks to function correctly? Anyway to confirm (and document if there is any "gotchas")?

I'm overriding `Module.prototype.require` which is what node's require goes to. I did find some oddities with jest's require mechanizm where overriding the prototype did not override require, so that may still be an issue in places like VS code. I do forward to the original require though. So if VS code overrides `Module.prototype.require` then I will override it and then call VS code's require.

AHumanFromCA:
That statement was meant more for a host package and not a plugin.

I was referring to a structure that might look like this:

- /project-root
  - node_modules
    - host-package
    - @brightside/imperative
   - index.js

Where index.js may initialize the Plugin Management Facility for plugins that could be installed into the host-package (which is a cli when installed globally) so that they may use the APIs.

My case very closely represents the VS Code extension for zowe-cli that includes both Imperative and Zowe CLI as part of the dependencies. The module loader, as it currently sits, would have issues providing Zowe CLI to any plugins that were initialized in that runtime.

AHumanFromCA:
@tucker01 I've added some documentation to the `Plugins.md` file in the imperative wiki under branch `fix/33` if you want to review it for content.

tucker01:
Thanks! I'll take a peek when I get a chance.

PR #95, opened 2018-10-29
Enhancement/91

lyzla01:
Adding StringOrEmpty option type.  This option enable user to specify "" as input value.

### DIFF
--- a/__tests__/__integration__/cmd/src/cli/validate/string-or-empty/string-or-empty.definition.ts
+++ b/__tests__/__integration__/cmd/src/cli/validate/string-or-empty/string-or-empty.definition.ts
@@ -1,0 +1,35 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { ICommandDefinition } from "../../../../../../../packages/index";
+
+// Range for option length-range
+const LENGTH_RANGE_START: number = 2;
+const LENGTH_RANGE_END: number = 8;
+
+// Range for option eggs
+const EGGS_RANGE_START: number = 1;
+const EGGS_RANGE_END: number = 12;
+
+export const stringOrEmptyTestCommand: ICommandDefinition = {
+    name: "string-or-empty",
+    description: "Test string or empty option.",
+    type: "command",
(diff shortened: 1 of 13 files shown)


packages/io/__tests__/IO.test.ts
```diff
@@ -52,16 +52,6 @@ describe("IO tests", () => {
         expect(IO.isDir("pretend/file")).toBe(false);
     });
 
-    it("should get an error for no input on normalizeExtension", () => {
```
AHumanFromCA: Instead of removing this test, how about modifying it to test the new behavior. The test name could change to `it('should return no extension')` or something like that.

packages/cmd/src/doc/option/ICommandOptionDefinition.ts
```diff
@@ -18,9 +18,10 @@ import { ICommandOptionValueImplications } from "./ICommandOptionValueImplicatio
  * "existingLocalFile": a file for which fs.existsSync returns true
  * "json": a parseable JSON string
  * "array": an array of strings
+ * "stringOrEmpty" : allow string to be empty when defined
  */
 export type CommandOptionType
-    = "string" | "number" | "json" | "existingLocalFile" | "boolean" | "count" |
+    = "string" | "number" | "json" | "existingLocalFile" | "boolean" | "count" | "stringOrEmpty" |
```
AHumanFromCA: I would define `stringOrEmpty` at the end of the list after array since that is how it is ordered in the doc.

__tests__/__integration__/cmd/src/cli/validate/string-or-empty/string-or-empty.definition.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const stringOrEmptyTestCommand: ICommandDefinition = {
+    name: "string-or-empty",
+    description: "Test string or empty option.",
+    type: "command",
+    handler: __dirname + "/string-or-empty.handler",
+    options:
+        [
+            {
+                name: "string-or-empty",
```
AHumanFromCA: When I issue, `ts-node ./ validate string-or-empty`, from the cli directory it looks like validation passes and I assume that the value of `string-or-empty` is undefined? Or is the value ""?

If the value is "", how would this affect the case of specifying a default value and then omitting the option from the command. Is the default used instead of ""?

Perhaps you should echo the output of the option in your validation command to help understand this concept.

AHumanFromCA: Also, what happens if this option is required. I would expect that you now have to specify `--string-or-empty ""`, is this assumption correct?

Issue #96, opened 2018-10-30
Plugins cannot be installed in shipped products.


AHumanFromCA:
In a recent PR, npm apis were used directly to do installs. It looks like npm was added as a devDependency instead of a dependency. This leads to an `Cannot find module 'npm'` error on install.

The following needs to be completed for this issue to be resolved:

- [x] Move npm from devDependencies to dependencies.
- [ ] Adjust the linter to fail when items in packages are requiring to a devDependency.

AHumanFromCA:
These linting changes should also make their way to zowe-cli to ensure consistency

PR #97, opened 2018-10-30
Resolve the blocking portion of #96

AHumanFromCA:
- **[Fix]** It is now possible to install plugins again, let's try not to do that again (#96)


AHumanFromCA:
Cutting the line so this can be included in the PAX

PR #104, opened 2018-11-06
43 apiml integration 

tucker01:
# Overview
Implements the ability to specify "basePath" in the rest client session object. See the [Swagger](https://swagger.io/docs/specification/2-0/api-host-and-base-path/) doc for details on "base path". 

In summary, the "base path" is used/configured by API mediation layers and API Gateways to identify a service (and versioning). 

**Directly to the service:**
https://host:port/pets

**Gateway:**
https://host:port/api/v1/pet_service/pets

In the above example, `api/v1/pet_service` would be the "base path". 

CLI implementations which perform HTTP/REST calls can take advantage of this capability in order to be simultaneously compatible with a Gateway and going directly to a service.

## Additional Improvements
Adds additional fields/properties to the rest client error for debugging/diagnostics. The additional fields are also formatted into a printable error response (found in the `additionalDetails` property of the `ImperativeError`) and automatically formatted/surface if a RestClientError is thrown from a handler and caught by the Imperative command processor (normal error handling for command handlers).


packages/rest/src/client/AbstractRestClient.ts
```diff
@@ -365,39 +372,98 @@ export abstract class AbstractRestClient {
...
+        `HTTP(S) error status "${finalError.httpStatus}" received.\n` +
+        `Review request details (resource, base path, credentials, payload) and ensure correctness.`) +
+        "\n" +
+        "\nHost:      " + finalError.host +
+        "\nPort:      " + finalError.port +
+        "\nBase Path: " + finalError.basePath +
+        "\nResource:  " + finalError.resource +
+        "\nRequest:   " + finalError.request +
+        "\nHeaders:   " + headerDetails +
+        "\nPayload:   " + payloadDetails;
```
ChrisBoehmCA: This increased amount of information should hopefully be helpful to the user :+1:

packages/rest/src/session/doc/ISession.ts
```diff
@@ -53,6 +53,19 @@ export interface ISession {
...
+     * It is used to specify the first part of the resource in a URL that
+     * points to any API mediation layer, gateway, or router used at a site
+     * to dispatch web requests to services that are managed by that gateway.
+     * It is pre-pended to the resource path of the underlying service.
+     * When not supplied, or is an empty string we use the the resource path of
+     * the underlying service directly.
+     * @type {string}
+     * @memberof ISession
+     */
+    basePath?: string;
```
ChrisBoehmCA: Nice thorough explanation of the concept of base path :+1:

packages/rest/src/client/AbstractRestClient.ts
```diff
@@ -185,17 +188,11 @@ export abstract class AbstractRestClient {
              * Invoke any onError method whenever an error occurs on writing
              */
             clientRequest.on("error", (errorResponse: any) => {
-                let impError: IImperativeError = {
+                reject(this.populateError({
```
ChrisBoehmCA: Cool that we now get more details in the case of client request errors

packages/cmd/src/CommandProcessor.ts
```diff
@@ -522,7 +522,7 @@ export class CommandProcessor {
         args._ = commandArguments._;
 
         // Log for debugging
-        this.log.debug(`Full argument object constructed:\n${inspect(args)}`);
+        this.log.trace(`Full argument object constructed:\n${inspect(args)}`);
```
ChrisBoehmCA: Thanks for fixing this to not output password populated into the arguments

PR #114, opened 2018-11-12
Don't populate --name and --type arguments from profiles

### LINKED ISSUE #112: Options named --type are populated with the name of the profile type
This bug seems to have been introduced in #34. Since we are using every field of the profile mapping them to arguments, the `type` field which indicates the type of profile (e.g. `zosmf` or `tso`) is inserted into the arguments. 

So from a command handler, commandParameters.arguments.type would be `zosmf` instead of being blank.

ChrisBoehmCA:
When building arguments from profiles as implemented in #34, do not use the name or type fields of the profile as arguments since they are special reserved fields for profiles.

Add integration tests for the name/type scenario

Fix #112


AHumanFromCA:
Closing this to allow it to be re-worked.

PR #120, opened 2018-11-19
Eliminate name and type from loaded profiles 

### LINKED ISSUE #112: Options named --type are populated with the name of the profile type

ChrisBoehmCA:
Breaking changes: 
* When profiles are loaded, `name` and `type` are no longer injected into the contents. So `name` and `type` are no longer reserved words and `name` and `type` properties could be defined in a profile schema. Since these fields aren't present, they no longer are mapped to arguments due to the changes in #34, which fixes #112. 

image

* Profile manager APIs have changed to require `name` and `type` as a parameter in some cases where the value was previously determined from the contents of the profile. Unit tests have been updated to reflect this. Two affected APIs are save() and update()


tucker01:
It's worth mentioning that this PR also fixes an issue where `name` and `type` are incorrectly filled in during option precedence search (if a profile is present).

packages/profiles/__tests__/BasicProfileManager.save.test.ts
```diff
@@ -15,589 +15,589 @@ import { TestLogger } from "../../../__tests__/TestLogger";
...
-    expect(error.message).toMatchSnapshot();
-  });
-
-  it("should detect that the dependencies are present, but type is missing", async () => {
-    const prof = new BasicProfileManager({
-      profileRootDirectory: TEST_PROFILE_ROOT_DIR,
-      typeConfigurations: ONLY_APPLE,
-      type: APPLE_PROFILE_TYPE,
-      logger: TestLogger.getTestLogger()
+    it("should detect a blank name when creating a profile", async () => {
```
tucker01: should there be a test on save for omitting name entirely? I don't think I saw one?

packages/profiles/src/abstract/AbstractProfileManager.ts
```diff
@@ -397,20 +397,20 @@ export abstract class AbstractProfileManager<T extends IProfileTypeConfiguration
             `A request was made to save a profile of type "${this.profileType}", but no profile was supplied.`);
 
         // Ensure that the type is filled in - a mismatch will be thrown if the type indicates a type other than the manager's current type
-        parms.profile.type = parms.profile.type || this.profileType;
+        parms.type = parms.type || this.profileType;
 
         // Log the invocation
-        this.log.info(`Saving profile "${parms.profile.name}" of type "${parms.profile.type}"...`);
+        this.log.info(`Saving profile "${parms.name}" of type "${this.profileType}"...`);
```
tucker01: Should there be a ImperativeExpect check somewhere around here for if name was specified?

tucker01:
Thanks!

MikeBauerCA:
Any chance we could get this merged soon? Would be great to have for some upcoming demoes :) Otherwise, I need to continue specifying type directly on the command.

AHumanFromCA:
Done

PR #122, opened 2018-11-20
Cii badge compliance

MarkAckert:
Adds basic sonar scanning to the build


package.json
```diff
@@ -95,8 +95,10 @@
     "jest": "21.2.1",
     "jest-environment-node": "21.2.1",
     "jest-environment-node-debug": "2.0.0",
-    "jest-html-reporter": "2.0.0",
+    "jest-html-reporter": "^0.5.7",
```
ChrisBoehmCA: Why was the html reporter downgraded?

MarkAckert: I set html reporter back to 2.0.0. I think I originally had some issue with sonar & the the html reporter@2.0.0, but re-ran sonar locally with 2.0.0 and had no failure.

ChrisBoehmCA: Thanks

PR #135, opened 2018-12-13
Display available commands

tucker01:
Adding a root level option to display all available commands (with bin prefix) via the `--available-commands` or `--ac` option.

Implemented to address a usability concern from a user of an Imperative based CLI. The user indicated they were having difficulty understanding what commands were available in the CLI. They had to issue multiple help commands at different "group" levels to understand what was available. 

Thought I'd throw this together as a quick/simple solution. Feel free to make UX changes to option names and output. 

Example from an existing CLI:

image


ChrisBoehmCA:
I like it. I think if the output is too long there are various things people can do like redirect to a file or program like `less`. I think the display is cool too.

AHumanFromCA:
@tucker01 So if I did something like `zowe files --ac` will it show me only things under that group? If so thank you because I never know what the commands are after the 3rd level.

@jelaplan what's your take on this.

ChrisBoehmCA:
@AHumanFromCA  That is definitely possible but not implemented in this PR yet. I believe this only currently works on the root command of the CLI

tucker01:
@ChrisBoehmCA That is correct, it only works at the root level. It's a little more work to implement at each "group" and can be done (maybe separately) if desire-able. This is a quick first pass.

jelaplan:
@AhumanFromCA This looks good. There was an e-mail thread about this. I had suggested allowing a user to ask for the commands within a group. But, this whole thing is gravy so I'm not going to press. 
- Did we ever share this with the guy who asked for it? Couldn't hurt to send him a e-mail.
- I wish we had  a search for commands. I think we may have had that pre-imperitive. That would complement this feature. 
- This reveals some descriptions that might have been written better. It is interesting to see these very short descriptions and try to quickly scan the list for something. Maybe writing a description for a single command is a different task than writing descriptions for a long summary list.

jelaplan:
BTW, the e-mail exchange was titled Re: help "issue"

PR #146, opened 2019-01-25
Fix issue where profile name is not displayed when a secure field fails to save

### LINKED ISSUE #147: Profile name displayed as 'undefined' when secure fields fail to save
Example output: 
```
Command Error:
Unable to store the secure field "user" associated with the profile "undefined" of type "zosmf".

```

ChrisBoehmCA:
Fix bug introduced in  #120 
Fix #147


tucker01:
Thanks!

PR #148, opened 2019-01-29
Update jenkinsfile to accommodate all deployment

zFernand0:
- Enhance pipeline to implement CI for all intended release branches

Successful run: https://wash.zowe.org:8443/blue/organizations/jenkins/imperative/detail/master/216/pipeline/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,14 +17,20 @@ def UNIT_RESULTS = "${TEST_RESULTS_FOLDER}/unit"
 def INTEGRATION_RESULTS = "${TEST_RESULTS_FOLDER}/integration"
 
 /**
- * The name of the master branch
- */
-def MASTER_BRANCH = "master"
+* Development branches considered for release purposes
+*/ 
+def DEV_BRANCH = [
+    master: "master",
+    beta: "beta",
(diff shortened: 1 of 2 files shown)


AHumanFromCA:
I think both `@daily` and `@beta` are necessary.

`@daily` allows us to quickly get any changes from a registry in master of other repos.
`@beta` allows for a slightly more stable version to exist and allow plugin developers/clis to have some time to adjust to the changes.

If we think we don't need both, I would say `@beta` should go and we just move the latest `@daily` to `@latest` when we have a validated version.

tucker01:
Thanks @AHumanFromCA. 

@zFernand0, @MikeBauerCA, and I had a conversation last night regarding the usages of `@beta` and `@daily`. It wasn't clear that both prerelease versions (and validation) was necessary for changes that would eventually make its way to `@latest` and there may be situations where we wouldn't want one. Trying to keep the publishing/testing process as simple as possible. 

At any rate, it's a question at the moment and @zFernand0 (and you guys) can work out the details.

zFernand0:
Maintainer guidelines will be removed by [PR 192](https://github.com/zowe/zowe-cli/pull/192)

Issue #136, opened 2018-12-13
Unit test failures with strange "random" path changes


tucker01:
When running unit tests on Jenkins or local PC test will intermittently fail with changes to paths and slashes. It is seemingly random. Rerunning tests or replaying builds may pass. Witnessed in rest client tests, plugin management tests, among others. 
```
- Snapshot
+ Received
@@ -2,9 +2,9 @@
"headers": Object {
"Content-Type": "application/json",
},
"hostname": "test",
"method": "GET",
- "path": "/resource",
+ "path": "///resource",
"port": 443,
"rejectUnauthorized": true,
}
```

gejohnston:
The unit tests no longer produce these errors.

PR #153, opened 2019-01-31
Posix.Normalize the output path

zFernand0:
***DO NOT MERGE YET***
The purpose of this PR is to ensure that this random failure does not occur again in the future.

Please do not merge this PR right away as it could serve as a test in enough itself.

This is an attempt to address #136


zFernand0:
Seems like this is not a fix 😢 
https://wash.zowe.org:8443/blue/organizations/jenkins/imperative/detail/fix%2F136/1/tests

PR #162, opened 2019-02-08
Fix issue where required array type positionals could not be validated

### LINKED ISSUE #161: Required array-type positional arguments fail syntax validation
There is a bug where if a positional argument specified as an array by adding `...` to the end of the name such as in this example: 
```
  positionals: [
                    {
                        name: "my-array...",
                        required: true,
                        type: "string",
                        description: "gimme those strings"
                    }
                ],
```

Syntax validation will always fail with a message that my-array... was not specified

We have a PR in the works to fix this issue

ChrisBoehmCA:
Fix #161 where required array type positionals could not be validated
Add tests for this scenario


packages/cmd/src/syntax/SyntaxValidator.ts
```diff
@@ -203,7 +203,8 @@ export class SyntaxValidator {
             const missingPositionals: ICommandPositionalDefinition[] = [];
             for (const positional of this.mCommandDefinition.positionals) {
                 if (positional.required) {
-                    if (!commandArguments[positional.name]) {
+                    // Use replace to trim possible ... which is used for arrays
+                    if (!commandArguments[positional.name.replace("...","")]) {
```
zFernand0: :+1:

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@brightside/imperative",
-  "version": "2.1.0-alpha.201902072106",
+  "version": "2.1.1-alpha.201902072106",
```
zFernand0: :+1:

__tests__/src/packages/cmd/SyntaxValidator.integration.test.ts
```diff
@@ -325,6 +325,90 @@ describe("Imperative should provide advanced syntax validation rules", function
                     minValidOptions + "--always-required-string hello",
                     false, ["multiple", "--always-required-string"])();
             });
+        describe("We should be able to validate array positional arguments", function () {
```
zFernand0: Thanks for adding the test

PR #166, opened 2019-02-15
Zowe/zowe cli/142

AHumanFromCA:
- Implement shared pipeline as part of zowe/zowe-cli#142
- Also might fix issue #136


AHumanFromCA:
I have fixed all of the failing tests because of the required jest update. The only thing that has failed now is SonarQube. Looks like that server might be down right now.

PR #168, opened 2019-02-19
Allow a blank file to be written with IO.writeFileAsync

ChrisBoehmCA:
In certain contexts such as https://github.com/zowe/zowe-cli/issues/218, writing a blank file is valid.  When this change is published and consumed from Zowe CLI, this issue will be fixed


zFernand0:
👍

PR #170, opened 2019-02-20
Enhancement/130

### LINKED ISSUE #130: Integrate performance timing library in key places of imperative
Acceptance Criteria
-requires typescript update to 3.X
- key areas where performance can be improved in the imperative framework are identified
  - Start by placing a timer in each facility to record when it starts and stops. - DONE (ConfigManagementFacility and PluginManagementFacility)
  - Place in the requires function - DONE
  - Put a timer on how long it takes each plugin to load - DONE (Add plugin and Load plugin)
  - place in logging - DONE (added new common function and timing of logger. 
  - Also place in REST - DONE (AbstractRestClient.performRest())
  - Place in init - DONE (Imperative.init())
  - Places in command processing - DONE (CommandProcessor.invoke())
- publish

Estimate 5
Child of zowe/zowe-cli#133

PeteSwauger:
Added performance timing (closes #130) and updated typescript.


packages/imperative/src/Imperative.ts
```diff
@@ -100,6 +118,14 @@ export class Imperative {
     public static init(config?: IImperativeConfig): Promise<void> {
         return new Promise<void>(async (initializationComplete: () => void, initializationFailed: ImperativeReject) => {
             try {
+
+                const timingApi = PerfTiming.api;
+
+                if (PerfTiming.isEnabled) {
```
AHumanFromCA: In addition to timing the init function, I think we should also time the parse function.

If memory serves me correctly, both Imperative.init and Imperative.parse are called by implementing CLIs. Timing both will give us a better vision as to where the slow downs are.

packages/logger/src/Logger.ts
```diff
@@ -314,6 +276,42 @@ export class Logger {
...
+     * Log a message at the level requested
+     *  Example: 'Cheese is Gouda'
+     * @param method - level of logging
+     * @param message - printf or mustache style template string, or a plain string message
+     * @param args - printf or mustache style args
+     * @returns {any}
+     */
+    public writeToLog(method: string, message: string, args: any[]): string {
+
+        // TODO - determine how to check log level to avoid formatting and queuing messages that won't be written to the log.
```
AHumanFromCA: Is this TODO being done in a different branch?

zFernand0:
Does updating typescript constitutes a breaking change for users that integrate Imperative as a Dev Dependency?
Depending on the answer, this PR could be tagged with `release-major` and subsequently bump the major version accordingly. 
Note: Having a new major version means that features that we want to include as part of the Zowe 1.x releases, need to be ported to the incremental branch from this point forward.

AHumanFromCA:
> Does updating typescript constitutes a breaking change for users that integrate Imperative as a Dev Dependency?

Yes it would (for both dependency and dev) if they need to run tsc on their application and any module imports Imperative when their typescript is <= 3.0.x

packages/security/__tests__/CredentialManagerFactory.test.ts
```diff
@@ -155,20 +155,20 @@ describe("CredentialManagerFactory", () => {
             );
         });
 
-        // Note: For some reason using a mock function on logger was not producing the desired results, broke down and mocked stderr
```
AHumanFromCA: Is there any reason why this test was commented out in this branch? I see that it must have been in master since the PR is reporting that this is a new change.

PeteSwauger: I missed this change in my reverting of the logging.  Will revert locally and test again.

AHumanFromCA: Thanks!

PR #173, opened 2019-02-26
Check that tests compile during build

### LINKED ISSUE #169: Add check for tests compilation
This would prevent compilation errors in the test code from going unnoticed.

This can be achieved by adding another typescript config json file for the test source and using the --noEmit flag on `tsc` 

Example of doing this in zowe CLI: https://github.com/zowe/zowe-cli/blob/master/package.json#L44

ChrisBoehmCA:
Fix existing compilation errors
Fix #169


__tests__/src/packages/profiles/src/constants/BasicProfileManagerTestContants.ts
```diff
@@ -62,15 +62,15 @@ const secureOrangeProfile: ICommandProfileTypeConfiguration = {
       username: {
         optionDefinition: {
           description: "The username of the secure_orange",
-          type: "username",
+          type: "string",
```
zFernand0: Thanks for updating

package.json
```diff
@@ -32,6 +32,7 @@
     "build:packages": "gulp build",
     "build": "gulp build",
     "postbuild": "gulp build:install-all-cli-dependencies && gulp build:all-clis",
+    "checkTestsCompile": "echo \"Checking that test source compiles\" && tsc --noEmit -p tsconfig-tests.json",
```
zFernand0: What do you think about adding it in the postbuild npm script?

Jenkinsfile
```diff
@@ -72,7 +72,14 @@ node('ca-jenkins-agent') {
 
     // Build the application
     pipeline.build(
-        timeout: [time: 5, unit: 'MINUTES']
+        operation: {
+            sh "npm run build"
+            sh "npm run checkTestsCompile"
```
zFernand0: If we decide to make `checkTestCOmpile` part of the `postbuild` npm script, then we could remove this `operation`

package.json
ChrisBoehmCA: That's fine with me. I will change it

zFernand0:
merging

PR #177, opened 2019-03-06
Enhance config command group

js665999:
- Add GET command 
- Add LIST command 
- refactor AppSettings.ts

### DIFF
--- a/__tests__/__integration__/cmd/package-lock.json
+++ b/__tests__/__integration__/cmd/package-lock.json
@@ -34,6 +34,7 @@
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-2.1.1.tgz",
       "integrity": "sha1-1t4Q1a9hMtW9aSQn1G/FOFOQlMc=",
+      "optional": true,
       "requires": {
         "extend": "~3.0.0",
         "semver": "~5.0.1"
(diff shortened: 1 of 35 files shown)


tucker01:
As the `imperative.json` file is updated/expanded with additional configuration items, how will the user know what is able to be set? It doesn't appear to mention the "set-able" values in the command definition?

js665999:
> As the `imperative.json` file is updated/expanded with additional configuration items, how will the user know what is able to be set? It doesn't appear to mention the "set-able" values in the command definition?

List command is used to list all items which can be set

tucker01:
How will users know to use `list` to see the "set-able" props and how will users know what values are valid for those props? 

Also, with the option name change, this cannot be ported to lts-incremental (breaking).

jelaplan:
@BrandonJenkins14 Wrote
Hi, my only suggestion here is that if/when we add another override, we place a doc task to that story to add a new topic in the Configuration section of DocOps/Zowe doc. Something like "Using the Config Group" or "Setting Configuration Options", which can describe some of these complexities. 

Currently, the Secure Credential Plug-in doc explains how to set the CredentialManager value, which is probably sufficient right now.

ChrisBoehmCA:
Thanks for updating based on feedback. Please resolve the merge conflicts

PR #182, opened 2019-03-14
(lts-incremental) output plugins object in data field for plugins list --rfj

ChrisBoehmCA:
Same changes as #181 for lts-incremental


ChrisBoehmCA:
Included more commits than I wanted to

PR #184, opened 2019-03-15
Update jenkinsfile in LTS incremental

ChrisBoehmCA:
Jenkinsfile wasn't using the new library. Copied the file from master, changing @zowe to @brightside since this is LTS incremental


zFernand0:
This is technically a path, however, since the previous minor enhancement didn't deploy, we should increase the minor level at this point either way

PR #203, opened 2019-04-18
Escape special characters when generating Markdown, fix #202

### LINKED ISSUE #202: When generating Markdown from command help, # is not escaped
If I put a '#' character in my command's description, when the Markdown doc is generated this is not escaped. So it's interpreted as a heading. 

For example by including the following line in my description:
```
Upload a local directory to a USS directory.\n\n
An optional .zosattributes file in the source directory can be used to control file conversion and tagging.\n\n
An example .zosattributes file: \n
# pattern        local-encoding        remote-encoding
```
the generated Markdown looks like this:
### dir-to-uss<a name="command-dir-to-uss"></a>
Upload a local directory to a USS directory.

An optional .zosattributes file in the source directory can be used to control
file conversion and tagging.

An example .zosattributes file:
# pattern local-encoding remote-encoding

matthewpwilson:
It turns out there are a bunch of special characters that need escaping when generating Markdown.  This fix slightly over-escapes e.g. periods only need escaping when following a number. However, it's still corrrect and without implementing half a Markdown parser I think it's tricky to do better.

### DIFF
--- a/packages/cmd/__tests__/help/DefaultHelpGenerator.test.ts
+++ b/packages/cmd/__tests__/help/DefaultHelpGenerator.test.ts
@@ -40,6 +40,12 @@ describe("Default Help Generator", () => {
         produceMarkdown: false
     };
 
+    const PARMS_GENERATE_MARKDOWN: IHelpGeneratorFactoryParms = {
+        primaryHighlightColor: chalkColor,
+        rootCommandName: "test-help",
+        produceMarkdown: true
+    };
+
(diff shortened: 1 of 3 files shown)


ChrisBoehmCA:
Could you port this change to the lts-incremental branch as well please?

matthewpwilson:
> Could you port this change to the lts-incremental branch as well please?

Sure, will do.

zFernand0:
fix #202

PR #258, opened 2019-07-25
Get command group examples

uzust01:
Introduce new global command --help-examples (--he) 

Functionality/availability:
- Zowe/Bright (top tier) - does not work*/available in Global options**.
- Groups and commands - displays all the examples in the current definition and all its children/available in Global options.
- Objects - has the same functionality as regular help/available in Global options.

*It does not work at the top level, because there is no definition for it. It does not throw  errors, just re-executes _zowe_ command 
**It has to be available there otherwise (even when listed as an option on the lower levels) it does not work. The returned by _CommandPreparer.prepare_ object does not have it as an option in the options array. I could not find a way around this, so what I did is substitute the description at this level. (Open for suggestions on this part, I would like to have the option hidden at zowe level)

In conclusion, maybe this fixes https://github.com/zowe/zowe-cli/issues/409


zFernand0:
Someone brought up that having too many global options may restrict users on what option names and aliases they can use to avoid collisions.
@tucker01, any thoughts?

PeteSwauger:
I agree with Gene that help-web should move ahead of help-extended help.  I would be willing to pick up option 2 and update the plugins after web help is merged into master.

zFernand0:
We can have a quick discussion (over slack, or a quick call maybe).
Thanks for adding all this information (it helps provide some clarity).
The option we've been doing is `#3` since we may not have control over some other plugins that other entities may be building.

zFernand0:
question, are we going to keep the `--hex` alias?

uzust01:
I can change it if requested, I was keeping it only for the development, but it kind of stayed.
@jelaplan

jelaplan:
--he seems good. I tend to think shorter is better and --he matches up with --hw

tucker01:
I'm concerned that too many global options with aliases will pollute commands. In other words, every alias added as a global option restricts the usage of those characters at lower level commands for other options. I can definitely see a command wanting to use `--hex` as its own option.

I'm wondering if we should restrict new global options to full name only.

zFernand0:
> I'm concerned that too many global options with aliases will pollute commands. In other words, every alias added as a global option restricts the usage of those characters at lower level commands for other options. I can definitely see a command wanting to use `--hex` as its own option.
> 
> I'm wondering if we should restrict new global options to full name only.

I'm ok with this approach.
Should we be removing other aliases (or have them as "hidden/overwrite-able")?

uzust01:
> > I'm concerned that too many global options with aliases will pollute commands. In other words, every alias added as a global option restricts the usage of those characters at lower level commands for other options. I can definitely see a command wanting to use `--hex` as its own option.
> > I'm wondering if we should restrict new global options to full name only.
> 
> I'm ok with this approach.
> Should we be removing other aliases (or have them as "hidden/overwrite-able")?

Are we starting with --help-examples, should I remove the alias?

tucker01:
We had a quick chat at stand-up and we're going to move forward for now with the "no alias on global options" rule.

If you don't mind removing it, that would be great. Thanks!

uzust01:
OK, Removed the alias for --help-examples, what about --help-web ? Should we use this PR to remove that one or will it stay ? forever :D

tucker01:
Thank you! We can open a separate PR for web-help.

zFernand0:
> Thank you! We can open a separate PR for web-help.

FYI, that'd be another braking change since it was already delivered to `@latest`

tucker01:
Maybe it should just stay as-is then for now.

tucker01:
Is @gejohnston OK with this PR (listed as the reviewer)? Any other changes required?

zFernand0:
Version not selected, which led to the same alpha version 4.2.2.
Rerunning pipeline to bump to 4.3.0

PR #259, opened 2019-07-26
Fix integration test on windows

zFernand0:
*No description provided. - Fernando*


tucker01:
Since you didn't provide a description! I'm assuming this change is to compare the code coverage against the commit you branched from? This will hopefully prevent someone getting a "free pass" if the coverage is lets say "90%" and their PR adds no tests, but only brings it down to "89%" (which would be above the 80 threshold)?

zFernand0:
> Since you didn't provide a description! I'm assuming this change is to compare the code coverage against the commit you branched from? This will hopefully prevent someone getting a "free pass" if the coverage is lets say "90%" and their PR adds no tests, but only brings it down to "89%" (which would be above the 80 threshold)?

That's the expectation of this change.
And also to fix the integration test on Windows since we were mocking the entire NodeJS APIs for the OS (which cause a logger test to fail).

tucker01:
Awesome thanks!

PR #268, opened 2019-08-14
Fix profile update spacing

awharnBCM:
Resolves issue #265: Extra Spacing on Profile Update

Also changes the text displayed on successful profile update to 'Profile updated successfully' instead of 'Profile created successfully' and updates integration test accordingly.


zFernand0:
> Changes look good... generally changes to command output are consider major breaking changes. Should this cause a major bump @zFernand0 ?

You're right. It's a breaking change.

PR #278, opened 2019-08-26
Fix web help issues

### LINKED ISSUE #273: Postpone import of WebHelp dependencies
Use when-needed requires of these 3 dependencies:
* fs-extra
* opener
* marked

The intent is to avoid importing these modules on every command, when they are only needed for WebHelp.

### LINKED ISSUE #277: Web help command tree sometimes renders incorrectly
When Imperative `browserify`s the web help, the order that CSS files get bundled in is random. This order is important, because if the CSS files get bundled in the wrong order this can happen:
image

tjohnsonBCM:
* Fixes #273 by postponing import of web help dependencies until they are needed
* Fixes #277 by forcing browserify to concatenate CSS in required order

### DIFF
--- a/gulp/DevelopmentTasks.ts
+++ b/gulp/DevelopmentTasks.ts
@@ -293,6 +293,7 @@ const browserify: ITaskFunction = async() => {
     });
     b.bundle().pipe(fs.createWriteStream(__dirname + "/../web-help/dist/js/bundle-docs.js"));
 };
+browserify.description = "Browserify dependencies for web help";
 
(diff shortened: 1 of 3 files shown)


tucker01:
@tjohnsonBCM for `WebHelpGenerator.ts` what's the coverage when you run its unit test on your PC?

zFernand0:
@tucker01, should this PR have been delivered as `@zowe/imperative@4.3.2` ?

tucker01:
It's a patch.

PR #287, opened 2019-09-12
Web help enhancements

### LINKED ISSUE #279: Provide means of copying web help URL for specific command
Currently, when browsing web help on the Zowe docs GitHub site, the link in the address bar only shows `index.html`. If you are browsing help for a specific command, it would be useful to have a means of copying a link of the format `index.html?p=<command_id>`. This would create a shareable link that would point others to the specific command instead of the web help root page.

### LINKED ISSUE #282: Fuzzy search in web help command tree
Currently, if you enter "list data set" or "list dataset" in the web help search box, there are no matches. A dash is required in the query ("list data-set") so that it exactly matches the name of the command. It would be ideal if a space or no character can be substituted in place of the dash when searching.

### LINKED ISSUE #286: Web help - search all help content
Web help does not allow for searching the help content. It only allows for searching the command tree. 

From @jelaplan:
I can't search full text. This means that common things that a user would search on don't appear, such as the term 'examples' and search results are very limited for terms like 'profile' which appear in many groups. This is a weakness of the HTML version relative to the PDF version. While the HTML has a lot of advantages, full text search is an exception. It would be possible to show the full text in HTML as well to support full text searching.

tjohnsonBCM:
* Resolves #279 by always including name of current command in URL shown in address bar
* Resolves #282 by implementing fuzzy search for command names that contain hyphens
* Resolves #286 by adding a flat view that shows full text which users can search (screenshot below)
* Fixes bug where Browserify could incorrectly mix Bootstrap CSS with GitHub markdown CSS
* Moves dependencies needed by web help to devDependencies, now that we're using Browserify

image


zFernand0:
> Hey @zFernand0, are we still keeping dependencies locked into a particular version? Should the `^` be removed from `package.json`.

That's correct, we are still trying to keep everything locked down to a specific version inorder to avoid undesirable vulnerabilities.

tucker01:
> > Hey @zFernand0, are we still keeping dependencies locked into a particular version? Should the `^` be removed from `package.json`.
> 
> That's correct, we are still trying to keep everything locked down to a specific version inorder to avoid undesirable vulnerabilities.

OK perfect thanks. @tjohnsonBCM please remove the `^`.

tjohnsonBCM:
Removed `^` in `package.json` for web help dependencies. There are other dependencies that had `^` which I left unchanged.

The Jenkins build after my last commit had an odd failure relating to Keytar in an integration test: https://wash.zowe.org:8443/blue/organizations/jenkins/imperative/detail/PR-287/2/pipeline
Reran build and it passed.

zFernand0:
> > @tjohnsonBCM please remove the `^`.
> 
> Removed `^` in `package.json` for web help dependencies. There are other dependencies that had `^` which I left unchanged.
> 
> The Jenkins build after my last commit had an odd failure related to Keytar in an integration test: https://wash.zowe.org:8443/blue/organizations/jenkins/imperative/detail/PR-287/2/pipeline
> Reran build and it passed.

Yeah, we have seen this happen before. We don't have a workaround other than rerunning the build. If this becomes an issue in the future, we can investigate further.
Thanks. :)

PR #293, opened 2019-09-30
Web help enhancements (lts-incremental)

tjohnsonBCM:
Same as #287, but for lts-incremental branch

#291, #292 should be merged first


package.json
```diff
@@ -131,11 +128,16 @@
...
     "jest-sonar-reporter": "2.0.0",
     "jest-stare": "1.11.1",
+    "jquery": "3.4.1",
     "jsdoc": "^3.6.2",
+    "jstree": "3.3.8",
     "keytar": "^4.11.0",
     "madge": "3.0.0",
+    "popper.js": "1.15.0",
     "prompt": "1.0.0",
+    "scroll-into-view-if-needed": "2.2.20",
```
zFernand0: Do we need to update the `TPSR` list?
@gejohnston, @tucker01, @JamesBauman, @BrandonJenkins14, any thoughts?
https://docs.zowe.org/stable/appendix/tpsr.html#imperative-cli-framework-dependency-attributions

An older version of the same package seems to be part of the API ML.

Issue #15, opened 2018-08-30
Provide instructions for installing package as a dependency


ChrisBoehmCA:
The README should provide instructions for how to install Imperative as a dependency

PR #304, opened 2019-10-14
Fix unexpected behavior after default profile is deleted

### LINKED ISSUE #43: defaultProfile value in meta.yaml file not cleared if default profile is deleted
_From @AHumanFromCA on June 26, 2018 19:27_

## From the Endevor Plugin Folks

When you delete the default profile, the profileName_meta.yaml still has the defaultProfile attributes. This leads to weird error message "Cannot read property 'hostname' of undefined" in our side. (the undefined is the ghost default profile)

## My Take

It looks like deleting the default profile leads to us not unassigning it as default. When the default is called on by a future command, this *ghost* profile is loaded as undefined leading to these unexpected errors. It probably should say something like: default profile not yet assigned...or whatever we normally would say in this scenario.

_Copied from original issue: gizafoundation/imperative#114_

tjohnsonBCM:
* Fixes #43 (also [zowe-cli #108](https://github.com/zowe/zowe-cli/issues/108), [zowe-cli #541](https://github.com/zowe/zowe-cli/issues/541), [zowe-cli-cics-plugin #15](https://github.com/zowe/zowe-cli-cics-plugin/issues/15) once their Imperative dependency is updated)
* When attempting to load a default profile, if it cannot be loaded, instead of "Cannot read property 'host' of undefined" the following message will be shown:
image
* When a profile is deleted, Imperative now checks whether it is the default profile for its type. If so the default profile name for that type is reset to `null`, so that there will be no default profile rather than a non-existent one. This results in a message like the following:
image
* If a user manually deletes a profile YAML file, it is still possible for their default profile to become non-existent. Implementing #305 would add a command that allows users to recover from this state.

### DIFF
--- a/packages/profiles/src/abstract/AbstractProfileManager.ts
+++ b/packages/profiles/src/abstract/AbstractProfileManager.ts
@@ -692,6 +692,16 @@ export abstract class AbstractProfileManager<T extends IProfileTypeConfiguration
         {tag: `Internal Profile Management Error`});
     }
 
+    // If the meta file exists - read to check if the name of the default profile is the same as
+    // the profile that was deleted. If so, reset it to null.
+    if (this.locateExistingProfile(this.constructMetaName())) {
+      const meta = this.readMeta(this.constructFullProfilePath(this.constructMetaName()));
+      if (meta.defaultProfile === parms.name) {
+        this.log.debug(`Profile deleted was the default. Clearing the default profile for type "${this.profileType}".`);
+        this.clearDefault();
+      }
+    }
+
     return response;
   }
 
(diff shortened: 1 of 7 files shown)


packages/profiles/src/abstract/AbstractProfileManager.ts
```diff
@@ -470,11 +471,20 @@ export abstract class AbstractProfileManager<T extends IProfileTypeConfiguration
       this.log.debug(`The default profile for type "${this.profileType}" is "${parms.name}".`);
 
       // If we don't find the default name and we know fail not found is false, then return here
-      if (isNullOrUndefined(parms.name) && !parms.failNotFound) {
-        return this.failNotFoundDefaultResponse("default was requested");
-      } else if (isNullOrUndefined(parms.name)) {
-        this.log.error(`No default profile exists for type "${this.profileType}".`);
-        throw new ImperativeError({msg: `No default profile set for type "${this.profileType}".`});
+      if (isNullOrUndefined(parms.name)) {
```
tucker01: `isNullOrUndefined` is a deprecated function. The preferred way of checking is `parms.name == null`. Note the double equals `==`.

tucker01:
Hey @tjohnsonBCM, can you checkout the failing test?

tjohnsonBCM:
There was a transient failure of a CredentialManager unit test, ran the build 2 more times and it passed.

PR #256, opened 2019-07-18
remove npm dependency

nurra01:
Remove npm from local dependency.

PR #308, opened 2019-10-21
Fixes vulnerability and adds typedoc

zFernand0:
- Removes npm as a dependency of imperative (hence why it removed the vulnerability)
- Add typedoc script 

Signed-off-by: zFernand0 <fernando.rijocedeno@broadcom.com>


zFernand0:
Basically porting https://github.com/zowe/imperative/pull/256 now :cry:

PR #309, opened 2019-10-23
Add Print button to web help pages

tjohnsonBCM:
Adds a Print button at the top of the docs page for each CLI command.

Before:
image
After:
image

This makes possible a new method for generating the CLI reference PDF. In the flattened view of web help, it is now possible to print the HTML page as a PDF and a table of contents will be inserted.


jelaplan:
The angle characters look better and more suitable to a breadcrumb.

Issue #316, opened 2019-10-29
Abstract.performRest Overload


dkelosky:
`AbstractRestClient.performRest()` has been overloaded for streams and other options, but the method name is a bit more difficult to use now.  Can this be overloaded to add an options object for the optional parms?

For example:
```typescript
        return new this(session).performRest(resource, HTTP_VERB.GET, reqHeaders, undefined, responseStream,
            undefined, normalizeResponseNewLines, undefined, task);
```

It's a little more difficult to use the API having to know the positional places to specify `undefined`.

Perhaps a prototype like something like this would help API users:
```
    public performRest(resource: string, request: HTTP_VERB, reqHeaders?: any[], 
                       options: {writeData?: any,
                       responseStream?: Writable, requestStream?: Readable,
                       normalizeResponseNewLines?: boolean, normalizeRequestNewLines?: boolean,
                       task?: ITaskWithStatus
                       })
```

So the above example would look like:
```
        return new this(session).performRest(resource, HTTP_VERB.GET, reqHeaders, { responseStream,
            normalizeResponseNewLines, task});
```

dkelosky:
As a PoC, I have:
```
    /**
     * Overload function to perform the actual http REST call with appropriate user input using an option object (following angular's HTTPClient)
     * @param {IRestOptions} options
     * @memberof AbstractRestClient
     */
    public request(opts: IRestOptions) {
```

With, `IRestOptions` being:
```
/*
* This program and the accompanying materials are made available under the terms of the
* Eclipse Public License v2.0 which accompanies this distribution, and is available at
* https://www.eclipse.org/legal/epl-v20.html
*
* SPDX-License-Identifier: EPL-2.0
*
* Copyright Contributors to the Zowe Project.
*
*/
import { HTTP_VERB } from "../types/HTTPVerb";
import { Writable, Readable } from "stream";
import { ITaskWithStatus } from "../../../../operations";

/**
 * Interface to performRest function
 * @export
 * @interface IRestOptions
 */
export interface IRestOptions {

    /**
     * URI for this request
     * @type {string}
     * @memberof IRestOptions
     */
    resource: string;

    /**
     * REST request type GET|PUT|POST|DELETE
     * @type {HTTP_VERB}
     * @memberof IRestOptions
     */
    request: HTTP_VERB;

    /**
     * Headers to include with request
     * @type {any[]}
     * @memberof IRestOptions
     */
    reqHeaders?: any[];

    /**
     * Data to write on this REST request
[...]

PR #348, opened 2020-01-13
Add new rest client functions

Alexandru-Dumitru:
- add generic restClient methods that return full `client` object as default or a subset of it based on defined query
    - these are needed to help with addressing https://github.com/zowe/zowe-cli/issues/596 (porting ETag functionality into master)
- proposal to switch to `performRestCall()` instead of `performRest()`, thus addressing https://github.com/zowe/imperative/issues/316
    - improved option handling compared to current implementation
- preemptively marked `performRest()` as @deprecated (in which version is up to discussion) and made it as a wrapper to `performRestCall()`


packages/rest/src/client/RestClient.ts
```diff
@@ -16,6 +16,10 @@ import { AbstractRestClient } from "./AbstractRestClient";
 import { JSONUtils } from "../../../utilities";
 import { Readable, Writable } from "stream";
 import { ITaskWithStatus } from "../../../operations";
+import { IFullResponseOptions } from "./doc/IFullResponseOptions";
+import { IRestClientResponse } from "./doc/IRestClientResponse";
+import { IOptionsFullRequest } from "./doc/IOptionsFullRequest";
+import { CLIENT_PROPERTY } from "./types/AbstractRestClientProperties";
```
dkelosky: Would you be up for updating the remaining functions so that they don't call the recently deprecated `performRest` function?

packages/rest/src/client/AbstractRestClient.ts
```diff
@@ -196,24 +197,13 @@ export abstract class AbstractRestClient {
...
+     * @returns {Promise<string>}
      * @throws  if the request gets a status code outside of the 200 range
      *          or other connection problems occur (e.g. connection refused)
+     * @memberof AbstractRestClient
      */
-    public performRest(resource: string, request: HTTP_VERB, reqHeaders?: any[], writeData?: any,
-                       responseStream?: Writable, requestStream?: Readable,
-                       normalizeResponseNewLines?: boolean, normalizeRequestNewLines?: boolean,
-                       task?: ITaskWithStatus): Promise<string> {
+    public performRestCall(options: IRestOptions): Promise<string> {
```
dkelosky: This is pure opinion, but what about a shorter function name for the go-forward REST function?  For example, you had `request` or perhaps `call` or `send` or something else?

packages/rest/src/client/doc/IRestClientResponse.ts
```diff
@@ -0,0 +1,33 @@
...
+*
+*/
+
+import { Writable, Readable } from "stream";
+import { ITaskWithStatus } from "../../../../operations";
+import { Session } from "../../session/Session";
+import { Logger } from "../../../../logger";
+
+export interface IRestClientResponse {
+    requestSuccess?: boolean;
```
dkelosky: since it's a new interface, do you want to add all that typedoc nonsense :) ?

packages/rest/src/client/doc/IRestClientResponse.ts
```diff
@@ -0,0 +1,33 @@
...
+    requestFailure?: boolean;
+    data?: Buffer;
+    dataString?: string;
+    response?: any;
+    session?: Session;
+    log?: Logger;
+}
+
+// requestFailure: "requestFailure" as CLIENT_PROPERTY,
```
dkelosky: perhaps these old comments can be deleted

packages/rest/src/client/types/AbstractRestClientProperties.ts
```diff
@@ -0,0 +1,22 @@
...
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+export type CLIENT_PROPERTY = "requestSuccess" | "requestFailure" | "data" | "dataString" | "response" | "session" | "log";
+export const CLIENT_PROPERTY = {
+    requestSuccess: "requestSuccess" as CLIENT_PROPERTY,
```
dkelosky: ditto to above on adding typedoc

packages/rest/src/client/RestClient.ts
Alexandru-Dumitru: Sure I can update them.

packages/rest/src/client/AbstractRestClient.ts
Alexandru-Dumitru: Yeah, I thought `request` was not specific enough, but I was thinking from a substantive perspective, not a verb perspective 😛 . I will switch it back to `request`.

packages/rest/src/client/doc/IRestClientResponse.ts
Alexandru-Dumitru: I guess it makes sense. Will do that

packages/rest/src/client/doc/IRestClientResponse.ts
Alexandru-Dumitru: Ah, I thought I got them all. Thanks for catching this.

packages/rest/src/client/types/AbstractRestClientProperties.ts
Alexandru-Dumitru: done with a2f090a

packages/rest/src/client/doc/IRestClientResponse.ts
Alexandru-Dumitru: done with a2f090a

packages/rest/src/client/doc/IRestClientResponse.ts
Alexandru-Dumitru: done with a2f090a

packages/rest/src/client/AbstractRestClient.ts
Alexandru-Dumitru: done with a2f090a

packages/rest/__tests__/client/AbstractRestClient.test.ts
```diff
@@ -482,4 +484,114 @@ describe("AbstractRestClient tests", () => {
...
+
+        const restOptions: IOptionsFullResponse = {
+            resource: "/resource",
+        };
+
+        // create list of all properties
+        const listOfClientProperties = Object.keys(CLIENT_PROPERTY);
+
+        const data = await RestClient.getExpectFullResponse(new Session({hostname: "test"}), restOptions);
+        listOfClientProperties.forEach((property) => expect(data[`${property}`]).toBeDefined());
```
zFernand0: Should we look for the actual data instead of just checking that the given properties are defined?
No change required!

Alexandru-Dumitru: Yes, we can have another expect to check the data :) The property check is there to make sure it matches the interface.

zFernand0: No need to add it though, 
This serves the purpose of verifying that the right things should be defined? :smile:

Alexandru-Dumitru: I just remembered why I took out  `expect(data).toMatchSnapshot();` from those tests. The reason is that the snapshot that is generated on my machine differs a bit from the one generated when running in Jenkins (for some reason..), and the tests are failing because of that :(

zFernand0: no worries : )

zFernand0:
coverage was already below 80% on that one file.
Overall still above 80. 
Merging

Issue #350, opened 2020-01-15
Line contains condition twice.


ishche:
This line doesn't look right. Should it be chacking user name and password?
https://github.com/zowe/imperative/blob/df4573345fdee5c05e9d6e57b0f704856bd56a74/packages/rest/src/session/AbstractSession.ts#L309

PR #360, opened 2020-02-25
Fixes for Line contains condition twice

apsychogirl:
Fixes - #350

image

Lemme know if this needs to be changed!


tjohnsonBCM:
Closing because the same fix is included in #362

Issue #356, opened 2020-01-21
Export IRestOptions type


tjohnsonBCM:
The type `IRestOptions` which was added in #348 should be exported, so that it can be imported by CLI plugins that use Imperative.

This should be a simple change, it only requires adding a line in `packages/rest/index.ts`.

Alexandru-Dumitru:
Missed that detail. My bad ☹️

apsychogirl:
Hi, Can I work on this issue?

zFernand0:
Hey @apsychogirl, Feel free to assign yourself to the issue and open a PR once ready to contribute 😄

PR #363, opened 2020-02-25
Exported IRestOptions in packages/rest/index.ts

apsychogirl:
Fixes - #356 
image

Amended line 14


tjohnsonBCM:
Closing because the same fix is included in #362

PR #365, opened 2020-03-04
Don't load secure profile fields when deleting profile

tjohnsonBCM:
This is part of the fix for the issue https://github.com/zowe/zowe-cli/issues/665. (The other part of the fix is the PR https://github.com/zowe/zowe-cli-scs-plugin/pull/7.)

* Sets "noSecure = true" during loading of all profiles in `AbstractProfileManager.delete` to check for profile dependencies. This prevents fields that may or may not be secure from being loaded unnecessarily.
* Adds `ILoadAllProfiles` type, which was referenced in a comment but not defined
* Redefines `ICliLoadAllProfiles` and `ICliLoadProfile` types to mirror the types `ILoadAllProfiles` and `ILoadProfile`, since they no longer have additional properties of their own.

### DIFF
--- a/packages/cmd/src/doc/profiles/parms/ICliLoadProfile.ts
+++ b/packages/cmd/src/doc/profiles/parms/ICliLoadProfile.ts
@@ -11,12 +11,23 @@
 
 import { ILoadProfile } from "../../../../../profiles";
 
-export interface ICliILoadProfile extends ILoadProfile {
-    /**
-     * If true, fields that indicate "secure" are not loaded. The properties will still be present in the profiles
-     * loaded with a value of "securely_stored".
-     * @type {boolean}
-     * @memberof ILoadProfile
-     */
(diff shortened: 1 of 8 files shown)


packages/cmd/src/doc/profiles/parms/ICliLoadProfile.ts
```diff
@@ -11,12 +11,23 @@
...
-     * @type {boolean}
-     * @memberof ILoadProfile
-     */
-    noSecure?: boolean;
-}
+/**
+ * Profile Manager "loadProfile" input parameters. Indicates which profile to load (named or default) and if
+ * not finding the profile should be considered and error, etc.
+ * @export
+ * @deprecated - Use ICliLoadProfile instead because it is spelled correctly
```
zFernand0: Thanks for deprecating.
We should have this in the Changelog???

Issue #364, opened 2020-03-02
Optional secure profile fields fail to load


ChrisBoehmCA:
ssh profiles in zowe CLI have an optional field keyPassphrase which is marked as secure: true. If the user does not enter keyPassphrase when creating a profile, the value in the profile for keyPassphrase is `null`, but the SCS plugin still attempts to load it from the vault.

Reproduction:
Create an ssh profile with the SCS plugin installed with no keyPassphrase, then attempt to use it on a zowe uss issue ssh "ls" command

tjohnsonBCM:
This won't be fixed in the SCS plugin which was mentioned in the issue until https://github.com/zowe/zowe-cli-scs-plugin/pull/11 is merged.

tjohnsonBCM:
Fixed in https://github.com/zowe/zowe-cli-scs-plugin/pull/11

PR #366, opened 2020-03-11
Fix error when optional secure fields in profile are missing

tjohnsonBCM:
This is part of the fix for the issue https://github.com/zowe/imperative/issues/364. (The other part of the fix is the PR https://github.com/zowe/zowe-cli-scs-plugin/pull/11.)

Adds a boolean param `optional` to methods that load credentials. If set to true, the secure field being loaded is considered optional, meaning that failure to load it will be ignored.

### DIFF
--- a/packages/cmd/src/profiles/CliProfileManager.ts
+++ b/packages/cmd/src/profiles/CliProfileManager.ts
@@ -201,7 +201,7 @@ export class CliProfileManager extends BasicProfileManager<ICommandProfileTypeCo
              * @param {string} propertyNamePath - The path to the property
              * @return {Promise<string>}
              */
-            securelyLoadValue = async (propertyNamePath: string): Promise<any> => {
+            securelyLoadValue = async (propertyNamePath: string, _: any, optional?: boolean): Promise<any> => {
                 let ret;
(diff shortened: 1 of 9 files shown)


packages/cmd/src/profiles/CliProfileManager.ts
```diff
@@ -353,7 +354,7 @@ export class CliProfileManager extends BasicProfileManager<ICommandProfileTypeCo
             this.log.debug("Setting profile field %s from command line option %s", propNamePath, optionName);
             if (secureOp && prop.secure) {
                 this.log.debug("Performing secure operation on property %s", propNamePath);
-                return secureOp(propNamePath, propValue);
+                return secureOp(propNamePath, propValue, !prop.optionDefinition.required);
```
tjohnsonBCM: Should I add a test around the addition on this line? Codecov says it's already covered, but I'm not sure how.

zFernand0: Sorry for not saying anything earlier. 
the line coverage won't be affected, but the branch coverage is. No real concern with adding tests now or later :)

PR #372, opened 2020-03-25
fix: defaultValue with falsy values not displayed

liuxh0:
Generated help doesn't include text for default value in case of falsy
values like 0 or false.


packages/cmd/src/help/abstract/AbstractHelpGenerator.ts
```diff
@@ -173,7 +173,7 @@ export abstract class AbstractHelpGenerator implements IHelpGenerator {
 
             // build the option help text
             let optionText = option.description;
-            const defaultValueText = option.defaultValue ? this.grey("\nDefault value: " + option.defaultValue) : "";
+            const defaultValueText = option.defaultValue !== undefined ? this.grey("\nDefault value: " + option.defaultValue) : "";
```
tjohnsonBCM: Thanks for fixing this for cases where `defaultValue = false`.

The strict check for undefined means that `defaultValue = null` would be displayed in help content. Since `null` and `undefined` should both be invalid for `defaultValue`, could you change the check to:
```typescript
option.defaultValue != null
```

liuxh0: Hi @tjohnsonBCM ,
now I have a new commit. It checks both null and undefined.

zFernand0:
@liuxh0, could you update the changelog with your recent changes?

liuxh0:
@zFernand0 
Of course I would do. Should I update under version 4.5.4?

zFernand0:
> Of course I would do. Should I update under version 4.5.4?

I believe you want to have a new entry in the changelog.
Like in here: https://github.com/zowe/imperative/blob/master/CONTRIBUTING.md#changelog-update-guidelines

liuxh0:
I've updated changelog. Is this okay?

zFernand0:
I believe the header needs to match with the Jenkinsfile specification.

https://github.com/zowe/imperative/blob/master/Jenkinsfile#L175


@awharn, should we handle this in the library?

liuxh0:
@zFernand0 
Sorry it's my bad. I wrote it wrong.

zFernand0:
No worries, I believe we could try to address this kind of situations in the library themselves :smile:

zFernand0:
> No worries, I believe we could try to address this kind of situations in the library themselves 😄

After some discussion we noticed that we recent changes should always be a second-level header (i.e. `## Recent Changes`) because `# Changelog` already is the top level header. 😄 

@liuxh0, thanks for addressing so quickly.

zFernand0:
Sorry for not noticing sooner, but it'd be great if we could link this to an issue 😄

PR #380, opened 2020-03-30
feat: allowable values match whole input

liuxh0:
Regular expressions defined for allowable values are wrapped with ^ and $ by default in order to match the whole input. Exception is when user already uses ^ or $ in definition.

In my opinion this is a breaking change, although it only affects very few cases.

See idea and discussion in zowe/zowe-cli#692

### DIFF
--- a/packages/cmd/src/syntax/SyntaxValidator.ts
+++ b/packages/cmd/src/syntax/SyntaxValidator.ts
@@ -298,13 +298,17 @@ export class SyntaxValidator {
                     const optionDefCopy: ICommandOptionDefinition = JSON.parse(JSON.stringify(optionDef));
                     // Use modified regular expressions for allowable values to check and to generate error
                     optionDefCopy.allowableValues.values = optionDef.allowableValues.values.map((regex) => {
-                        // If user-defined regex already contains ^ and/or $,
-                        // we assume user is aware of the usage and thus keep user's definition.
-                        if (regex.startsWith("^") || regex.endsWith("$")) {
-                            return regex;
+                        // Prepend "^" if not existing
+                        if (!regex.startsWith("^")) {
+                            regex = "^" + regex;
                         }
-                        // Otherwise wrap the regex in order to match the whole input
-                        return `^${regex}$`;
+
+                        // Append "$" if the last char is an escaped "$" or chars other than "$"
+                        if (!regex.endsWith("$") || regex.endsWith("\\$")) {
+                            regex = regex + "$";
+                        }
+
+                        return regex;
                     });
                     if (!this.checkIfAllowable(optionDefCopy.allowableValues, commandArguments[optionName])) {
                         this.invalidOptionError(optionDefCopy, responseObject, commandArguments[optionName]);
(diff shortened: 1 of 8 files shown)


packages/cmd/src/syntax/SyntaxValidator.ts
```diff
@@ -294,8 +294,20 @@ export class SyntaxValidator {
...
                 if (!isNullOrUndefined(optionDef.allowableValues)) {
-                    if (!this.checkIfAllowable(optionDef.allowableValues, commandArguments[optionName])) {
-                        this.invalidOptionError(optionDef, responseObject, commandArguments[optionName]);
+                    // Make a copy of optionDef, so that modifications below are only used in this place
+                    const optionDefCopy: ICommandOptionDefinition = JSON.parse(JSON.stringify(optionDef));
+                    // Use modified regular expressions for allowable values to check and to generate error
+                    optionDefCopy.allowableValues.values = optionDef.allowableValues.values.map((regex) => {
+                        // If user-defined regex already contains ^ and/or $,
+                        // we assume user is aware of the usage and thus keep user's definition.
+                        if (regex.startsWith("^") || regex.endsWith("$")) {
```
tjohnsonBCM, line 303: Not sure if this assumption is safe. It may lead to unexpected behavior if users want an exact match for strings containing these characters. For example, if there is a CLI option for currency symbols, `\$` could be an allowable value regex that should not be treated as special.

liuxh0: Yes you are right. The check `regex.endsWith("$")` is not correct. It should be `regex.endsWith("$") && !regex.endsWith("\\$")`.

But as you said in the last comment that we will treat this as a bugfix, and thus non-breaking change, I will always add ^ and $ to match exactly.

liuxh0:
Don't know why unit tests fail. All unit tests pass on my machine.

tjohnsonBCM:
> Don't know why unit tests fail. All unit tests pass on my machine.

There are differences between Node.js v13 and older versions in how output of the `util.inspect` function is formatted. This was causing the snapshot difference because the Jenkins pipeline uses Node v10. I've fixed this in master and updated your branch so it should pass now 😄

liuxh0:
Thanks.
Indeed I use Node v13.12.0. Good to know.

packages/cmd/src/syntax/SyntaxValidator.ts
```diff
@@ -294,8 +294,24 @@ export class SyntaxValidator {
...
+                    const optionDefCopy: ICommandOptionDefinition = JSON.parse(JSON.stringify(optionDef));
+                    // Use modified regular expressions for allowable values to check and to generate error
+                    optionDefCopy.allowableValues.values = optionDef.allowableValues.values.map((regex) => {
+                        // Prepend "^" if not existing
+                        if (!regex.startsWith("^")) {
+                            regex = "^" + regex;
+                        }
+
+                        // Append "$" if the last char is an escaped "$" or chars other than "$"
+                        if (!regex.endsWith("$") || regex.endsWith("\\$")) {
```
zFernand0, line 307: Could we maybe add a test around this (or copy an existing one and modify the data)? i.e. pass the following string: `^abc\\$$` 😄

zFernand0: And maybe don't rely on snapshots for a regular string. (`abc`) :smile:

liuxh0: Now I have added more tests on allowable values. Have a look and please let me know if there's something I can help.

zFernand0: Awesome! 
Clever way of adding more permutations.

**Before:**
image

**After:**
image

PR #382, opened 2020-04-03
Tokens branch to be merged into "next"

gejohnston:
The purpose of this PR is to enable a next snapshot to be published which contains initial token support. The CLI repository can then use the published next version of imperative to successfully build.

A separate "tokens" branch was created to isolate token development.
Unit tests and integration tests were corrected as needed.
Latest updates from master was merged into tokens.
The next branch was then also merged into tokens to ensure a clean PR.


Jenkinsfile
```diff
@@ -25,8 +25,7 @@ node('ca-jenkins-agent') {
 
     // Protected branch property definitions
     pipeline.protectedBranches.addMap([
-        [name: "master", tag: "latest", dependencies: ["@zowe/perf-timing": "latest"]],
-        [name: "next", tag: "next", prerelease: "next", dependencies: ["@zowe/perf-timing": "latest"]],
```
tjohnsonBCM, line 29: I think we want to keep this line in the Jenkinsfile for the `next` branch.

zFernand0:
I do have a few reservations on this.
Up until e99e179, there were a lot of additions, but after updating from master and next, all changes will get combined and the only thing this PR will be representing is an update to the `next` branch from `master` :cry:

dkelosky:
I looked this over, but for the files changed, I didn't see anything directly related to tokens.  No other feedback from me.

gejohnston:
@zFernand0, I am not clear about the tracking information that will be lost.

How do we want to proceed from where this currently stands?

zFernand0:
> @zFernand0, I am not clear about the tracking information that will be lost.
> 
> How do we want to proceed from where this currently stands?

Does this PR (#383) look closer to the changes you (@gejohnston) were making and to the final state we (all) believe `next` should be in? 

If so, I think we can just go with that (after we evaluate the differences) : )
I haven't looked at all the changes closely to say I'm comfortable with all there is. Sorry :cry:

PR #383, opened 2020-04-06
[DO NOT MERGE] Test PR

zFernand0:
please disregard the massive deletion. 
That's mostly the package-lock files contained in the Test CLIs that we use for integration test.

**This is a replica of #382 right before updating from master and next.**


zFernand0:
closing PR for now

PR #387, opened 2020-04-08
Add Bearer type for Session in rest

YuanboXue-Amber:
`ISession.type` originally has these 3 value `"none" | "basic" | "token"`
I add "bearer".

However, I think `token` here should be changed into `cookie` (also `AbstractSession.TYPE_TOKEN` to `AbstractSession.TYPE_COOKIE`). But that would probably break someone who is using it.

This is my first time making changes to imperative. Open to all critics and suggestions :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Add Bearer token in rest Session
+
 ## `4.5.6`
 
 - Fix allowable values not exactly matching input
--- a/packages/rest/__tests__/client/AbstractRestClient.test.ts
+++ b/packages/rest/__tests__/client/AbstractRestClient.test.ts
@@ -594,4 +594,38 @@ describe("AbstractRestClient tests", () => {
         restOptions.dataToReturn.forEach((property) => expect(data[`${property}`]).toBeDefined());
         listOfClientProperties.forEach((property) => expect(data[`${property}`]).not.toBeDefined());
     });
+
+    it("should create buildOptions according to input parameter options", async () => {
+
+        const httpsRequestFnc = jest.fn((options, callback) => {
+            expect(options).toMatchSnapshot();
+            const emitter = new MockHttpRequestResponse();
+            ProcessUtils.nextTick(() => {
+                callback(new EventEmitter());
+                ProcessUtils.nextTick(() => {
+                    emitter.emit("error", "value");
+                });
+            });
+            return emitter;
+        });
+
+        (https.request as any) = httpsRequestFnc;
+
+        let error;
+        try {
+            await RestClient.getExpectString(
+                new Session({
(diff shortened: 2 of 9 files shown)


dkelosky:
I suppose the sonar stuff and build problems need sorted out before this can be merged.

I tested out your code with:
```
import { Session, RestClient } from "@zowe/imperative";

(async () => {

    const sess = new Session({
        hostname: "some.apiml.host",
        port: 60454,
        protocol: "https",
        type: "bearer",
        tokenValue: "...long-token-here..."
    });

    try {

        const req = await new RestClient(sess).request({
            request: "GET",
            resource: "/api/v1/gateway/auth/query"
        });
        console.log(JSON.stringify(req, null, 2));

    } catch (err) {
        console.log(err)
    }

})();

```

And got a proper JSON response.

YuanboXue-Amber:
> I suppose the sonar stuff and build problems need sorted out before this can be merged.
> 
> I tested out your code with:
> 
> ```
> import { Session, RestClient } from "@zowe/imperative";
> 
> (async () => {
> 
>     const sess = new Session({
>         hostname: "some.apiml.host",
>         port: 60454,
>         protocol: "https",
>         type: "bearer",
>         tokenValue: "...long-token-here..."
>     });
> 
>     try {
> 
>         const req = await new RestClient(sess).request({
>             request: "GET",
>             resource: "/api/v1/gateway/auth/query"
>         });
>         console.log(JSON.stringify(req, null, 2));
> 
>     } catch (err) {
>         console.log(err)
>     }
> 
> })();
> ```
> 
> And got a proper JSON response.

Thanks for checking.
- Build failed because of changelog. Should I go with 4.5.7 or 4.6.0?

tjohnsonBCM:
> * Build failed because of changelog. Should I go with 4.5.7 or 4.6.0?
Please use the header "Recent Changes" in the changelog rather than a specific version number. 😄 
When the PR is merged into master it will automatically be updated to the appropriate version number. More details about how the changelog gets updated are [here](https://github.com/zowe/imperative/blob/master/CONTRIBUTING.md#changelog-update-guidelines).

YuanboXue-Amber:
Aha I didn't read the guideline properly :P 

I updated changelog and also added more tests to make sonar happy. And put everything in 1 commit.

dkelosky:
> However, I think token here should be changed into cookie (also AbstractSession.TYPE_TOKEN to AbstractSession.TYPE_COOKIE). But that would probably break someone who is using it.

I think you're right about this - perhaps for a future version we can fix this.

zFernand0:
> > However, I think token here should be changed into cookie (also AbstractSession.TYPE_TOKEN to AbstractSession.TYPE_COOKIE). But that would probably break someone who is using it.
> 
> I think you're right about this - perhaps for a future version we can fix this.

Should this be captured in a separate issue?

YuanboXue-Amber:
Thank you guys for this chance :)

Issue #357, opened 2020-01-21
Allow ImperativeConfig properties to be accessed before Imperative has been initialized


ghost:
Currently, some properties of the `ImperativeConfig` class like `cliHome` cannot be accessed until `Imperative.init` has been called. These properties should be accessible without needing to wait to initialize the Imperative command tree.

mheuzey:
If this change goes through, these properties could be used to create default values for some functions. For example, most people probably leave their profiles in the default directory, so `CliProfileManager` could use `path.join(ImperativeConfig.instance.cliHome, "profiles")` as a default value for `profileRootDirectory`. Then you would only need to specify the profile type when calling the constructor.

JTonda:
This enhancement has had no activity for 12 months. The issue also has less than 5 up-votes by the community. No action on this enhancement is targeted for the next 2 calendar quarters. Therefore, this enhancement is being closed. If you feel that this enhancement should continue to be available for community up-votes, you may reopen this issue.

PR #399, opened 2020-05-12
Tokens

gejohnston:
Implemented the infrastructure to enable CLI commands to process tokens.

- Modified AbstractSession.buildSession() to use a token for authentication if a token is present in the underlying REST session configuration object.
- Added a new CredsForSessCfg.addCredsOrPrompt function that places credentials (including a possible token) into a session configuration object.
    - Credentials are obtained from the command line, environment variables, or a profile.
    - If no credentials are available, it will prompt for a user name and password.
    - Any prompt will timeout after 30 seconds so that it will not hang an automated script.
- All commands will automatically benefit from token handling.
- Only commands that convert to CredsForSessCfg.addCredsOrPrompt() will benefit from prompting for user name and password.


packages/rest/src/session/AbstractSession.ts
```diff
@@ -256,86 +235,107 @@ export abstract class AbstractSession {
     private buildSession(session: ISession): ISession {
         const populatedSession = session;
 
-
         // set protocol if not set
-        if (isNullOrUndefined(populatedSession.protocol)) {
+        if (populatedSession.protocol === undefined || populatedSession.protocol === null) {
```
tjohnsonBCM, line 239: A shorter replacement for `isNullOrUndefined(x)` is `x == null` (only 2 equal signs, so effectively handles both cases of "is null or undefined"). Not sure if this is important as `x === null || x === undefined` should be functionally identical. I only mention it because the suggestion of `x == null` has made to me on other PRs.

gejohnston: I used what I did because that is what NodeJS recommends. I admit that does not make it the right answer.

https://nodejs.org/api/util.html#util_util_isnullorundefined_object

packages/rest/index.ts
```diff
@@ -19,6 +19,13 @@ export * from "./src/client/RestConstants";
 export * from "./src/client/RestStandAloneUtils";
 export * from "./src/client/RestClientError";
 
+/* The following line is valid, but tslint barfs on it.
+ * So, we exclude this file in imperative/tslint-packages.json.
+ */
+export * as SessConstants from "./src/session/SessConstants";
```
dkelosky, line 25: Not second guessing, just curious, do we often use "export ... as" in these files instead of "export * from"?

packages/rest/src/session/CredsForSessCfg.ts
```diff
@@ -0,0 +1,207 @@
...
+             */
+            if (optsToUse.requestToken) {
+                // Set our type to token to get a token from user and pass
+                impLogger.debug("Using basic authentication to get token");
+                finalSessCfg.type = SessConstants.AUTH_TYPE_TOKEN;
+                if (tokenTypeExists) {
+                    finalSessCfg.tokenType = cmdArgs.tokenType;
+                } else {
+                    // When no tokenType supplied, user wants bearer
+                    finalSessCfg.tokenType = SessConstants.TOKEN_TYPE_LTPA;  // TODO:Gene: change to TOKEN_TYPE_APIML
```
tjohnsonBCM, line 142: Should this TODO be removed?

packages/rest/src/session/SessConstants.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * tokenType property value for a JWT token
+ */
+export const TOKEN_TYPE_JWT = "jwtToken";
+
+/**
+ * tokenType property value for an API Mediation Layer token
+ */
+export const TOKEN_TYPE_APIML = "apimlAuthenticationToken";
```
tjohnsonBCM, line 74: As Imperative is a generic CLI framework, do we want to avoid code specific to the APIML? If so, is this an exception to that? Seems like it would be difficult to avoid in this case, since we need the Imperative session to recognize `TOKEN_TYPE_APIML` as a valid token type when used by Zowe CLI and plugins.

packages/utilities/src/CliUtils.ts
```diff
@@ -403,6 +402,108 @@ export class CliUtils {
...
+
+        // ask user the desired question and then asynchronously read answer
+        ttyIo.prompt();
+        let answerToReturn: string = null;
+        ttyIo.on("line", (answer: string) => {
+            answerToReturn = answer;
+            ttyIo.close();
+        }).on("close", () => {
+            if (hideText) {
+                // The user's Enter key was echoed as a '*', so now output a newline
```
tjohnsonBCM, line 464: Is the Enter key echoed as a `*`? If so, should we disable that behavior for the Enter key in the `_writeToOutput` method below?

When entering a password, if its short (like mainframe passwords which are limited to 8 chars), I pay attention to the number of `*` characters. If I saw 9 instead of 8, I could be confused and think I had entered my password wrong. Maybe this is just me being weird, not sure 😄

Jenkinsfile
```diff
@@ -26,6 +26,7 @@ node('ca-jenkins-agent') {
     // Protected branch property definitions
     pipeline.protectedBranches.addMap([
         [name: "master", tag: "latest", dependencies: ["@zowe/perf-timing": "latest"], aliasTags: ["zowe-v1-lts"]],
+        [name: "next", tag: "next", prerelease: "next", dependencies: ["@zowe/perf-timing": "latest"]],
```
tjohnsonBCM, line 29: Should this be removed if the PR is being merged into master?

packages/cmd/src/profiles/CommandProfiles.ts
```diff
@@ -110,6 +176,17 @@ export class CommandProfiles {
...
     }
 
+    /**
+     * Internal accessor for the meta map
+     * @readonly
+     * @private
+     * @type {Map<string, IProfileLoaded[]>} - The profile Map
+     * @memberof CommandProfiles
+     */
+    private get metaMap(): Map<string, IProfileLoaded[]> {
```
tjohnsonBCM, line 186: Curious what is the purpose of the profile `metaMap`? It seems to map profiles based on their type, but I don't see it referenced anywhere in this branch or the `initial-token-support` branch of Zowe CLI.

packages/rest/src/session/SessConstants.ts
dkelosky: Yes - I think you're correct @tjohnsonBCM

packages/rest/src/session/AbstractSession.ts
zFernand0: Interesting that the link you provided says that they cannot guarantee backwards compatibility and that it has been deprecated since `v4.0.0`.

Should we open an issue so that we change all of these in a future PR?

packages/rest/index.ts
zFernand0: I'm afraid I don't see other index files having something exported with a given name. 
If we change it to `export * from`, it can still be imported as `sessConstants` if desired whenever it's used.
Should we try to avoid giving names to export since it could be a beaking change if we change (for example) from `SessConstants` to `sessConstants` in the future?

dkelosky: perhaps, and perhaps this is the cause of the linting failure as well

packages/rest/src/session/AbstractSession.ts
```diff
@@ -256,86 +235,107 @@ export abstract class AbstractSession {
...
+            if (session.user     !== undefined && session.user     !== null &&
+                session.password !== undefined && session.password !== null)
+            {
+                // ok
+            } else if (session.base64EncodedAuth !== undefined && session.base64EncodedAuth !== null) {
+                // ok
+            } else {
+                throw new ImperativeError({
+                    msg: "Must have user & password OR base64 encoded credentials",
+                    additionalDetails: "For CLI usage, see 'zowe zosmf login --help'"
```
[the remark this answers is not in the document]
zFernand0: Good point on using the rootCommandName.
Another question: Should we have this `zosmf login` command reference here? Or can it be changed to `${this.config.rootCommandName} auth login` ?

packages/cmd/src/profiles/CommandProfileLoader.ts
```diff
@@ -146,6 +153,30 @@ export class CommandProfileLoader {
...
+     * Builds the command meta map for input the the command map object for the command handlers
+     * @private
+     * @param {IProfileLoaded[]} responses - The full list of profiles loaded for this command
+     * @param {Map<string, IProfile[]>} map - The meta map to populate
+     * @memberof CommandProfileLoader
+     */
+    private buildCommandMetaMap(responses: IProfileLoaded[], map: Map<string, IProfileLoaded[]>) {
+        for (const resp of responses) {
+            if (resp.profile) {
+                if (isNullOrUndefined(map.get(resp.type))) {
```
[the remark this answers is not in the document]
zFernand0: Similar comment as before.
Maybe we should open an issue to address/replace any deprecated functions in `NodeJS < 8`

packages/rest/src/session/AbstractSession.ts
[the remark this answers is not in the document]
dkelosky: Good points - both were my oversight

Jenkinsfile
zFernand0: It won't affect its behavior.
I'm fine either way 😋

packages/utilities/src/CliUtils.ts
zFernand0: I believe there is a security concern if the masking of your password matches your password length.
So, it should be fine if we output 9 (or 10) `*` even if there is a restriction.

packages/cmd/src/profiles/CommandProfiles.ts
zFernand0: This allows for things like `this.metaMap.get(type)[0];` due to the capabilities of the `getter`. However, not sure if we really need one if it's going to be `private` to the class

packages/cmd/src/profiles/CommandProfiles.ts
```diff
@@ -67,6 +131,8 @@ export class CommandProfiles {
         if (this.map.get(type) != null) {
             if (name != null && name.trim().length > 0) {
                 for (const prof of this.map.get(type)) {
+
+                    // TODO(Kelosky): is this in error if the name is not contained in IProfile?
```
zFernand0, line 135: I like the `Kelosky TODOs`, but should we try to handle it (in case it is *in* error)?

packages/rest/__tests__/session/CredsForSessCfg.test.ts
```diff
@@ -0,0 +1,211 @@
...
+            port: 11,
+            rejectUnauthorized: true,
+        };
+        const args = {
+            user: "FakeUser",
+        };
+
+        /* Jest hangs forever when you request data from stdin.
+         * mock-stdin should let you supply data. We could not make that work.
+         * At least with mock-stdin we only get our expected 30 second timout.
```
zFernand0, line 190: Wondering if it'd be possible to use the `jest.advanceTimerByTime(ms)` function
https://jestjs.io/docs/en/timer-mocks#advance-timers-by-time

We may prevent this single file from taking over a minute to run. 😋

packages/rest/src/session/CredsForSessCfg.ts
zFernand0: Similar TODOs in tests

packages/rest/src/session/doc/IOptionsForAddCreds.ts
```diff
@@ -0,0 +1,36 @@
...
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+/**
+ * Interface for options supplied to CredsForSessCfg.addCredsOrPrompt()
+ * @export
+ * @interface ISession
+ */
+export interface IOptionsForAddCreds {
```
zFernand0, line 17: Thank you for creating this interface 👍

tslint-packages.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tslint.json",
   "linterOptions": {
-    "exclude": ["**/__tests__/**/*"]
+    "exclude": ["**/__tests__/**/*", "packages/rest/index.ts"]
```
zFernand0, line 4: 😕 Could we try to resolve it some other way?

packages/cmd/src/profiles/CommandProfiles.ts
dkelosky: I added this comment because I think it was a but in the existing code base.  I didn't capture the steps to reproduce (which I should have).  I think perhaps these could be tracked as separate defects?

packages/cmd/src/profiles/CommandProfileLoader.ts
```diff
@@ -115,11 +116,17 @@ export class CommandProfileLoader {
             const responses: IProfileLoaded[] = await this.loadAll(loadList);
             this.log.debug(`"${responses.length}" profiles loaded.`);
             this.buildCommandMap(responses, profileMap);
+            this.buildCommandMetaMap(responses, profileMetaMap);
             this.log.trace(`All profiles loaded for command: ${this.definition.name}...`);
         }
 
+        // add meta
+        const cmdProfiles = new CommandProfiles(profileMap);
+        cmdProfiles.addMeta(profileMetaMap);
```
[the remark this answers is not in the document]
dkelosky: initially it was for backwards compatibility, but i could make it optional in the constructor right?

packages/cmd/src/profiles/CommandProfiles.ts
dkelosky: The `private` getter `metaMap` is used in the public `getMeta`.  `getMeta` is used in the future login handler (indirectly in the base handler).

I could remove this since it doesn't make sense.  I added it just to fit in the existing coding pattern of the file

packages/rest/src/session/AbstractSession.ts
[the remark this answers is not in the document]
dkelosky: Is it possible to relay on config from imperative here?  I am able to use Session (and AbstractSession) and RestClient without doing an init on imperative.

```ts
import { GetJobs } from "@zowe/cli";
import { Session, ISession, Logger, LoggingConfigurer } from "@zowe/imperative";

(async () => {
    // init dummy logger for imperative
    Logger.initLogger(LoggingConfigurer.configureLogger(".sample", { name: "sample" }));

    // connection info (host, user, password, etc...)
    const connectionInfo: ISession = {
        hostname: "some-host",
        user: "user",
        password: "pass",
        type: "basic",
        rejectUnauthorized: false
    };
    
    // create a session & get jobs
    const session = new Session(connectionInfo);
    const jobs = await GetJobs.getJobs(session);

    // print all obtained job names
    jobs.forEach((job) => {
        console.log(`Got job: ${job.jobname} ${job.status}`);
    });
})();
```

Should I make the error message more generic? `For CLI usage, see <CLI> login --help` ?

packages/cmd/src/profiles/CommandProfileLoader.ts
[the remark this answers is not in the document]
zFernand0: I was just curious. But thanks for implementing. 
I saw you didn't return the `new CommandProfiles` though 😋 
No big deal!

packages/rest/src/session/AbstractSession.ts
```diff
@@ -256,86 +235,107 @@ export abstract class AbstractSession {
...
+                if (session.user     !== undefined && session.user     !== null &&
+                    session.password !== undefined && session.password !== null)
+                {
+                    // ok
+                } else if (session.base64EncodedAuth !== undefined && session.base64EncodedAuth !== null) {
+                    // ok
+                } else {
+                    throw new ImperativeError({
+                        msg: "Must have user & password OR tokenType & tokenValue.",
+                        additionalDetails: "For CLI usage, see 'zowe zosmf login --help'"
```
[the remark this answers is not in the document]
dkelosky: same question as above on the hard coding - can we rely on having ImperativeConfig here?

packages/cmd/src/profiles/CommandProfileLoader.ts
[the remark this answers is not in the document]
dkelosky: 😞 ugh... ill fix that

tslint-packages.json
dkelosky: potentially the `export * from` will fix this.

packages/cmd/src/profiles/CommandProfiles.ts
dkelosky: do you have a preference on keeping internal accessor / getters or removing them?

packages/rest/src/session/AbstractSession.ts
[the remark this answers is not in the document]
dkelosky: as @tucker01 would say - "what do?"

packages/rest/src/session/SessConstants.ts
dkelosky: perhaps the same applies with LTPA

packages/rest/src/session/AbstractSession.ts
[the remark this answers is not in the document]
tjohnsonBCM: Good point, currently ImperativeConfig properties are inaccessible until `Imperative.init` has been run (#357).

packages/cmd/src/profiles/CommandProfiles.ts
zFernand0: I don't mind keeping it (to be consistent with the `coding pattern` of the file 😋

packages/rest/src/session/AbstractSession.ts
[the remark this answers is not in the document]
zFernand0: Apologies for not using words 😞 
I agree with Timothy, and also with making it more generic. 😋 
`<CLI> auth login <service> --help`

packages/rest/src/session/AbstractSession.ts
zFernand0: https://github.com/zowe/imperative/issues/400

packages/cmd/src/profiles/CommandProfileLoader.ts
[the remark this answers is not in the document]
zFernand0: https://github.com/zowe/imperative/issues/400

packages/rest/index.ts
gejohnston: We were using numerous duplicated literals, like "LtpaToken2" throughout the code. I wanted to create a single symbol that was used everywhere (like SessConstants.TOKEN_TYPE_LTPA). I also wanted the ISession interface to avoid using duplicated literals. Instead, specifying a type that can be any one of several symbols:

    type TOKEN_TYPE_CHOICES = typeof TOKEN_TYPE_LTPA | typeof TOKEN_TYPE_JWT |
             typeof TOKEN_TYPE_APIML

I found no way to export type definitions from a Class into the ISession interface. Thus SessConstants is not a class. Each symbol had to be exported individually. See comments at the beginning of SessConstants.ts.

Because the symbols are not within a class, if we use "export * from", we lose the namespace qualifier. Every symbol would have to be referenced just by the name, like TOKEN_TYPE_LTPA. By using "export * as SessConstants" we are able to use expressions like SessConstants.TOKEN_TYPE_LTPA, which I feel is very helpful in finding the value when reading code.

My research on the phrase "export * as SessConstants" within NPM identifies that it is valid syntax, and clearly it compiles and works correctly. Lint does not report that phrase as an error - lint just crashes. Thus, I decided that was a bug in lint.

Despite being different than the other ways that we export, I would lobby that the benefits outweigh the differences.

dkelosky: Sorry, I prematurely pushed changes to remove `export ... as` and reverted the lint rule - trying to knock off some of these comments.  

I could:
- revert my recent changes
- convert `SessConstants.ts` to `namespace` so it can be imported in `ISession.ts`
- leave my recent changes (we do export lots of symbols without a namespace e.g. `CoreMessages.ts`)

Side note: we use classes with all `public static readonly` for the purpose of `namespacing`  but is different than using the `namespace` keyword.  

Please let me know

gejohnston: At this point, I will settle for anything that compiles. Note, the CLI must also compile successfully. The current errors in the CLI are like this:

packages/rest/src/api/ZosmfRestClient.ts:12:74 - error TS2724: Module '"../../../../../imperative/lib"' has no exported member 'SessConstants'. Did you mean 'RestConstants'?                                   
                                                                                                        
12 import { IImperativeError, Logger, RestClient, TextUtils, RestConstants, SessConstants } from "@zowe/imperative";

dkelosky: @zFernand0 , @tjohnsonBCM  - any strong opinion?

dkelosky: "what do?" haha

dkelosky: i reverted the changes in the meantime so that @gejohnston can build.  whatever changes are recommended, i'll push both at the same time

zFernand0: Not a strong opinion, just a comment :yum:

So by doing `export * as SessConstants from <filepath>` we can then import it like `import { SessContants } from <package>`? 

If that's the behavior we want, and not having to do `import * as SessConstants from <package>/file/path` (which we don't usually see very often), I think the "consistent" way is to create a "namespace" (a class in out case) to export the constants that we want out of it.

packages/cmd/src/profiles/CommandProfiles.ts
dkelosky: ok perfect

packages/rest/src/session/AbstractSession.ts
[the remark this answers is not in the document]
dkelosky: fixed

packages/rest/index.ts
dkelosky: leaving @gejohnston  implementation

packages/rest/src/session/SessConstants.ts
gejohnston: I would think of this as the set of token types that Imperative recognizes, including ones that we use, and ones that we do not use. As others are needed by the community, they could be added to the list.

packages/utilities/src/CliUtils.ts
gejohnston: I originally tried to test for the Enter key value within the _writeToOutput method. I was unable to get the test to work. Perhaps I did something wrong, but after enough attempts, I just gave up.

Jenkinsfile
gejohnston: This was just a carry-over from the original POC. I will just delete the "next" line now.

packages/rest/__tests__/session/CredsForSessCfg.test.ts
gejohnston: jest.advanceTimersByTime is meant to be run after an asynchronous operation has been started. However, we await our prompt for credentials making it synchronous. Calling advanceTimersByTime after the prompt is too late. I even tried calling advanceTimersByTime  before and after the prompt. The test hung until it hit a jest global timeout of 3 minutes.

Because you made me look at this problem again, I did find an alternative approach. I mock our CliUtils.sleep function:

    const sleepReal = CliUtils.sleep;
    CliUtils.sleep = jest.fn();

Now the entire test suite runs in about 1.5 seconds. Thanks for prodding me.

packages/rest/src/session/CredsForSessCfg.ts
gejohnston: The login command currently in the CLI generates zosmf tokens. If we change the default token value to APIML now, then we will try to use a zosmf token with an APIML token type. These TODO items are reminders that we must change these values once we move the APIML token logic from the POC plugin into the CLI.

packages/utilities/src/CliUtils.ts
gejohnston: I went back to this problem again. A different but very simple approach prevents a star from being displayed when the Enter key is typed.

__tests__/config/unit/jest.config.json
```diff
@@ -10,6 +10,7 @@
     "coverageReporters": [
         "json",
         "lcov",
+        "html",
```
dkelosky, line 13: @zFernand0 

This proposed change gives this on `npm run test:unit --coverage`:

image

dkelosky: ok im crazy - ill revert this

PR #401, opened 2020-05-18
Add functionality to disable population of default values in profiles

awharn:
Adds the --dd or --disable-defaults flag to the profile creation handler, allowing for the creation of profiles without their specified default values. This is vital for allowing the base profile to handle the values of these options without introducing breaking changes.

### DIFF
--- a/packages/imperative/src/profiles/handlers/CreateProfilesHandler.ts
+++ b/packages/imperative/src/profiles/handlers/CreateProfilesHandler.ts
@@ -41,6 +41,7 @@ export default class CreateProfilesHandler implements ICommandHandler {
             type: profileType,
             args: commandParameters.arguments,
             overwrite: commandParameters.arguments.overwrite,
+            disableDefaults: commandParameters.arguments.disableDefaults,
             profile: {}
         };
         /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
+- Add the --dd flag to profile creation to allow the profile to be created without the default values specified for that profile.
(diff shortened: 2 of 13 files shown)


packages/imperative/src/profiles/handlers/CreateProfilesHandler.ts
```diff
@@ -41,6 +41,7 @@ export default class CreateProfilesHandler implements ICommandHandler {
             type: profileType,
             args: commandParameters.arguments,
             overwrite: commandParameters.arguments.overwrite,
+            disableDefaults: commandParameters.arguments.disableDefaults,
```
gejohnston, line 44: Disclaimer: I can see that the precedent was already set, and I would not change your implementation at this time.

However, since commandParameters.arguments are already included in the createParms variable, does commandParameters.arguments.disableDefaults need to be passed in separately? Again, don't change it. I was just interested in your thoughts.

zFernand0:
Closed the PR intentionally for a second so that I can `Scan the repository` in the wash.zowe.org Jenkins server in order to clean up the merge conflict generated on rebase.
Should be fine after doing that, so re-opened it again.

packages/imperative/src/profiles/handlers/CreateProfilesHandler.ts
zFernand0: That's a good point.
May we should deprecate overwrite instead of adding a new thing to the interface and just process them from the `args` options :yum:

zFernand0: Created https://github.com/zowe/imperative/issues/403 as a reminder.
Thanks again @gejohnston

zFernand0:
getting `Sending interrupt signal to process` due to timeouts on wash.zowe.org right this moment. 
Will rerun later before merging! 😋

PR #407, opened 2020-05-28
Fix optional secure fields not deleted when overwriting a profile

### LINKED ISSUE #403: Deprecate any interface property duplication
Disclaimer: I can see that the precedent was already set, and I would not change your implementation at this time.

However, since commandParameters.arguments are already included in the createParms variable, does commandParameters.arguments.disableDefaults need to be passed in separately? Again, don't change it. I was just interested in your thoughts.

_Originally posted by @gejohnston in https://github.com/zowe/imperative/pull/401/files_

### LINKED ISSUE #406: Optional secure fields not deleted when overwriting a profile
Given that my profile schema has secure fields that are not required, 
When updating the profile (whether it's with update or save+overwrite)
Then the optional values stored in the vault/keychain do not get properly removed.

There may be some overlap with #379 in the way the secure fields are handled between BaseProfile and CLIProfile managers

tjohnsonBCM:
Fixes #406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Fix optional secure fields not deleted when overwriting a profile
+
 ## `4.6.2`
 
 - Fix vulnerabilities by updating yargs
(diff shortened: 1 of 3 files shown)


tjohnsonBCM:
> Tested that it removes the optional secure field from the vault. 👍
> Wondering if it's expected to store null values?
> image

This is expected behavior for now, the issue with storing null values was resolved in the "auto-auth-commands" branch.

zFernand0:
> Tested that it removes the optional secure field from the vault. 👍
> Wondering if it's expected to store null values?
> image

The behavior was already fixed in the `auto-auth-commands` branches. And it also cleans up the previously store null values 😋 
Thanks for clarifying.
image

PR #409, opened 2020-06-02
Auto auth commands

awharn:
Update command preparer to load base profiles as a fallback if something is not defined in service profile
Modify handling of positionals
Add auth command


packages/cmd/src/CommandPreparer.ts
```diff
@@ -28,9 +29,10 @@ export class CommandPreparer {
      * Prepare the command definition and apply any pass on traits to children.
      * After a definition has been prepared, it should be considered final.
      * @param {ICommandDefinition} original - The original command definition tree to "prepare"
+     * @param {ICommandProfileTypeConfiguration} baseProfile - An optional base profile to add to command definitions
      * @return {ICommandDefinition} A copy of the original that has been prepared
      */
-    public static prepare(original: ICommandDefinition): ICommandDefinition {
+    public static prepare(original: ICommandDefinition, baseProfile?: ICommandProfileTypeConfiguration): ICommandDefinition {
```
zFernand0, line 35: I believe this was one of the implementation decisions made in order to make it backwards compatible.
Looks good though! 👍

packages/imperative/src/auth/builders/AuthLoginCommandBuilder.ts
```diff
@@ -0,0 +1,76 @@
...
+            type: "command",
+            description: this.mConfig.login?.description,
+            handler: this.mConfig.handler,
+            options: [
+                {
+                    name: "show-token", aliases: ["st"],
+                    description: "Show the token when login is successful. If specified, does not save the token to a profile.",
+                    type: "boolean"
+                },
+                ...(this.mConfig.login?.options || [])
```
zFernand0, line 61: Nice use of typescript advanced features.
Reminder, people will have to update their version of typescript to continue development on their plugins.
But that's an expected developer operation, not a breaking change. Right?

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
```diff
@@ -0,0 +1,233 @@
...
+        // login to obtain a token.
+        const tokenValue = await this.doLogin(this.mSession);
+
+        // update the profile given
+        if (loadedProfile != null && loadedProfile.name != null && !params.arguments.showToken && tokenValue != null) {
+            await Imperative.api.profileManager(this.mProfileType).update({
+                name: loadedProfile.name,
+                args: {
+                    "token-type": this.mSession.ISession.tokenType,
+                    "token-value": tokenValue
```
zFernand0, line 120: Maybe I forgot, but why do we have to set the `hyphened-representation` and not the `camelCase` one?

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
```diff
@@ -0,0 +1,233 @@
...
+                name: "default",
+                type: this.mProfileType,
+                args: copyArgs,
+                overwrite: false,
+                profile: {}
+            };
+
+            const answer: string = await CliUtils.promptWithTimeout(
+                "We require the host, port, and token must be stored on disk for use with future commands. Is this acceptable? [y/N]: ");
+            if (answer == null || !answer.toLowerCase().startsWith("y")) {
```
zFernand0, line 153: is it safe to use `startsWith` or should we do `===`?

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
```diff
@@ -0,0 +1,233 @@
...
+            { requestToken: true, defaultTokenType: this.mDefaultTokenType }
+        );
+
+        this.mSession = new Session(sessCfgWithCreds);
+
+        // login to obtain a token.
+        const tokenValue = await this.doLogin(this.mSession);
+
+        // update the profile given
+        if (loadedProfile != null && loadedProfile.name != null && !params.arguments.showToken && tokenValue != null) {
```
zFernand0, line 115: Very interesting logic.
if we simplify it it will look sort of like:
```javascript
if (tokenValue != null && !params.arguments.showToken) {
  if (loadedProfile != null && loadedProfile.name != null) { // OR just   if (loadedProfile?.name != null) ???
    ...
  } else if (loadedProfile != null || loadedProfile.name != null) { // OR just   else ???
    ...
  }
}
```

- What if the `tokenValue` is null and the `doLogin` function never threw an error?
  - What if the `tokenValue` is null and `showToken` was specified?

Just trying to account for programmer mistakes (since I'm always messing things up myself) 😋

packages/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -0,0 +1,285 @@
...
+                while (answer === "") {
+                    answer = await CliUtils.promptWithTimeout(
+                        "Enter the host name of your service: "
+                    );
+                    if (answer === null) {
+                        throw new ImperativeError({msg: "We timed-out waiting for host name."});
+                    }
+                }
+                finalSessCfg.hostname = answer;
+            }
```
zFernand0, line 133: It'd be nice if we could reuse this entire block of code since it's kind of the same for every option, with the exception of string messages (which should ideally be extracted somewhere for localization purposes, but I'm having a hard time believing we will ever get to localizing the framework)

No changes requested, just whining! 😋

packages/imperative/src/Imperative.ts
```diff
@@ -678,6 +683,20 @@ export class Imperative {
             }
             rootCommand.children.push(CompleteProfilesGroupBuilder.getProfileGroup(allProfiles, this.log));
         }
+        const authConfigs: {[key: string]: ICommandProfileAuthConfig[]} = {};
+        if (loadedConfig.profiles != null) {
+            loadedConfig.profiles.forEach((profile) => {
+                if (profile.authConfig != null) {
+                    for (const requiredOption of ["user", "password", "tokenType", "tokenValue"]) {
```
tjohnsonBCM, line 690: Should we also require that an authConfig object have "host" and "port" options defined?

packages/imperative/src/auth/builders/AuthLoginCommandBuilder.ts
awharn: I think so, but would like to hear more feedback from other reviewers.

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
awharn: Because the option is specifically defined as `token-type`, and specifying the camel case one does not actually populate the option in the profile during the update because it does not go through Yargs.

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
awharn: My concern with === was that it would fail if a user entered "yes". I suppose we could just add logic to look for either "y" or "yes". I don't know why that didn't occur to me while I was writing this logic.

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
awharn: We could add logic to throw an error if tokenValue is null. I suppose we were counting on the REST handler in doLogin to throw some sort of error if something went wrong. In this case, I don't think I see a situation where it wouldn't throw an error, but if someone adds another plugin to provide token auth, then I suppose it would be good to handle this.

packages/messages/src/CoreMessages.ts
```diff
@@ -345,3 +344,27 @@ export const unexpectedProfileCreationError: IMessageDefinition = {
 export const unexpectedProfileUpdateError: IMessageDefinition = {
     message: `An unexpected profile update error occurred: \n{{unexpectedError}}`
 };
+
+export const loginAuthsCommandSummary: IMessageDefinition = {
+    message: `Login to an authentication server`,
+};
+
+export const loginAuthsCommandDesc: IMessageDefinition = {
+    message: `${loginAuthsCommandSummary.message}.`,
```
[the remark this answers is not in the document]
awharn: Probably.

packages/imperative/src/auth/builders/AuthCommandBuilder.ts
```diff
@@ -0,0 +1,69 @@
...
+     * @param {Logger} mLogger - logger instance to use for the builder class
+     * @param {IImperativeAuthConfig} mAuthConfig - the config for the auth type
+     */
+    constructor(protected mProfileType: string,
+                protected mLogger: Logger,
+                protected mAuthConfig: ICommandProfileAuthConfig) {
+
+        this.mConfig = mAuthConfig;
+        if (this.mConfig == null) {
+            throw new Error(`Auth Builder Error: No auth config was supplied.`);
```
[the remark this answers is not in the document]
awharn: Probably!

__tests__/__integration__/cmd/__tests__/integration/cli/auth/__scripts__/base_profile_and_auth_login_show_token_rfj_create.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+baseUser=$1
+basePass=$2
+
+# First create a base profile
+cmd-cli profiles create base-profile "test_base" --user "$baseUser" --password "$basePass"
```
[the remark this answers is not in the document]
awharn: Probably. I just used that to create the profile for that other script.

packages/imperative/src/auth/builders/AuthLogoutCommandBuilder.ts
```diff
@@ -0,0 +1,69 @@
...
+            description: this.mConfig.logout?.description,
+            handler: this.mConfig.handler,
+            options: this.mConfig.logout?.options,
+            examples: this.mConfig.logout?.examples,
+            profile: {
+                optional: [this.mProfileType]
+            },
+            customize: {}
+        };
+        // authCommand.customize[ProfilesConstants.PROFILES_COMMAND_TYPE_KEY] = this.mProfileType;
```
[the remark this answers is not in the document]
awharn: I assume we might be able to remove it? I am not quite sure what that does or where it came from.

__tests__/__integration__/cmd/__tests__/integration/cli/auth/Cmd.cli.auth.login.fruit.integration.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        // the output of the command should include token value
+        expect(response.stdout.toString()).toContain("fakeUser:fakePass@fakeToken");
+
+        response = runCliScript(__dirname + "/__scripts__/base_profile_and_auth_login_show_profiles.sh", TEST_ENVIRONMENT.workingDir);
+
+        // the output of the command should not include token value
+        expect(response.stderr.toString()).toBe("");
+        expect(response.status).toBe(0);
+        expect(response.stdout.toString()).not.toContain("tokenType:  jwtToken");
```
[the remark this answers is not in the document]
awharn: Changes made.

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
```diff
@@ -0,0 +1,233 @@
...
+                overwrite: false,
+                profile: {}
+            };
+
+            const answer: string = await CliUtils.promptWithTimeout(
+                "We require the host, port, and token must be stored on disk for use with future commands. Is this acceptable? [y/N]: ");
+            if (answer == null || !answer.toLowerCase().startsWith("y")) {
+                throw new ImperativeError({msg: `A login command was issued, but no ${this.mProfileType} profiles exist,` +
+                    ` the show token flag was not specified, or we were not given permission to create a profile.` +
+                    ` To resolve this problem, either specify the --showToken flag, create a base profile by using information found in` +
```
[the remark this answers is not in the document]
awharn: Fixed

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
```diff
@@ -0,0 +1,233 @@
...
+                    ` '${ImperativeConfig.instance.rootCommandName} profiles create ${this.mProfileType} --help',` +
+                    ` or re-run the login command and agree to the prompt requesting to store information on disk.`});
+            } else {
+                const createResponse: IProfileSaved = await Imperative.api.profileManager(this.mProfileType).save(createParms);
+                params.response.console.log(`\nProfile created successfully.\nTo create new profiles, which we use to store information, see ` +
+                `'${ImperativeConfig.instance.rootCommandName} profiles create ${this.mProfileType} --help'.\n` +
+                `To update this profile's information, see` +
```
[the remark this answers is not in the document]
awharn: Fixed.

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
awharn: Fixed.

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
awharn: Fixed.

packages/imperative/src/auth/builders/AuthCommandBuilder.ts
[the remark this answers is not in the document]
awharn: Fixed.

__tests__/__integration__/cmd/__tests__/integration/cli/auth/__scripts__/base_profile_and_auth_login_show_token_rfj_create.sh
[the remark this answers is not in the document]
awharn: Script name updated.

packages/imperative/src/auth/builders/AuthLoginCommandBuilder.ts
tjohnsonBCM: I think so as well. Since Typescript is a dev dependency, it shouldn't be a breaking change for end users.

packages/rest/src/session/ConnectionPropsForSessCfg.ts
gejohnston: If we do a bunch more prompting in a future effort, maybe we should replace (or wrapper) CliUtils.promptWithTimeout with a function that takes a name for the data value, forms the prompt string, does the while loop, and returns the answer or issues the timeout error.

packages/imperative/src/Imperative.ts
awharn: Maybe? I was under the impression it would based on the prompting that was added, but I could be wrong.

packages/imperative/src/auth/builders/AuthLogoutCommandBuilder.ts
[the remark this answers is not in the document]
zFernand0: It's commented out anyways. I'm fine leaving it or removing it. 
Just curious if someone knew its purpose before being commented. 😋

packages/imperative/src/Imperative.ts
awharn: Added,

gejohnston: I think that this is the right move, since I think this is about whether those properties are defined in the profile, not about whether they are required, or have been supplied.

PR #414, opened 2020-06-16
Don't load token value into Session object if user or password are supplied

### LINKED ISSUE #413: Can't use service profile after storing token in base profile
Steps to Reproduce:
* Run the "auth login" command in Zowe CLI, and allow it to create a base profile with a token value stored in it.
* Create a service profile in Zowe CLI with user and password fields set.
* Run a CLI command and supply the name of your service profile as a CLI option.

The command fails with:
```
Command Error:
Expect Error: Required parameter 'tokenValue' must be undefined
```

tjohnsonBCM:
Fixes #413

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Don't load token value into Session object if user or password are supplied
+
 ## `4.7.0`
 
 - Add the --dd flag to profile creation to allow the profile to be created without the default values specified for that profile.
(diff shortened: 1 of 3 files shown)


gejohnston:
Won't we also need a PR on the CLI to depend on the new version of imperative?

tjohnsonBCM:
> Won't we also need a PR on the CLI to depend on the new version of imperative?

Yes, after this PR is merged and the new Imperative version is published to Zowe Artifactory, we'll need a PR for the CLI as well.

zFernand0:
> > Won't we also need a PR on the CLI to depend on the new version of imperative?
> 
> Yes, after this PR is merged and the new Imperative version is published to Zowe Artifactory, we'll need a PR for the CLI as well.

if it's only to update the version number for imperative, then it should just work by re-running the pipeline after this one is merged in :yum:

We can open the PR for transparency though.

zFernand0:
Changelog entry will still be required. so, a PR is needed.
Thanks @tjohnsonBCM for clarifying :yum:

PR #419, opened 2020-06-22
Fix web help failing to load in IE11

### LINKED ISSUE #393: Web help doesn't load in IE11
Launching the web help in IE11 results in a mostly blank page:
image

Although Microsoft discourages use of IE11 in favor of modern alternatives, it is still commonly used by some enterprises, and supported better by accessibility software than newer browsers. If it's not too difficult, web help should be made usable in IE11.

tjohnsonBCM:
Fixes #393 
image


zFernand0:
Glad to see IE11 still working! 😋 
image

PR #421, opened 2020-06-24
Revert change to type of tokenType

tjohnsonBCM:
Signed-off-by: Timothy Johnson <timothy.johnson@broadcom.com>

### DIFF
--- a/packages/rest/src/session/doc/ISession.ts
+++ b/packages/rest/src/session/doc/ISession.ts
@@ -93,14 +93,14 @@ export interface ISession {
     base64EncodedAuth?: string;
 
     /**
-     * e.g. LTPA2
+     * Type of token in `tokenValue`, e.g. LTPA2
      * @type {string}
      * @memberof ISession
      */
-    tokenType?: SessConstants.TOKEN_TYPE_CHOICES;
+    tokenType?: string;
 
     /**
-     * aldgfuq;iwgueroquw3evrqj4e0f9iuq0324fq9034ifgq0394fjg08934ujg0q93j4fq3j4f09jq40[f9j]
+     * Token value of type `tokenType`
      * @type {string}
      * @memberof ISession
      */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 - Fix web help failing to load in Internet Explorer 11
 - Fix `--help-web` not working on macOS when DISPLAY environment variable is undefined
+- Revert accidental breaking change to `ISession.tokenType`, by changing its type to "string".
 
 ## `4.7.2`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 - Fix web help failing to load in Internet Explorer 11
 - Fix `--help-web` not working on macOS when DISPLAY environment variable is undefined
-- Revert accidental breaking change to `ISession.tokenType`, by changing its type to "string".
+- Change type of `ISession.tokenType` to "string" (for compatiblity with versions older than 4.7.0).
 
 ## `4.7.2`
 


packages/rest/src/session/doc/ISession.ts
```diff
@@ -93,14 +93,14 @@ export interface ISession {
...
-     * e.g. LTPA2
+     * Type of token in `tokenValue`, e.g. LTPA2
      * @type {string}
      * @memberof ISession
      */
-    tokenType?: SessConstants.TOKEN_TYPE_CHOICES;
+    tokenType?: string;
 
     /**
-     * aldgfuq;iwgueroquw3evrqj4e0f9iuq0324fq9034ifgq0394fjg08934ujg0q93j4fq3j4f09jq40[f9j]
```
tjohnsonBCM, line 103: 🤔

packages/rest/src/session/doc/ISession.ts
```diff
@@ -93,14 +93,14 @@ export interface ISession {
     base64EncodedAuth?: string;
 
     /**
-     * e.g. LTPA2
+     * Type of token in `tokenValue`, e.g. LTPA2
      * @type {string}
      * @memberof ISession
      */
-    tokenType?: SessConstants.TOKEN_TYPE_CHOICES;
+    tokenType?: string;
```
gejohnston, line 100: Is it possible to define the property as:

    tokenType?: SessConstants.TOKEN_TYPE_CHOICE | string;

I know that this would result in no better constraint than just string. However, it could provide documentation value. A programmer can find the set of values built into imperative, and use a symbolic constant for its value.

Another thought: is a value of "guid" generally valuable to others? Should we add "guid" to the list of SessConstants.TOKEN_TYPE_CHOICES?

If the answer to both of these questions is "no" or "impractical", then we should just commit the code the way it is.

tjohnsonBCM: If we implement the first suggestion, it won't make a difference in JSDoc:
image

awharn: I like the idea. It would probably be very helpful for developers. As for guid, I am not familiar enough to make a decision there.

gejohnston: I meant that a programmer trying to program an appropriate value would see SessConstants.TOKEN_TYPE_CHOICES, look up its definition in SessConstants.ts, and find:

TOKEN_TYPE_LTPA = "LtpaToken2"
TOKEN_TYPE_JWT = "jwtToken"
TOKEN_TYPE_APIML = "apimlAuthenticationToken"

I would then write code using TOKEN_TYPE_APIML.

awharn: That is also true. I don't see a problem with having both in there, and it could reduce problems developers have down the line if they decide to look into what the TOKEN_TYPE_CHOICES values are.

zFernand0: > adding "custom-tokenType" values

I don't see this a a sustainable practice (i.e. adding every single custom `tokenType` value to imperative) just to allow for a functionality that used to be there before.

> Having `TOKEN_TYPE_CHOICES | string`

Since it does not affect the jsdoc, I believe it won't affect the generated typedoc documentation.

zFernand0:
Interesting that the CliProfileManager wasn't change and codevov reports lower coverage.
https://codecov.io/gh/zowe/imperative/compare/c7bc9c1f35cd8932e30cb5e68a2c9c81ec05e246...e65d569cc435f41e21ef22ad5c2aa62686e9471a/changes
@dkelosky could you help me understand this one? :yum:

tjohnsonBCM:
> I'm wondering however, why we never made `positionals` an optional api.
> https://github.com/zowe/imperative/blob/c7bc9c1f35cd8932e30cb5e68a2c9c81ec05e246/packages/cmd/src/doc/handler/IHandlerParameters.ts#L53
> 
> Should we add something in the changelog about this (since we are forcing developers to implement it as part of the extension of the IHandlerParameters abstract class)?

Agree it would be good to either make `positionals` optional, or document in the changelog that it is now required.

PR #423, opened 2020-06-29
GitHub Actions

awharn:
Add workflow files to enable the use of GitHub actions in the Imperative repository, easing the burden on Jenkins.


.github/workflows/imperative.yml
```diff
@@ -0,0 +1,75 @@
...
+      if: ${{ always() && steps.build.outcome == 'success' }}
+      uses: papeloto/action-zip@v1
+      with:
+        files: __tests__/__results__/
+        dest: results.zip
+        recursive: false
+
+    - name: Archive Results
+      id: upload
+      if: ${{ always() && steps.build.outcome == 'success' }}
```
dkelosky, line 65: For my own education, what happens without the check for `always() && steps.build.outcome == 'success' `

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Check Changelog Updated
+      id: checkchangelogupdated
+      if: ${{ contains( github.event.pull_request.labels.*.name, 'no-changelog') != true }}
+      uses: awharn/check_changelog_action@v0.0.1
```
dkelosky, line 19: github action pro 😎

.github/workflows/imperative.yml
```diff
@@ -0,0 +1,75 @@
...
+      run: npm run build --if-present
+
+    - name: Unit Tests
+      id: unit
+      if: ${{ always() && steps.build.outcome == 'success' }}
+      run: npm run test:unit >> file.txt
+
+    - name: Unlock Keyring
+      if: ${{ always() && steps.build.outcome == 'success' && matrix.os == 'ubuntu-latest' }}
+      uses: tjohnsonBCM/unlock-keyring@master
```
dkelosky, line 47: you too are a github action pro 😎

.github/workflows/imperative.yml
awharn: It depends.  Without `always()`, it won't run when the tests fail, which is exactly when we want to look at the artifacts for the tests. However, then it will try to archive artifacts if the build failed, and they won't be present, so we added the extra logic to make sure that it built.

.github/workflows/audit.yml
```diff
@@ -0,0 +1,24 @@
+name: Audit
+
+on:
+  pull_request:
+    branches:
+      - master
+  schedule:
+    - cron: '0 10 * * *'
```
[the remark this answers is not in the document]
awharn: I believe it will also show up in the actions tab, and possibly as a red dot on the last commit. I will double check that.

.github/workflows/imperative.yml
dkelosky: thanks - this is good knowledge for the zowe explorer workflow!

.github/workflows/imperative.yml
```diff
@@ -0,0 +1,75 @@
...
+
+    - name: Integration Tests
+      id: integration
+      if: ${{ always() && steps.build.outcome == 'success' }}
+      run: npm run test:integration >> file.txt
+
+    - name: Zip Results
```
tjohnsonBCM, line 54: I think this stage of zipping the test results could be removed. The GitHub workflow for Zowe Explorer uploads the folder directly, and it is automatically zipped.
```yaml
- name: Upload test results
  uses: actions/upload-artifact@v1
  with:
    name: results
    path: results/
```

This might also resolve the issue we had with artifacts being doubly zipped 🙂

tjohnsonBCM:
> do you guys have any recommendations for the zowe explorer workflows based on the work you did here?

Not that much, I think we followed a lot of what you did 🙂 Maybe the check for `always() && steps.build.outcome == 'success'` that Andrew explained above, would also be good for ZE.

dkelosky:
Thank you both - I'll try to get that in a future PR soon onto Zowe Explorer.

.github/workflows/imperative.yml
awharn: I took this action, but just want you to be aware why it was being done. Archiving artifacts does indeed put them in a zip file, but they are being zipped without compression. Therefore, with this change, the archived artifacts swell in size from 3.5MB per build (~31.5MB total) to 25MB per build (~225MB total).

tjohnsonBCM: @awharn Sorry I think you mentioned this before and I forgot :cry: Interesting that GitHub creates a ZIP without compressing. Considering the compression benefits we were getting from the separate ZIP step, I'm OK with either method (using the ZIP step or not).

awharn: I have no real preference. AFAIK there isn't a storage limit on public repos. If it were private, I think we'd only get 9 runs out of the thing before we ran out of space. The only real argument I can make is download time, but 25 Megabytes shouldn't be too bad.

tjohnsonBCM: Not sure if the 25 MB indicates how much space is actually being used on GitHub's servers, or if GitHub remembers how big the folder was before compressing it. I see this in web browser:
image
But when I download the ZIP, I see this in Windows Explorer:
image

awharn: Oh. Hm. Well, if that is the case, then we can just get rid of the zip step. But I think it is odd it displays the uncompressed size on the Web UI right next to the download button.

PR #424, opened 2020-07-16
Add timeout to ISession

awharn:
Add timeout to ISession for convenience.

### DIFF
--- a/packages/rest/src/session/doc/ISession.ts
+++ b/packages/rest/src/session/doc/ISession.ts
@@ -115,7 +115,7 @@ export interface ISession {
     rejectUnauthorized?: boolean;
 
     /**
-     * The amount of time to specify in the X-IBM-Response-Timeout header
+     * A convenience variable to make it easier to implement timeouts on extended or custom REST clients
      * @type {number}
      * @memberof ISession
      */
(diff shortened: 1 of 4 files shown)


packages/rest/src/session/doc/ISession.ts
```diff
@@ -114,6 +114,13 @@ export interface ISession {
      */
     rejectUnauthorized?: boolean;
 
+    /**
+     * The amount of time to specify in the X-IBM-Response-Timeout header
```
dkelosky, line 118: is this header and implementation more specific to zowe cli & zosmf than imperative?

awharn: The idea to add timeout to the ISession is to make it available for REST clients to be able to extend the base REST client and define a timeout as they see fit. The information provided on that comment in particular was specific to zOSMF due to the testing I was doing, and I forgot to change it to something more generic.

dkelosky: Ok gotcha!

zFernand0:
@awharn, is this supposed to be closed now?

awharn:
That is correct. I will go ahead and close it. This wasn't needed.

Issue #428, opened 2020-07-22
allowableValues doesn't work with array type


liuxh0:
When I define an array-typed option with allowableValues, the validation doesn't seem to work properly. 

Semantically I think this should be interpreted that this option accepts multiple values and each of the values should be allowable.

Currently, the validity is checked by this method:
https://github.com/zowe/imperative/blob/ce314e1c3079696664fffa4382ac8b55a2853183/packages/cmd/src/syntax/SyntaxValidator.ts#L543-L553

If the option is of type array, `input` is also of type array. Although the parameter of `RegExp.test()` [should be of type string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test), it will somehow first convert other types to string and then perform regular expression match.

I want to extend the support for array. I offer to create a PR if this concept is okay.

PR #429, opened 2020-07-23
Array with allowable values

liuxh0:
#428

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Add allowableValues support for array
+
 ## `4.7.4`
 
 - Fix update profile API storing secure fields incorrectly when called without CLI args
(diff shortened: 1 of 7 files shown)


tjohnsonBCM:
LGTM. Please merge the latest changes from `master` into your branch.

scifipony:
Please ensure in testing that an array of no elements (an option with no argument) is not flagged in error if **allowableValues** _is not coded_, or this will be a breaking change.

liuxh0:
Hi @scifipony , I am not really sure what you mean.

The scenario you described will be flagged in error because an array is expected but a boolean is provided. The behavior is exactly the same before my change. Why this is a breaking change?

However in Zowe, an array-type option with no argument will result in an empty array instead of a boolean true. I am not familiar but I assume Zowe did some error handling.

scifipony:
@liuxh0 

> in Zowe, an array-type option with no argument will result in an empty array instead of a boolean true.

Correct.

> I assume Zowe did some error handling

Best I can tell, it does not.  It certainly does not flag as in-error an array-type option with no argument.  I want to ensure that this behavior continues.  Should Zowe start flagging it, it would be a breaking change for my product.

liuxh0:
@scifipony 
I am still a little bit confused. Before and after, the behavior is the same. How can I help here?

scifipony:
> @scifipony
> I am still a little bit confused. Before and after, the behavior is the same. How can I help here?

All I was asking to _ensure the behavior didn't change_.  If the behavior is the same "before and after" the changes you made, then everything is fine.

PR #430, opened 2020-07-27
Run on pull request as well as push

awharn:


awharn:
The failing test on macOS is just the timeout test. 999ms is close enough to 1 second, right? 😄

PR #431, opened 2020-07-30
AbstractSession.storeCookie() support for run-time defined cookie names.

Tulsagrammer:
Commit b1a0fad introduced a breaking change to AbstractSession.storeCookie() for APIs and applications generating cookie names that can't be fully known at build time. The current code assumes the "=" character separating cookie name and cookie value immediately follows the value stored in tokenType. This assumption breaks down when the server generates a cookie name starting with a fixed character string followed by an additional character string generated at run time.

This proposed change continues to support static cookie names as well as dynamically generated cookie names. Imperative's test:unit was run successfully against this change.

Dynamically generated cookie names might be used by APIs and servers that can listen for connections on multiple ports, where the port has some significance to the server. If that server used a fixed character string for the cookie name, it might have difficulty distinguishing the source of a returned cookie. In our case, we have a server that generates cookie names using the string "EJESWEB_" followed by the port number on which the request was accepted.

### DIFF
--- a/packages/rest/src/session/AbstractSession.ts
+++ b/packages/rest/src/session/AbstractSession.ts
@@ -220,9 +220,9 @@ export abstract class AbstractSession {
                 // if element begins with tokenType, extract full tokenType and tokenValue.
                 if (element.indexOf(this.mISession.tokenType) === 0) {
                     // parse off token value, splitting element at first "=".
-                    const temp: string[] = element.split("=");
-                    this.ISession.tokenType  = temp[0];
-                    this.ISession.tokenValue = temp[1];
+                    const split = element.indexOf("=");
+                    this.ISession.tokenType  = element.substring(0, split);
+                    this.ISession.tokenValue = element.substring(split + 1);
                 }
             });
         });
--- a/packages/rest/src/session/AbstractSession.ts
+++ b/packages/rest/src/session/AbstractSession.ts
@@ -217,10 +217,12 @@ export abstract class AbstractSession {
             const authArr = auth.split(";");
             // see each field in the cookie, e/g. Path=/; Secure; HttpOnly; LtpaToken2=...
             authArr.forEach((element: string) => {
-                // if we match requested token type, save it off for its length
+                // if element begins with tokenType, extract full tokenType and tokenValue.
                 if (element.indexOf(this.mISession.tokenType) === 0) {
-                    // parse off token value, minus LtpaToken2= (as an example)
-                    this.ISession.tokenValue = element.substr(this.ISession.tokenType.length + 1, element.length);
+                    // parse off token value, splitting element at first "=".
+                    const ejesCookie: string[] = element.split("=");
+                    this.ISession.tokenType  = ejesCookie[0];
+                    this.ISession.tokenValue = ejesCookie[1];
                 }
             });
         });
--- a/packages/rest/src/session/AbstractSession.ts
+++ b/packages/rest/src/session/AbstractSession.ts
@@ -218,11 +218,16 @@ export abstract class AbstractSession {
             // see each field in the cookie, e/g. Path=/; Secure; HttpOnly; LtpaToken2=...
             authArr.forEach((element: string) => {
                 // if element begins with tokenType, extract full tokenType and tokenValue.
-                if (element.indexOf(this.mISession.tokenType) === 0) {
+                if (element.indexOf(this.ISession.tokenType) === 0) {
                     // parse off token value, splitting element at first "=".
                     const split = element.indexOf("=");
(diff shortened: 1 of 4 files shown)


packages/rest/src/session/AbstractSession.ts
```diff
@@ -217,10 +217,17 @@ export abstract class AbstractSession {
...
+                // if element begins with tokenType, extract full tokenType and tokenValue.
+                if (element.indexOf(this.ISession.tokenType) === 0) {
+                    // parse off token value, splitting element at first "=".
+                    const split = element.indexOf("=");
+                    if (split >= 0) {
+                        this.ISession.tokenType  = element.substring(0, split);
+                        this.ISession.tokenValue = element.substring(split + 1);
+                    }
+                    else {
+                        this.ISession.tokenValue = "";
```
tjohnsonBCM, line 229: This may introduce a breaking change if someone has a condition checking whether `tokenValue` is null or undefined. Could the `else` clause here be removed to leave `tokenValue` undefined if not present?

Tulsagrammer: Certainly. I'll include that change along with the unit test you suggested.

Tulsagrammer: Requested changes have been implemented.

zFernand0: Thanks for adding the test and addressing the comments! 👍 

To really illustrate the problem, could you possibly use a random tokenType and non-realistic tokenValue (something that would result in `tokenType = "cool" ; tokenValue = "beans"`) and isntead of using snapshots, explicitly expect the token value to be what you set it to (`"beans"` for example).

The reason behind the request to avoid snapshots for tests catching breaking changes is that someone can easily overlook a snapshot value with the `--updateSnapshot` flag for `jest`. 
For example, I copied the same exact test (with `tokenType="LtpaToken"` instead of `LtpaToken2`) and the only thing that changed was the initial equal sign (`=7KM...` ). However it reported that the tests passed since there was no snapshot already for it in master.

Tulsagrammer: Unit test has been updated to reflect zFernand0's suggestion.

zFernand0: Thanks a lot @Tulsagrammer.
Were you able to delete the snapshot? I believe I can still see 4 files changed (and one of them being the snapshot addition)

Tulsagrammer: @zFernand0 just deleted the snapshot the earlier unit test created. Now showing three files.

zFernand0:
Just need to address the merge conflict, and this PR should be good to go! 😋

Tulsagrammer:
@zFernand0: I'm sorry to bug you, but I'm lost. I'm still getting "Conflicting files" on CHANGELOG.md. I just don't see a conflict: I'm just adding a change comment for my storeCookie above the existing 4.7.5 change entry. (The most recent change entry on master.) Any suggestions or advice on what I'm doing wrong?

ChrisBoehmCA:
If you're using an editor like VSCode, you should be able to try to pull master into your branch and resolve the conflict graphically. https://code.visualstudio.com/docs/editor/versioncontrol (Section `Merge conflicts`). The git conflict is probably something simple but I haven't pulled your branch to see. Also I think you are supposed to use a "Recent Changes" header in the changelog in pull requests and the version is assigned later after the PR is merged. Fernando can confirm

Tulsagrammer:
@zFernand0: the changelog conflict has been addressed. I notice that two checks appear to have failed. Looking at the details of "Imperative CI / test (10.x, windows-latest)" it appears that the issues causing the test to fail are not related to the files in my pull request. In particular, "packages/rest/__tests__/session/Session.test.ts" has a PASS. As for "continuous-integration/jenkins/pr-merge", the other failing test, I have no idea what's making that test fail.

Tulsagrammer:
@zFernand0: as a follow-up to my previous message, it appears that "continuous-integration/jenkins/pr-merge" is failing on issues associated with integration testing. More specifically, "__tests__/src/packages/profiles/CliProfileManager.credentials.spec.ts" appears to be having problems with secure credential storage: "Couldn't create item: The secret was transferred or encrypted in an invalid way." A second issue seems to be associated with tests of CredentialManagerFactory.ts.

tjohnsonBCM:
@Tulsagrammer Re-ran the GitHub and Jenkins builds, both failures were transient and the builds passed the second time. The "codecov/project" check is still failing but can be ignored, as it often shows inaccurate reports and "codecov/patch" shows that the PR has sufficient coverage.

Issue #422, opened 2020-06-26
Research: Explore the possibility to expose a callback function when doPrompting is false


zFernand0:
The `addPropsOrPrompt` works great for CLI as the client of this functionality.
There is an opportunity to improve the SDK further so that people can implement their own mechanism for asking end users for information *at the right time*. For this purpose, we could evaluate what would it take to provide a callback mechanism instead of skipping the gathering of required properties.

https://github.com/zowe/imperative/blob/c9de181194933c2ba82e8d85513a26aa3a87b445/packages/rest/src/session/ConnectionPropsForSessCfg.ts#L123

gejohnston:
Here are some implementation thoughts on this subject:

- The function that we would modify is: 

    imperative\packages\rest\src\session\ConnectionPropsForSessCfg.ts - 
    ConnectionPropsForSessCfg.addPropsOrPrompt()

- In the CLI, addPropsOrPrompt is called from a command handler like this:
```
const sessCfg: ISession =  {
    rejectUnauthorized: commandParameters.arguments.rejectUnauthorized,
    basePath: commandParameters.arguments.basePath
};
const connectableSessCfg = await ConnectionPropsForSessCfg.addPropsOrPrompt<ISession>(
     sessCfg, commandParameters.arguments
);
mySession = new Session(connectableSessCfg);
```

- This is the signature of the addPropsOrPrompt function:
```
public static async addPropsOrPrompt<T>(
    initialSessCfg: T,
    cmdArgs: ICommandArguments,
    options: IOptionsForAddConnProps = {}
): Promise<T>
```

- This is the definition for IOptionsForAddConnProps:
```
export interface IOptionsForAddConnProps {
    requestToken?: boolean;
    doPrompting?: boolean;
    defaultTokenType?: SessConstants.TOKEN_TYPE_CHOICES;
}
```

- When not supplied, the default value for doPrompting is true, so the CLI currently prompts without having to set this option.
    - With the new option proposed below, the doPrompting option may become unnecessary, but I think that we would keep the doPrompting option, so that we do not introduce a breaking change.

- We could add a new option property named getValuesCallback. It would be a function that does its own prompting.
    - The getValuesCallback function must accept one parameter named 'neededProperties'. It would be a map like the following:

    ```
        {
            user: null,     <--  example
            neededProp2: null,
            neededProp3: null
        }
    ```

[...]

PR #433, opened 2020-08-03
Sso callback

jellypuno:
This is to address issue #422 Explore the possibility to expose a callback function when doPrompting is false.

In this PR, I have added 
- a new option called getValuesBack. This is callback function that will be used by addPropsOrPrompt().
- new unit tests to test this option


jellypuno:
I don't know how to fix the snapshots. I was able to edit it temporarily in Jenkins but I do not want to commit those changes in github actions file. How do I fix this?

jellypuno:
> IMO the following behavior seems unexpected:
> 
> ```
>      * There are 2 ways to call this:
>      * - doPrompting = false && getValuesBack !== null. This option will ask the values for the
>      *   needed properties and it will provide an array that lists all the properties that it needs.
>      * - doPrompting = true && getValuesBack !== null. This option will ask the values for the
>      *   needed properties one by one. For more info, please check .\ConnectionPropsForSessCfg.test.ts
> ```
> 
> Should we be making use of the `doPrompting` property in this way? In both cases we are prompting for values. The only difference seems to be whether we prompt for them individually or all at once.
> 
> An alternative suggestion would be to have two optional and mutually exclusive properties `getValueBack` and `getAllValuesBack` which would work like this:
> 
> * `doPrompting = true && getValueBack != null` - This option will ask the values for the needed properties one by one, since the callback requests for only one value.
> * `doPrompting = true && getAllValuesBack != null` - This option will ask the values for the needed properties and it will provide an array that lists all the properties that it needs, since the callback requests for all values at once.
> * `doPrompting = false` - This option will keep the same behavior as before. No prompting will be done of any sort, either on the CLI or through callback functions.
> 
> This would allow the function signatures to have more specific types and perhaps be more self-explanatory:
> 
> ```
> getValueBack? (property: string): any
> getAllValuesBack? (properties: string[]): any[]
> ```

@tjohnsonBCM Makes sense. I can do that.

packages/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -119,33 +122,79 @@ export class ConnectionPropsForSessCfg {
             finalSessCfg.tokenValue = cmdArgs.tokenValue;
         }
 
+        // This function will provide all the needed properties in one array
+        if (optsToUse.getAllValuesBack) {
```
tjohnsonBCM, line 126: What happens if both properties are set on the options object? Currently it looks like if getValueBack and getAllValuesBack are both set, then getAllValuesBack will take priority. Is this the expected behavior, or should we throw an error if both are set saying that the properties are mutually exclusive?

packages/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -119,33 +122,79 @@ export class ConnectionPropsForSessCfg {
...
-                        "Enter the port number for your service: "
-                    );
-                    if (answer === null) {
-                        throw new ImperativeError({msg: "Timed out waiting for port number."});
-                    } else {
-                        answer = Number(answer);
-                        if (isNaN(answer)) {
-                            throw new ImperativeError({msg: "Specified port was not a number."});
+                if (optsToUse.getValuesBack) {
+                    answer = optsToUse.getValuesBack("port");
```
tjohnsonBCM, line 185: Should getValueBack/getAllValuesBack be an async function that returns a promise and can be awaited like `CliUtils.promptWithTimeout`?

packages/rest/src/session/doc/IOptionsForAddConnProps.ts
```diff
@@ -43,4 +43,15 @@ export interface IOptionsForAddConnProps {
      * The default value is TOKEN_TYPE_JWT.
      */
     defaultTokenType?: SessConstants.TOKEN_TYPE_CHOICES;
+
+    /**
+     * Specifies the functionality that external applications will use for prompting.
+     * Activate this function if needed propoerties are to be prompted one by one
```
tjohnsonBCM, line 49: "propoerties" -> "properties"

packages/rest/src/session/doc/IOptionsForAddConnProps.ts
```diff
@@ -43,4 +43,15 @@ export interface IOptionsForAddConnProps {
...
+     */
+    getValuesBack? (properties: string): any;
+    /**
+     * Specifies the functionality that external applications will use for prompting.
+     * Activate this function if all the needed propoerties are prompted all at once
```
tjohnsonBCM, line 54: "propoerties" -> "properties"

packages/rest/src/session/doc/IOptionsForAddConnProps.ts
```diff
@@ -43,4 +43,15 @@ export interface IOptionsForAddConnProps {
      * The default value is TOKEN_TYPE_JWT.
      */
     defaultTokenType?: SessConstants.TOKEN_TYPE_CHOICES;
+
+    /**
+     * Specifies the functionality that external applications will use for prompting.
+     * Activate this function if needed propoerties are to be prompted one by one
+     */
+    getValuesBack? (properties: string): any;
```
tjohnsonBCM, line 51: Would it make sense for this to be singular (value instead of values, property instead of properties) since it fetches only a single value?
`getValueBack? (property: string): any;`

packages/rest/src/session/ConnectionPropsForSessCfg.ts
jellypuno: I don't know. I mean, they can throw in 2 different functions in those options. What do you think @zFernand0 @crawr @katelynienaber ?

packages/rest/src/session/ConnectionPropsForSessCfg.ts
jellypuno: yeah. why not. i'll add that

packages/rest/src/session/doc/IOptionsForAddConnProps.ts
jellypuno: changed.

jellypuno:
@tjohnsonBCM I added the changes based on your comments.

packages/rest/src/session/ConnectionPropsForSessCfg.ts
katelynienaber: I am not really sure why we need all of the following: `getValuesBack, getAllValuesBack, doPrompting`.

Couldn't we just ask that people pass in a prompting function if they want to prompt? And in the CLI, wherever `addPropsOrPrompt` was called with `doPrompting = true` before, we just pass in the prompting block that used to be inside of `addPropsOrPrompt` itself (the block starting at line [164](https://github.com/zowe/imperative/blob/5eb5c5b0ec8069bf7a1e119df6ee6222eff01807/packages/rest/src/session/ConnectionPropsForSessCfg.ts#L164) would be removed). Then we can tell whether they want to prompt by whether they passed in a function, and it would get rid of confusion surrounding `doPrompting`. If they passed in the function, we prompt. Otherwise, we return an incomplete session (or an array of all the fields that were missing).

If we really have to keep `doPrompting`, I don't see any difference between returning user input one field at a time, or all at once. I think giving people the option to decide how they want it is confusing. They can just receive all the prompted values in an array (maybe an array with just one value) and if they want to handle them one at a time they can do that in their code.

tjohnsonBCM: @katelynienaber We do have to keep `doPrompting` for now, unless we're OK with a breaking change.

> They can just receive all the prompted values in an array (maybe an array with just one value) and if they want to handle them one at a time they can do that in their code.

That is a good point. I believe the discussion in #422 resulted in a decision to give people an option between returning user input one field at a time, or all at once. But perhaps this option causes unnecessary confusion? @jellypuno @zFernand0

jellypuno: In my opinion, we could implement it like this for now then if users think that it is confusing then we can re-factor it. If I look at it 10 years for now, there might be features that can be added to getAllValuesBack that is not available in getValueBack.

jellypuno: So.. I talked to @dkelosky  and he gave me his opinion about this matter. And since this 2 option is giving others confusion, I just decided to change it and follow Dan and Katelyn's suggestions. 

Current approach: There's one option called getValuesBack. Whether the external app wants to prompt the values one by one or all at once then it is the apps decision. doPrompting is set to false when getValuesBack is  not null. If doPrompting === true then it will trigger the original approach.

@tjohnsonBCM @katelynienaber @zFernand0 Please let me know what you think. I really want this PR to move forward because the are 2 PRs in Zowe Explorer that is dependent on this.

PR #448, opened 2020-10-04
Fix: Use Proper Daemon Process Exit Code

dkelosky:
Note: comments here that reference Zowe CLI are only to illustrate specific `imperative` issues. 

This fix will allow for a native client to exit with the same exit code the Zowe CLI ends with.   For example, if a `zowe` command fails and ends with an exit code of `1` the daemon client with end with the same code. 

Also, fixed by this PR:
- `zowe files list ds ...` will now return the full list of data sets returned  
- `zowe jobs submit jobs --directory ...` no longer hangs the client
- `zowe proflist ...` (invalid command group) no longer hangs the client 
- `zowe uss issue ssh` no longer fails
- `--help-web` properly launches

`--cwd` conflicted with existing commands.  This was renamed to `--lcd`.


MikeBauerCA:
I tested all of the claimed fixes above besides `zowe uss issue ssh`. The fixes I tested seemed complete with the exception of `--help-web` working properly. It works fine at the top level (`zowe --hw`). However, at lower levels (e.g. `zowe files --hw`) the web help is launched quickly but the command hangs on the terminal.

Thanks for working on this feature. I really like the responsiveness 👍

packages/utilities/src/DaemonUtils.ts
```diff
@@ -0,0 +1,151 @@
...
+     * @memberof DaemonUtils
+     */
+    public static readonly X_ZOWE_DAEMON_END = "x-zowe-daemon-end:";
+
+    /**
+     * Version 1 of the headers means there are 7 total.
+     * @static
+     * @memberof DaemonUtils
+     */
+    public static readonly X_ZOWE_HEADERS_V1 = 1;
```
gejohnston, line 86: This is a minor item about variable names:
    X_ZOWE_HEADERS_V1 = 1
    X_ZOWE_HEADERS_V1_LEN = 7

The name and the value both appear to lock the variable name into version 1. If we ever produced version 2, which change below would feel inclined to make?
    X_ZOWE_HEADERS_V1 = 2
or also change the variable name?
    X_ZOWE_HEADERS_V2 = 2

gejohnston (approved):
Things look good.

Not related to your changes, but what do you think we should be doing about the numerous code coverage issues raised in pull requests that are not related to the code that was actually changed in the pull request?

packages/utilities/src/DaemonUtils.ts
dkelosky: I'll rename the two you pointed out.   `X_ZOWE_HEADERS_V1` and `X_ZOWE_HEADERS_V1_LEN` are initial values, not headers.  

X_ZOWE_DAEMON_VERSION is supposed to hold the first version.

dkelosky:
What I typically do for code coverage issues is whine / complain to @zFernand0 about it...

I'm not sure what to do, but it always seems to cause me a lot of troubles.  Besides the item you pointed out, I sometimes get SonarCloud issues on coverage while CodeCov has no problems and vice versa.

PR #450, opened 2020-10-14
JSON config

tucker01:
I'm opening up a draft PR to start the discussion on my JSON config implementation. **This PR does not meet contribution requirements and should not be merged.**

https://github.com/zowe/zowe-cli/issues/749
https://github.com/zowe/zowe-cli/issues/833

# Goals
- Consolidate profiles, settings, plugins into a config file
- Support project and global configurations
- Support static "site-wide" config
- Support type-less/freeform profiles
- Support JSON schema for editing
- Support external config consumers
- Support secure fields
- Support a more "npm/node"-like model of config
- Eliminate/reduce imperative settings, plugins, and profiles code
- Eliminate redundant profile info
- Eliminate the strict requirement of a profile type

At the moment the configuration files are JSON. It is trivial (hopefully) to support YAML.

Nomenclature is subject to change. Concepts are the important bit. 

These changes should NOT break existing configurations. Imperative will apply the "new way of doing things" based on the existence of a config.

**NOTE:** Not everything may function exactly as described... let me know if you have issues. Tests have yet to be written.

# Overview
Imperative settings (settings.json), profile (yaml profiles), and plugin (plugins.json) configuration has been consolidated (or eliminated wherever possible) into  `<app>.config.json` (e.g. `zowe.config.json`).

`<app>.config.json` can exist in two places:
* **project:** found via a search from the `cwd` (e.g. `~/projects/mfauto/zowe.config.json`)
* **global:** in the CLI home (e.g. `~/.zowe/zowe.config.json`) 

Alongside either `<app>.config.json`, you can create an `<app>.config.user.json`. 

All four files are identical in structure. However, their **intention** differs:

* **global \<app\>.config.json:** Global configuration which applies to all invocations of a CLI. For example, secure credentials may be a requirement at your site... you would specify a secure credential plugin in the global config. Also, this configuration could be hosted at your site and pulled, massively reducing CLI setup. 

* **global \<app\>.config.user.json:** User specific global configuration. For example, your site admin may create a global `<app>.config.json` that contains hosts and services. You would specify your credentials for those services in the global user config.

* **project \<app\>.config.json:** Project configuration which can be committed to version control. For example, your project may use a particular host/port/service... you would specify this in the project config and commit to version control. 

* **project  \<app\>.config.user.json:** User specific project configuration. For example, your project `<app>.config.json` may specify a test system. You would specify your credentials for that test system in the project user config.  

Internally, each config file location is known as a "layer". 

# Basic Structure
```typescript
export interface IConfig {
    defaults: { [key: string]: string };
    profiles: { [key: string]: IConfigProfile };
    plugins: string[];
    secure: string[];
    properties: { [key: string]: any };
}
```

property | desc
---|---
profiles | named profiles 
defaults | defaults for profile types
properties | properties that apply to all profiles
plugins | the list of plugin to load
secure | properties that are secure

# Example
**example global \<app\>.config.json:**
```json
{
    "$schema": "schema.json",
    "profiles": {
        "host1": {
            "properties": {
                "host": "my.host.net",
                "port": 12345,
                "rejectUnauthorized": false
            },
        }
    },
    "defaults": {
        "zosmf": "host1",
    },
    "plugins": [
       "@zowe/secure-credential-store-for-zowe-cli"
    ],
    "secure": [],
    "properties": {}
}
```

**example global \<app\>.config.user.json:**
```
{
    "profiles": {
        "host1": {
            "properties": {
                "user": "managed by Zowe CLI",
                "password": "managed by Zowe CLI"        
            }
        }
    },
    "defaults": {},
    "properties": {},
    "plugins": [],
    "secure": [
        "profiles.host1.properties.user",
        "profiles.host1.properties.password"
    ]
}
```

In the above example, the global config and global user configuration files are merged to create the complete global configuration. Note that profile `host1` is specified in both files... In the config file, the host/port is specified... In the user config file, the credentials. 

Note that there is no type specified in the above example. Profiles are now "blobs" of properties organized by the user. However, you may specify `type` alongside `properties` to provide JSON schema autocompletion which can assist in creating profiles that are suitable for specification on the CLI (e.g. `--zosmf-p`). 

# Merge Strategy 
The config files are merged in the following order (with precedence most to least in descending order): 

* project user config
* project config
* global user config
* global config 

Certain merge strategies are applied to each config property. 

For example, profiles are deep merged at the project layer and the global layer, but are not merged across the global and project layer. Instead, if a profile exists in the project layer and the global layer, the global layer profile is discarded in favor of the project profile. 

# Nesting Profiles
To reduce the duplication of profile info, you can infinitely nest profiles to create a structure that logically groups and reuses properties. 

**example:**
```json
{
    "$schema": "./schema.json",
    "profiles": {
        "lpar1": {
            "properties": {
                "host": "lpar1.hostname.net",
                "rejectUnauthorized": false
            },
            "profiles": {
                "endevor": {
                    "properties": {
                        "port": 12345,
                        "protocol": "http"
                    },
                    "profiles": {
                        "prd": {
                            "properties": {
                                "instance": "INST1",
                                "environment": "PRD",
                                "system": "APPSYS",
                                "subsystem": "APPSUBSYS"
                            },
                            "profiles": {
                                "p1": {
                                    "properties": {
                                        "stage-number": "1"
                                    }
                                },
                                "p2": {
                                    "properties": {
                                        "stage-number": "2"
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    },
    "defaults": {
        "zosmf": "lpar1",
        "endevor": "lpar1.endevor.prd"
    },
    "plugins": [
        "@broadcom/endevor-for-zowe-cli"
    ],
    "secure": [],
    "properties": {}
}
```

In the above example, we've created a root level profile called `lpar1`. `lpar1.properties` contain the properties (such as the hostname) that apply to all services available on `lpar1`. Nested under `lpar1` is an `endevor` service, which uses port `12345`... and nested under that are additional profiles indicating Endevor specific locations. 

On the CLI, we can reference, for example, the `p2` location by specifying `--endevor-location lpar1.endevor.prd.p2`. All profiles along this "path" will be merged to create the final profile.

Similarly, on the CLI, we can specify `--zosmf-p lpar1` to access z/OSMF. In this case, our z/OSMF instance uses the default port of `443`. If it did not, we could specify a `zosmf` profile under `lpar1`, with just the port:
```json
"zosmf": {
    "properties": {
        "port": 123
    }
}
```
And specify `--zosmf-p lpar1.zosmf`

In the past, we'd have to duplicate information in the zosmf and endevor profiles, such as host or subsystem. Now, we only need to specify this information once per set of nested profiles. 

This structure should work well with the APIML (single host with multiple services).

# Using Profiles on the CLI
With the removal of the strict type requirement, you may specify any config profile or profile path on any profile option.

A profile path causes all properties along the path to be merged. See the Endevor example above. 

# Applying Properties to All Profiles
In some cases, you may have multiple profiles at the root of the config (multiple hosts/lpars) and you would like to share, for example, credentials among them. You can use the root level `properties` property to achieve this. Properties specified on the `properties` property (lol) will be applied to each root level profile. 

# Config Commands
Unfortunately, `config` is already a command group, so I temporarily added a group called `cnfg` (pending a better name). The `cnfg` group provides commands to list your config (and the locations where properties may be specified), create the config, modify the config, and so on. 

See `zowe cnfg -h` for details. 

**NOTE:** Outside of secure properties, the config is intended to be edited by hand. You can use the `zowe cnfg set` command with the `--secure` option to set secure fields. 

# Schema 
The JSON schema is dynamically generated based on your available plugins:
```
zowe cnfg schema
```
And can be included via the `$schema` property. 

# Plugins
Plugins are loaded from one of two places:
* `node_modules` in your project
* `node_modules` in your `~/.zowe` 

`npm` should be used to install/manage plugins. For example, if you have specific plugins for a project, you could specify them in `package.json` and add them to your project `<app>.config.json`. The user would then clone, issue `npm install` and the plugins are now available/usable... no `zowe` specific steps required. 

To install plugins into the global location use the `npm --prefix` option. 

# Deprecations
The following should be deprecated:
* Legacy profile management and commands
* App Settings (settings.json) and config commands (which manipulate settings)
* plugins.json and plugins commands

# API
The config is managed with `Config.ts`. It can/should be used outside of the CLI if access to the config is required (for example, by the vscode extensions).


dkelosky:
I tested this out - it's awesome!  

There are a few usability ideas but these could all be ignored since it's a great feature as-is:
* add some examples for the profile types
image
* add "$schema": "https://json-schema.org/draft/2019-09/schema#" to your emitted `schema.json`
* add a command to secure fields that are manually entered into `zowe.config.user.json`

tucker01:
Thanks @dkelosky! 

Will definitely add the schema URL.

Examples would be great... I'm collecting the schemas from each plugin. Would it make sense for folks to update their schemas with examples (we can add the "empty" by default)?

Currently to set a secure field you would issue `zowe cnfg set path.to.property value --secure`. I would really like to support manually creating a config and then having a command/action to secure all fields. What would make the most sense? Maybe `zowe cnfg secure` would look at the `secure` property, secure them all, and update with `managed by blah`?

ChrisBoehmCA:
It looks cool, I have it installed and will try it out more soon.  

I think this new config concept could fix this issue https://github.com/zowe/zowe-cli/issues/225 . If you could specify an arbitrary length array of paths, you could solve this use case of one shared location set up by an administrator with all the plugins folks might need.  Node has a setting built in that is similar but probably can't be used directly to implement the feature (NODE_PATH environmental variable https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders)

I also think this will make it easier to manage credentials for systems with the same security environment, since you can re-use them for multiple profiles.

tucker01:
Thanks for checking it out! It would be great if we could "assume" that "global for all users" location, but a config property for alternative paths would work.

dkelosky:
Recommending plugins update their schemas to also have `examples` makes sense if they want the intellisense experience.  `{}` seems like a nice one to add by default, but it's just my opinion.  Likewise, I like `zowe cnfg secure`, but it's also just my opinion.

dankox:
This looks awesome. We always have discussions/suggestions which property should be in which profile (endevor vs. location). This will finally stop all of that :) 
I really like the nesting. Great work guys!

dkelosky:
Here's an example of building the config for those that can't easily test this PR:

intellisense

MarkAckert:
These changes look really cool, I'm looking forward to trying them out 👍 . 

Two thoughts based solely on reading through the doc are below. IMO this PR should go forward regardless...too many benefits.

* UX/Learning Curve - With the profiles being type-less/free-form, I can see there being a little more user error in working with profiles or discrete fields (especially if fields are set through command-line parameters). Similarly, there's _potentially_ less "shared" errors and experiences between users since they may use drastically different schema - resulting in less "searchable" errors and a steeper learning curve.

* UX / Contexts - Does it make sense to expand `defaults` to include additional coarsely-grained contexts? That is, it would be great if I could have groups of profiles wrapped up into "contexts", and quickly switch between these contexts.

tucker01:
Thanks @MarkAckert 

Your points are well taken.

Ultimately, the profile is translated into the CLI arguments, maybe providing a mechanism to dump the arguments and the locations from which they came might help eliminate some confusion? 

As far as groupings, that is partially the purpose of nesting profiles. For example, you could start with a root level profile called `qa` or `dev` and list all applicable services under those. But on the flip side, it might make more sense to organize your root level profiles by host or by service and then what you're suggesting would help. I could group them into `qa` or `dev`. 

Something to think about for sure.

MarkAckert:
> Thanks @MarkAckert
> 
> Your points are well taken.
> 
> Ultimately, the profile is translated into the CLI arguments, maybe providing a mechanism to dump the arguments and the locations from which they came might help eliminate some confusion?
> 
> As far as groupings, that is partially the purpose of nesting profiles. For example, you could start with a root level profile called `qa` or `dev` and list all applicable services under those. But on the flip side, it might make more sense to organize your root level profiles by host or by service and then what you're suggesting would help. I could group them into `qa` or `dev`.
> 
> Something to think about for sure.

After thinking a little more, I'm not sure whether the "contexts" make any sense. I think they're well represented by the combination of project settings and global settings. If you're working with Mainframe applications, you're probably project-oriented when using the CLI and the "context" you want aligns naturally with the project settings. 

The counter-argument I can think of is: I frequently run the CLI from "random" directories to debug ad-hoc situations, and I switch back and forth between multiple systems. I want to quickly specify context on-the-fly (`zowe cnfg set-context xyz-system-services`, `zowe cnfg set-context abc-system-services`). This amounts to ignoring project settings completely, and focusing on global+user settings. 

I don't know enough about how people use the CLI today, but project settings are probably the better path. The ad-hoc situations are increasingly and better covered by the VSCode extension.

phaumer:
This looks really great. Sorry, I did not try it yet or did a detailed code walk-through, so I two questions about how it works:

(1) Do I understand this correctly that every user has a copy of the schema in ~/.zowe/schema.json and when a new plugin gets installed that that schema file will be dynamically updated with new profile type and property definitions? Will there also be an API to update the schema programmatically? In Zowe Explorer we did not require that the user actually has Zowe CLI and the plugins installed and called `CliProfileManager.initialize()` to set-up the initial .zowe directory. Could there (or is there) a way to also contribute additional json schema through such a call that would prepare the directory and schema.json file even if there is no plugin installed locally? (The Zowe Explorer VS Code extensions would come with that code).

(2) Given now that all profiles are in one file and the secure properties are even singled out in their own array, could they now be stored in keytar now also as just one array of objects with one key only, to avoid the issue on MacOS that prompts for permission to access the keyring for each an every property one by one (e.g. 20 prompts for 10 profiles with two secure properties):  https://github.com/zowe/zowe-cli-scs-plugin/issues/30

Alexandru-Dumitru:
Pretty cool!

I don't want to pile on with additional requests, but I was thinking if we could have another parameter like `alias` for profiles, similar to `defaults`. In the above depicted format, we would have to pass something like `--endevor-profile lpar1.endevor.prd.p2` but I would like to have `--endevor-profile endevor2`
```
"defaults": {
        "zosmf": "lpar1",
        "endevor": "lpar1.endevor.prd"
    },
"aliases": { // or "alias" singular
        "endevor1": "lpar1.endevor.prd.p1"
        "endevor2": "lpar1.endevor.prd.p2"
}
```
My case is that I generally use the default profiles, but seldomly I switch the command with `--xxx-profile` flag. In this case, I would like to keep using the same syntax as before, by using the profile name and not pass the whole nested structure. 

I know this can already be achieved by creating a new 1st level profile entry with all required parameters, but this would mean to have duplicate entries, right?. I am also aware that this is not implemented in imperative, so the profile loading process would need to also check this field, but since we (I mean you) redo this, why not slide this in, if it's not to much work. 😅

This would be more like syntactic sugar for multi-level nesting of profiles. 
I hope I am making some sense, and please correct me if I got something wrong.

Other than that, I can't wait to work with this new structure and thanks for working on this @tucker01.

PS: Googling and typing at the same time. Can this be achieved with #ref? would this syntax work?

```
"profiles": {
    "big_nested_lpar1": {...},
    "aliasToEnd1" : "#/profiles/big_nested_lpar1/endevor/prd/p1",
    "aliasToEnd2" : "#/profiles/big_nested_lpar1/endevor/prd/p2",
}
```
If this is achievable in this way, I am good 😋 .

tucker01:
Thanks @phaumer. Let me try to address the questions:

1. The schema is generated on demand by `zowe cnfg schema` and includes whatever plugins you have installed. The command outputs the schema to stdout so you can redirect to a file of your choice. 

With that being said, if you are willing, would you mind "fleshing out" some specs on what would make the most sense for Zowe explorer and schema generation? I'm not aware of all the pieces and parts of imperative or Zowe CLI that Zowe explorer currently exploits. 

2. This is a great suggestion. I'll mess around and get back to you.

tucker01:
@Alexandru-Dumitru Thanks!

The `alias` concept makes perfect sense. Let me experiment a bit and I'll get back to you.

gejohnston:
I realize that our emphasis has been on letting VS Code intellisense inform an end user of errors as (s)he types. With everything in one file, the file could get pretty large. It might prove useful if we had a command that validated the user's config file against his/her schema.

`zowe cnfg validate my.config.json my.schema.json`

A blog https://davidwalsh.name/json-validation displayed some pretty simple code that uses the utility https://github.com/kriszyp/commonjs-utils to perform validation. It looks like it would be pretty easy to wrap that into a command. I make no claims about the quality of this utility, but there are probably other ways to get the same result.

tucker01:
@gejohnston we actually have this capability in the current profile code: https://github.com/zowe/imperative/blob/741f24f7246028879317dbd9590b91d7efbdef39/packages/profiles/src/abstract/AbstractProfileManager.ts#L994

It uses: https://www.npmjs.com/package/jsonschema 

This schema is validated upon profile load. 

A command sounds good.

packages/cmd/src/CommandProcessor.ts
```diff
@@ -707,19 +708,112 @@ export class CommandProcessor {
...
+            // "fake out" the cli util to only populate options for profiles
+            // that have not been fulfilled by the config
+            const p: ICommandProfile = {
+                required: [],
+                optional: [],
+                suppressOptions: this.definition.profile.suppressOptions
+            };
+
+            if (this.definition.profile.required)
+                p.required = this.definition.profile.required.filter(type => fulfilled.indexOf(type) < 0);
```
gejohnston, line 813: I may be missing something, but it looks like p.required and p.optional are set, but never used.

tucker01: Good catch, thank you. The "faked" definition is supposed to be passed to `CliUtils.getOptValueFromProfiles` to avoid including options from profiles that are already "fulfilled" by the config. I'll update in my next commit. 

I'll defer to my PR disclaimer `NOTE: Not everything may function exactly as described... let me know if you have issues. Tests have yet to be written.` 😅

packages/imperative/src/cnfg/cmd/set/set.handler.ts
```diff
@@ -0,0 +1,71 @@
...
+            try {
+                value = JSON.parse(value);
+            } catch (e) {
+                throw new ImperativeError({ msg: `could not parse JSON value: ${e.message}` });
+            }
+        }
+
+        // Set the value in the config, save the secure values, write the config layer
+        config.set(params.arguments.property, params.arguments.value, {
+            secure: params.arguments.secure,
```
tjohnsonBCM, line 66: This PR looks awesome!

With the current implementation, is it true that if a user forgets to provide the `--secure` flag, they could accidentally store a credential in plain text? Currently, once the SCS plugin is installed, any fields designated as secure in profile schemas will always be stored securely.

tucker01: Thanks @tjohnsonBCM!

That is correct. `config` profiles are simple key/value pairs and there is no concept of "profile schema". 

The intention is to create the config by hand. However, I wanted a generic mechanism to set secure fields... So funnily enough, setting secure fields is the reason `zowe cnfg set` exists.

I don't see why someone would prefer `zowe cnfg set` over editing by hand (or editing the JSON programmatically), so perhaps we could replace `zowe cnfg set` with something like `zowe cnfg secure <property path> <value>`? 

I think that would eliminate potential confusion... i.e. why would one even use `zowe cnfg set`? ... and hopefully addresses your concern of "accidentally" saving sensitive info in plain text.

What do you think? 

Also, @dkelosky was looking for a mechanism to completely create the config by hand and then secure fields specified in `secure`. I'd still like to support the request, but after a config is created/secured, we'll need a mechanism to update secure fields (for example, password expiration).

tucker01:
@phaumer I've implemented a single load/save for secure fields. It will access the secure vault once during imperative initialization and a single vault entry contains all secure fields for the "app" (e.g. `zowe` cli). 

Hopefully I didn't regress any previous functionality in this PR... I'll do some manual tests.

gejohnston:
I find the multiple "path" properties confusing.  Sometimes it is a file system path to a config file. Sometimes it is dotted path to a property like "a.b.c.PropertyName"

When I encounter code like the following, it is difficult to know which path we are talking about.

someObject.path = someOtherObject.path

It would be hard to debug if someone mistakenly assigns one type of path to a property that is supposed to be the other type of path.

tucker01:
> I find the multiple "path" properties confusing. Sometimes it is a file system path to a config file. Sometimes it is dotted path to a property like "a.b.c.PropertyName"
> 
> When I encounter code like the following, it is difficult to know which path we are talking about.
> 
> someObject.path = someOtherObject.path
> 
> It would be hard to debug if someone mistakenly assigns one type of path to a property that is supposed to be the other type of path.

Hm... would more verbose names help? Or perhaps type aliases?

gejohnston:
I was thinking that perhaps more distinctive names would help, like some abbreviations of filePathName and propertyPathName.

The code is already a little confusing due to lots of nested references to items named "properties" and "profiles". But that is due to the nesting of objects in the config, and therefore, I think, unavoidable. The two "path" properties are distinct and could be helped by using different names.

Disclaimer: Perhaps I am just whining about little things while I try to figure out what is going on in the app.

tucker01:
> I was thinking that perhaps more distinctive names would help, like some abbreviations of filePathName and propertyPathName.
> 
> The code is already a little confusing due to lots of nested references to items named "properties" and "profiles". But that is due to the nesting of objects in the config, and therefore, I think, unavoidable. The two "path" properties are distinct and could be helped by using different names.
> 
> Disclaimer: Perhaps I am just whining about little things while I try to figure out what is going on in the app.

I can definitely update to have more distinctive vars. I haven't added any jsdoc just yet, which might help with arguments, etc.

packages/imperative/src/cnfg/cmd/set/set.handler.ts
tjohnsonBCM: I like the idea of naming the command `zowe cnfg secure` to eliminate potential confusion.

Perhaps instead of removing `zowe cnfg set`, it could be kept for setting non-secure settings. While I agree that it would normally be preferrable to edit JSON manually or programmatically, I could see value in having commands similar to NPM for shell scripts to use: `zowe cnfg get`, `zowe cnfg set`, `zowe cnfg delete`, etc.

t1m0thyj:
This PR was superseded by #452 which has been merged into next branch.

PR #451, opened 2020-10-26
Updates for Node 14

t1m0thyj:
No need to publish a new version after merge, as the only changes are:
* Update dev dependencies for Keytar tests and Jest test reporters
* Update GitHub workflows to use Node 10, 12, 14 instead of 8, 10, 12


awharn (approved):
Looks good to me!

PR #452, opened 2020-10-27
JSON config in next

tucker01:
Moving the JSON config to next, see the original PR for details: #450 

Changes will be noted in this PR.


tucker01:
Based on the convo with @t1m0thyj in the original PR, I renamed `zowe cnfg set` to `zowe cnfg secure`. 

I'm on the fence about `set`, `get`, and `delete`. I see the value for targeting the specific config files (global vs user, etc.), but "dealing" with JSON in scripts is nbd. Trying to stick to the bare minimum and go with feedback from there.

gejohnston:
I read some of the source code related to these changes (still much more to read and understand). I also built and experimented with the functionality. As I did so, I wrote down thoughts that occurred to me about the design and implementation.

Disclaimer: These are knee-jerk reactions that may not be fully thought out. The intention is to focus the Engineering team on questions that we may need to resolve to implement a deliverable product.

The link to my document is below. Perhaps others will add confirmation or differing perspectives into this PR on some of the items from that document.

https://docs.google.com/document/d/1BtnMkXqOMELSMya7rWm6etF3g5z_Xx2hYNbCcYOhOvw/edit?usp=sharing

tucker01:
Thanks @gejohnston 

(4) in the list really sticks out to me. I'll have to think about this.

tucker01:
I created a branch for the "active" impl: https://github.com/zowe/imperative/tree/json-config-next-active 

I'll keep both to evaluate for now... However, "active" addresses a few shortcomings and provides better reuse of "common" config (like user creds). 

### Small example
**zowe.config.json**
```
{
    "$schema": "./schema.json",
    "profiles": {
        "host1": {
            "properties": {
                "host": "host1",
                "port": 123
            }
        },
        "host2": {
            "properties": {
                "host": "host2",
                "port": 1234
            }
        },
        "all": {
            "properties": {
                "rejectUnauthorized": false
            }
        }
    },
    "plugins": [],
    "active": [
        "all"
    ],
    "secure": []
}
```

**zowe.config.user.json**
```
{
    "$schema": "./schema.json",
    "profiles": {
        "creds": {
            "properties": {
                "user": "user",
                "password": "password",
                "account": "12345678"
            }
        }
    },
    "active": [
        "creds",
        "host1"
    ],
    "plugins": [],
    "secure": []
}
```

tucker01:
After a chat with the CLI team, it seems like we can solve the sharing of credentials with the `base` profile. The consensus was, we'll remove `properties` and keep `defaults`. Users would then specify "global" credentials in a base profile and set it as the default in their global config.

t1m0thyj:
Renamed `secure` command that sets a single secure property to `set-secure`.

Added new `secure` command that prompts for values for all fields designated as secure in `zowe.config.json`

tucker01:
The `properties` property has been removed

packages/config/src/Config.ts
```diff
@@ -0,0 +1,666 @@
...
+            defaults: {},
+            plugins: [],
+            secure: []
+        };
+    }
+
+    public static async load(app: string, opts?: IConfigOpts): Promise<Config> {
+        opts = opts || {};
+
+        ////////////////////////////////////////////////////////////////////////
```
dkelosky, line 69: A petty observation - but I think this is the only file in imperative using this block commenting style.  It's no problem, but is certainly eye catching in that it doesn't fit with our other source files.

packages/config/src/Config.ts
```diff
@@ -0,0 +1,666 @@
...
+import { IConfigLoadedProfile, IConfigLoadedProperty } from "./doc/IConfigLoadedProfile";
+
+enum layers {
+    project_user = 0,
+    project_config,
+    global_user,
+    global_config
+};
+
+export class Config {
```
dkelosky, line 36: could we / should we have some js doc in this file ?

packages/config/src/Config.ts
```diff
@@ -0,0 +1,666 @@
...
+        return new class {
+
+            ////////////////////////////////////////////////////////////////////
+            ////////////////////////////////////////////////////////////////////
+            // Profiles API
+            ////////////////////////////////////////////////////////////////////
+            ////////////////////////////////////////////////////////////////////
+
+            // tslint:disable-next-line
+            public profiles = new class {
```
dkelosky, line 186: there are several tslint rules disabled - is it for inner classes?  would it be "better" to have several, separate class files for ease of testing?

packages/config/src/Config.ts
```diff
@@ -0,0 +1,666 @@
...
+    }
+
+    // TODO Does this need to recurse up through nested profiles?
+    private loadProfile(path: string): IConfigLoadedProfile {
+        const profile = lodash.get(this.properties, path);
+        if (profile == null) {
+            return null;
+        }
+
+        const loadedProfile = require("lodash-deep").deepMapValues(profile, (value: any, p: string) => {
```
dkelosky, line 588: should this be an import with types?

packages/config/src/Config.ts
```diff
@@ -0,0 +1,666 @@
...
+
+    public static async load(app: string, opts?: IConfigOpts): Promise<Config> {
+        opts = opts || {};
+
+        ////////////////////////////////////////////////////////////////////////
+        // Create the basic empty configuration
+        const _ = new Config();
+        (_ as any).config = {};
+        _._layers = [];
+        _._home = opts.homeDir || node_path.join(require("os").homedir(), `.${app}`);
```
dkelosky, line 74: should `os` be imported similarly to `import * as fs from "fs";` above to be used here and line 86?

packages/config/src/Config.ts
```diff
@@ -0,0 +1,666 @@
...
+        return loadedProfile;
+    }
+
+    ////////////////////////////////////////////////////////////////////////////
+    ////////////////////////////////////////////////////////////////////////////
+    // Static Utilities
+    ////////////////////////////////////////////////////////////////////////////
+    ////////////////////////////////////////////////////////////////////////////
+
+    public static search(file: string, opts?: any): string {
```
dkelosky, line 626: maybe we could have an interface for the opts here

packages/config/src/Config.ts
```diff
@@ -0,0 +1,666 @@
...
+        opts = opts || {};
+        if (opts.stop) opts.stop = node_path.resolve(opts.stop);
+        if (opts.gbl) opts.gbl = node_path.resolve(opts.gbl);
+        const p = findUp.sync((directory: string) => {
+            if (directory === opts.stop) return findUp.stop;
+            if (directory === opts.gbl) return;
```
dkelosky, line 632: should this be a `findUp.stop` too?

packages/imperative/src/config/cmd/init/init.handler.ts
```diff
@@ -164,14 +164,14 @@ export default class InitHandler implements ICommandHandler {
...
-     * Prompts for the value of a property on the CLI. Returns null if `--ci`
+     * Prompts for the value of a property on the CLI. Returns null if `--prompt false`
      * argument is passed, or prompt times out, or a blank value is entered.
      * @param propName The name of the property
      * @param property The profile property definition
      */
     private async promptForProp(propName: string, property: IProfileProperty): Promise<any> {
         // skip prompting in CI environment
-        if (this.arguments.ci) {
+        if (this.arguments.prompt === false) {
```
dkelosky, line 174: can we revisit this on standup?  i'm wondering about confusion with using `prompt=false` and i kind of liked the simpler flag `--ci`.  OR, how about something like `--no-prompt` ?

t1m0thyj: `--no-prompt` is also supported. For a Boolean arg `--x`, Yargs will automatically convert `--no-x` to `--x false`

dkelosky: image

dkelosky: giphy

dkelosky: i never knew this!

packages/config/src/Config.ts
t1m0thyj: No I believe this is deliberate. Should probably have added a comment here though 😢

If we matched the `opts.stop` directory, we want to return the directory we found and look no further.
If we matched the `opts.gbl` directory, we want to return `null` - there is no local config to load if it's in the same folder as global config.

dkelosky: if i have home as `C:\Users\dan\` and global as `C:\Users\dan\.zowe`, then I run my project from `C:\Users\dev\` do we just search to root?

t1m0thyj: Yes that's correct (had to test this 😋)

dkelosky: thanks! 

i'm curious on the reasoning then - is the thought we stop sooner in "most" cases?  (i'm not saying it's a bad idea; i'm just trying to understand)

t1m0thyj: I'm open to changing it 😁 

@tucker01 Could you explain the reasoning behind stopping in user home folder? As opposed to other tools like NPM that search for node_modules all the way up to the root of your drive.

tucker01: No idea, feel free to change it

t1m0thyj:
This PR should be ready to merge into `next` after #530 is merged.

I've created GitHub issues with the "team-profile" label for all TODOs in the code and unresolved comments on this PR.

Please review the latest commit https://github.com/zowe/imperative/pull/452/commits/47fd14a7dd0b2bd1a27d8187c3901f90c5572ab2, where I tried to summarize the changes of this PR in the [CHANGELOG](https://github.com/zowe/imperative/blob/json-config-next/CHANGELOG.md#recent-changes).

dkelosky:
Potentially we could add to the changelog that daemon mode supports prompting now

zFernand0:
> Potentially we could add to the changelog that daemon mode supports prompting now

Did I miss the daemon prompt support in the changelog?

Talking about changelog, Once `next` goes into `latest` should we condense all the timestamp changelogs for `5.0.0` into one?
Or should we keep the history? :yum:

t1m0thyj:
> Did I miss the daemon prompt support in the changelog?

Discussed with @dkelosky, I think the plan here is to wait to announce it in Zowe CLI changelog.

> Talking about changelog, Once `next` goes into `latest` should we condense all the timestamp changelogs for `5.0.0` into one?
> Or should we keep the history? 😋

I would vote for merging them into one to remove all traces of "next" at that point. 🙂

dkelosky:
> Talking about changelog, Once next goes into latest should we condense all the timestamp changelogs for 5.0.0 into one?
Or should we keep the history? 😋

It seems like they would all be condensed into major release notes 😄

PR #454, opened 2020-10-28
Allow Hidden Options

dkelosky:
Allow hidden options for command definitions and define a hidden global option for daemon usage, e.g. `hidden: true`:
```
        definition.options.push({
            name: Constants.DAEMON_CLIENT_DIRECTORY,
            group: Constants.GLOBAL_GROUP,
            description: "Daemon client directory",
            type: "string",
            hidden: true
        });
```


packages/cmd/src/help/abstract/AbstractHelpGenerator.ts
```diff
@@ -163,25 +163,28 @@ export abstract class AbstractHelpGenerator implements IHelpGenerator {
...
         }
         for (const option of this.mCommandDefinition.options) {
-            const group = option.group;
-            if (!this.groupToOption[group]) {
-                this.groupToOption[group] = [];
-            }
-            const caseSensitive = this.getCaseSensitiveFlagByOptionName(option.name);
-            const optionAndAliases = this.getOptionAndAliasesString(option, caseSensitive);
-            this.groupToOption[group].push(optionAndAliases);
+            if (!option.hidden) {
```
t1m0thyj, line 166: If this were written as:
```typescript
for (const option of this.mCommandDefinition.options.filter(opt => !opt.hidden)) {
```
it would eliminate the extra level of indentation added here.

Curious what you think about this? It definitely doesn't need to be changed 🙂

packages/cmd/src/CommandProcessor.ts
```diff
@@ -842,6 +854,7 @@ export class CommandProcessor {
         this.log.info(`Command "${this.definition.name}" completed with success flag: "${json.success}"`);
         this.log.trace(`Command "${this.definition.name}" finished. Response object:\n${inspect(json, { depth: null })}`);
         response.endStream();
+        process.chdir(this.mCwd);
```
t1m0thyj, line 857: Should this be wrapped in an `if (params.arguments.dcd)` clause, like the `process.chdir` above? I know it shouldn't matter here if we're `chdir`-ing to the original `cwd`, but IMO it seems ideal to keep `chdir` calls to a minimum.

t1m0thyj (approved):
LGTM ༼ つ ◕_◕ ༽つ Left a few questions in the comments

dkelosky:
both changes make sense to me - thanks for reviewing so quickly !

packages/cmd/src/help/abstract/AbstractHelpGenerator.ts
dkelosky: fix in last commit

packages/cmd/src/CommandProcessor.ts
dkelosky: i changed this to set a flag if `chdir` is done and only undo it if flag is set.  does that work for you?

t1m0thyj: Looks good 🙂

PR #455, opened 2020-10-28
Fix node-pty test hanging on Node 14 + Windows build

### LINKED ISSUE #453: Prompting integration test hangs on Windows + Node 14
The test in "cli.prompting.integration.test.ts" times out (or hangs in GH Actions). This may be related to the `node-pty` dev dependency, as it's the only test that uses this native module.

t1m0thyj:
⚠️ Don't merge yet, this worked locally but is still hanging in pipeline.

* Resolves #453. The only solution I could find was to kill the `node-pty` process before the "exit" event is emitted, otherwise it will hang indefinitely on Node 14 + Windows. Also changed test to listen for "exit" event instead of "error" and "end", which are undocumented and I couldn't get to work.
* Replaces dev dep `node-pty` with `node-pty-prebuilt-multiarch`. This package includes prebuilt native code like Keytar does, and makes `node-gyp` unnecessary for most users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,6 @@ Imperative CLI Framework is a command processing system that lets you quickly bu
 
 - [**Install Node.js package manager**](https://nodejs.org/en/download/package-manager) on your computer. Node.js® is a JavaScript runtime environment on which we architected Imperative CLI Framework.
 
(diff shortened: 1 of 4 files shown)


t1m0thyj:
Closing this PR, will open a new one that disables the offending test on Windows for now.

Issue #444, opened 2020-09-15
Add truncate option to TextUtils.getTable()


mheuzey:
I would like a `truncate` option for TextUtils.getTable(). The default behavior is to not do any wrapping/truncating. For very wide tables, this can cause several lines of wrapping per table row, which makes the table very unreadable. The `hardWrap` option will truncate each line, but also causes each column to have a max width. For a wide table with a large number of columns, each column gets limited to 1 character, and the net effect is vertical text, similar to consoles from the Matrix.

What I would like is an option to keep columns at their default width, but truncate the displayed row.

As a workaround, I could use TextUtils.wordWrap() to manually truncate the lines myself, but the `wrap-ansi` package trims whitespace by default, and the option to not trim whitespace isn't exposed by the wordWrap function.

PR #458, opened 2020-11-10
Expose trim parameter within TextUtils.wordWrap()

mheuzey:
When using `TextUtils.wordWrap()` to truncate lines, as in #444, whitespace between columns is removed. This change exposes the `trim` parameter from `wrap-ansi`, in order to prevent this behavior.

This is an optional parameter, and the default value from `wrap-ansi` is kept, so no existing code is affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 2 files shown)


zFernand0 (approved):
LGTM

zFernand0:
thanks for fixing the conflicts
`codecov/project` is complaining about a file that wasn't touched on this PR.
Merging

PR #464, opened 2020-11-17
Config and default cred mgr

gejohnston:
**Modifications**

Reuse the old, now-irrelevant, "config" group for the new single-profile config group of commands.

Handle secure storage of properties with the Credential Manager built into imperative. Once paired with the version of zowe-cli that includes keytar, secure storage will always, automatically be available for use. As a result, we eliminated the unneeded ability to override the credential manager. Any app that relies on imperative's secure storage must supply its own copy of Keytar. Zowe Explorer is such an app. Plugins will automatically get Keytar from zowe-cli.

- Remove old "config" command group.
- Change single profile "cnfg" command group to "config".
- Use only the default credential manager that is built into imperative.
- Incorporate newer logic from SCS plugin into imperative's built-in credential manager.
- Change the credential service name to "zowe" and the account to "secure_config_props".

Once this PR is merged into the json-config-next branch of imperative, and is used with the json-config-next branch of zowe-cli, full testing of the single-profile POC can begin.

**Unfinished work for future PRs**
- Fix broken automated tests. Note that tests are also already broken in the json-config-next branch that we would merge into.
- Stop creating old-school profiles directory and meta files.
- Implement a scheme to store secure content for multiple project configs.
- Implement a scheme to convert old-school profiles and secure content into single profile.


packages/security/src/DefaultCredentialManager.ts
```diff
@@ -49,156 +51,224 @@ import * as keytar from "keytar"; // Used for typing purposes only
...
+   * @throws {@link ImperativeError} when keytar.deletePassword returns false.
+   */
+  protected async deleteCredentials(account: string): Promise<void> {
+    this.checkForKeytar();
+
+    if (!await this.deleteCredentialsHelper(account)) {
+      throw new ImperativeError({
+        msg: "Unable to delete credentials.",
+        additionalDetails: this.getMissingEntryMessage(account)
+      });
```
t1m0thyj, line 130: In the SCS plugin we made a change to stop throwing this error. Reasons for doing so are listed in [this PR comment](https://github.com/zowe/zowe-cli-scs-plugin/pull/7).

Not necessarily suggesting any action here, as perhaps the scenarios listed there are no longer applicable now? If we're storing only a single secure value, I don't think we'll be deleting values from the secure vault too frequently anymore.

packages/security/src/DefaultCredentialManager.ts
```diff
@@ -49,156 +51,224 @@ import * as keytar from "keytar"; // Used for typing purposes only
...
+   * Calls the keytar getPassword service with {@link DefaultCredentialManager#service} and the
+   * account passed to the function by Imperative.
+   *
+   * @param {string} account The account for which to get credentials
+   * @returns {Promise<SecureCredential>} A promise containing the credentials stored in keytar.
+   *
+   * @throws {@link ImperativeError} if keytar is not defined.
+   * @throws {@link ImperativeError} when keytar.getPassword returns null or undefined.
+   */
+  protected async loadCredentials(account: string): Promise<SecureCredential> {
```
t1m0thyj, line 144: The `optional` argument should be added back here. I believe removing it is a breaking change.

Without this argument, CLI commands fail if an optional secure field (such as "secure_config_props") cannot be loaded. (An example of when this happens is in an Ubuntu Docker container, if Keytar fails to load and zowe.config.json has a non-empty "secure" array.)

This commit shows the behavior of the `optional` argument when it was added: https://github.com/zowe/zowe-cli-scs-plugin/pull/11/commits/f35697d8baddfa78bafe8c06cef0e42afc96da39

t1m0thyj (changes requested):
New "config" group LGTM, left a few comments about DefaultCredentialManager.

packages/security/src/DefaultCredentialManager.ts
gejohnston: I copied logic from the correct, newer, SCS repo. It contained the logic that you identified. I also altered some items that we concluded in team meetings. If you see anything else that was added in the past which I still may have overlooked, let me know.

packages/security/src/DefaultCredentialManager.ts
gejohnston: Same comment as below.

awharn:
I am not sure if this is a problem to resolve in this PR, but when paired with the json-config-next branch with the CLI, I get the following error:

```
[2020/11/19 20:11:17] [FATAL] [main.js:48] Error initializing Zowe CLI:
 ImperativeError: UNEXPECTED ERROR ENCOUNTERED
    at Function.<anonymous> (/home/jenkins/imperative/lib/imperative/src/Imperative.js:271:21)
    at Generator.throw (<anonymous>)
    at rejected (/home/jenkins/imperative/lib/imperative/src/Imperative.js:16:65) {
  mDetails: {
    msg: 'UNEXPECTED ERROR ENCOUNTERED',
    causeErrors: [Error: Error calling StartServiceByName for org.freedesktop.secrets: GDBus.Error:org.freedesktop.DBus.Error.Spawn.ExecFailed: Failed to execute program org.freedesktop.secrets: Operation not permitted]
  },
  mMessage: 'UNEXPECTED ERROR ENCOUNTERED'
```

This may be something that gets addressed by a future PR, but wanted to at least bring it up. 

This system in particular has Keytar installed, but is running Linux without a desktop environment, resulting in a failure to load secrets from the credential vault. The config file was manually changed to have an entry in the secure fields array. When this is the case, the CLI fails to run any command, regardless of whether or not it needs a credential, until the secure property is removed.

t1m0thyj (approved):
CredentialManager looks good to me now and works on my machine without "secure_config_props" needing to exist.

There is the outstanding issue that @awharn mentioned, not sure if it should be fixed in this PR or documented as an issue to resolve later?

gejohnston:
I have encountered the inability to run any zowe command when Windows CredentialManager is in a transient state of "A specified logon session does not exist. It may have been terminated".  I do not know if the problem would have occurred in the original POC, or if the latest changes to this PR fix that problem. It is hard to know since the condition is transient (and I avoid anything that might re-create that condition).

Since Andrew has identified a reproduceable error, I suggest that he open a Git issue (or create a Rally story or task).

If it is alright with others, I would prefer that we create a steady stream of smaller pull requests into json-config-next, to make the reviews easier and the divergence in our individual branches smaller.

Voice any dissent about merging this PR. I will wait until at least tomorrow to merge it.

awharn (approved):

PR #469, opened 2020-11-30
[json-config-next] Handle long credentials on Windows

t1m0thyj:
This attempts to work around the Windows limitation that does not allow credentials longer than 2,560 bytes.

The secure properties object is stored more concisely than before (click arrows to expand examples):
<details>
<summary>Old</summary>
<pre>
{
    "configs":[
        {"path":"C:\\dev\\imperative\\zowe.config.user.json","properties":[]},
        {"path":"C:\\dev\\imperative\\zowe.config.json","properties":[]},
        {"path":"C:\\Users\\me\\.zowe\\zowe.config.user.json","properties":[
            {"path":"profiles.base.properties.password","value":"XXXXXXXX"}
        ]},
        {"path":"C:\\Users\\me\\.zowe\\zowe.config.json","properties":[]}
    ]
}
</pre>
</details>

<details>
<summary>New</summary>
<pre>
{
    "C:\\Users\\me\\.zowe\\zowe.config.user.json":{
        "profiles.base.properties.password:"XXXXXXXX"
    }
}
</pre>
</details>

In addition, if running on Windows and the length of a credential exceeds 2,560 bytes, it is split into multiple chunks and stored across multiple fields of the vault. Each chunk is 2,560 bytes long, except the last one which is terminated with a NUL character (`\0`) to indicate end of value. It will look like this in the credential vault:
image

The way credentials are loaded/saved has not changed on non-Windows platforms, or for credentials less than 2,560 bytes.

Have not yet written tests for the new code. Plan to save that for a future PR, to keep this one small and so the implementation can be reviewed first.


gejohnston (approved):
This looks good.
It should enable us to store all real-world sized secure properties.

PR #470, opened 2020-12-01
Prevent TypeError because getFileName() return null and other unexpected problems

### LINKED ISSUE #449: Logger fails when logging an error message: Cannot read property 'indexOf' of null
It seems that the logger fails when an element of the stack trace does not have a file name.

```
TypeError: Cannot read property 'indexOf' of null
at Logger.getCallerFileAndLineTag (/u/tnzsys/zowe-test/node_modules/@zowe/imperative/lib/logger/src/Logger.js:325:53)
at Logger.debug (/u/tnzsys/zowe-test/node_modules/@zowe/imperative/lib/logger/src/Logger.js:175:40)
at Logger.timerified debug (<anonymous>)
at /u/tnzsys/zowe-test/node_modules/@zowe/imperative/lib/rest/src/client/AbstractRestClient.js:108:26
at new Promise (<anonymous>)
at ZosmfRestClient.request (/u/tnzsys/zowe-test/node_modules/@zowe/imperative/lib/rest/src/client/AbstractRestClient.js:70:16)
at Function.putExpectString (/u/tnzsys/zowe-test/node_modules/@zowe/imperative/lib/rest/src/client/RestClient.js:210:34)
at Function.<anonymous> (/u/tnzsys/zowe-test/node_modules/@zowe/imperative/lib/rest/src/client/RestClient.js:68:37)
at /u/tnzsys/zowe-test/node_modules/@zowe/imperative/lib/rest/src/client/RestClient.js:18:71
```

This seems like the line that is causing the problem: https://github.com/zowe/imperative/blob/master/packages/logger/src/Logger.ts#L357

plavjanik:
Resolves #449

It is important for z/OSMF performance test development. This exception masks a real exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
-## `4.9.1`
+## Recent Changes
 
 - BugFix: Fixed an issue when `TypeError` has been raised by `Logger.getCallerFileAndLineTag()` when there was not filename for a stack frame. [#449](https://github.com/zowe/imperative/issues/449)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## `4.9.1`
+
+- BugFix: Fixed an issue when `TypeError` has been raised by `Logger.getCallerFileAndLineTag()` when there was not filename for a stack frame. [#449](https://github.com/zowe/imperative/issues/449)
+
 ## `4.9.0`
 
 - BugFix: Updated `opener` dependency due to command injection vulnerability on Windows - [GHSL-2020-145](https://securitylab.github.com/advisories/GHSL-2020-145-domenic-opener)
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -352,15 +352,19 @@ export class Logger {
      * @returns {string} - file and line number
      */
     private getCallerFileAndLineTag(): string {
-        const frame: StackTrace.StackFrame[] = StackTrace.parse(new Error());
-        let callerStackIndex = 1;
-        while (frame[callerStackIndex].getFileName().indexOf(path.basename(__filename)) >= 0) {
-            // go up the stack until we're outside of the BrightsideLogger file
-            callerStackIndex += 1;
+        try {
+            const frame: StackTrace.StackFrame[] = StackTrace.parse(new Error());
+            let callerStackIndex = 1;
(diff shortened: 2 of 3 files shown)


plavjanik:
@dkelosky Who can help with the changelog and failing tests?

```
Error: The changelog has changed, but the required header is missing.
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## `4.9.1`
```
dkelosky, line 5: This should be `## Recent Changes` (https://github.com/zowe/imperative/blob/master/CONTRIBUTING.md#changelog-update-guidelines) - sorry for the troubles

dkelosky:
After the changelog update - let's see if the tests fail again.

CHANGELOG.md
plavjanik: Thanks. I just tried to follow the existing structure with numbers.

dkelosky: yeah - it's not clear... maybe we can put a hidden note in the changelog itself in the future

dkelosky:
sorry @plavjanik - I missed this over break.

awharn (approved):
This PR looks good - thanks for resolving that problem. There seems to be a merge conflict on `CHANGELOG.md`. Once the conflict is resolved, I believe this can be merged.

packages/logger/src/Logger.ts
```diff
@@ -352,15 +352,19 @@ export class Logger {
...
     private getCallerFileAndLineTag(): string {
-        const frame: StackTrace.StackFrame[] = StackTrace.parse(new Error());
-        let callerStackIndex = 1;
-        while (frame[callerStackIndex].getFileName().indexOf(path.basename(__filename)) >= 0) {
-            // go up the stack until we're outside of the BrightsideLogger file
-            callerStackIndex += 1;
+        try {
+            const frame: StackTrace.StackFrame[] = StackTrace.parse(new Error());
+            let callerStackIndex = 1;
+            while (!frame[callerStackIndex].getFileName() || (frame[callerStackIndex].getFileName().indexOf(path.basename(__filename)) >= 0)) {
```
t1m0thyj, line 358: @dkelosky Not sure if it's safe to assume that `frame[callerStackIndex].getFileName()` will always evaluate to a truthy value eventually? If there is a case where it doesn't, this seems like it could cause an infinite loop.

t1m0thyj (approved):

zFernand0:
LGTM!
Intermittent failure on wash Jenkins I believe. Rerunning build

dkelosky (approved):

PR #471, opened 2020-12-02
Remove LTS-incremental @brightside/core service

gejohnston:
As decided in a team meeting we will no longer attempt to read secure properties for the @brightside/core service which was used in the LTS-incremental branch.

Also, directly place services into allServices in the desired order. It seems simpler to follow the logic.

### DIFF
--- a/packages/security/src/DefaultCredentialManager.ts
+++ b/packages/security/src/DefaultCredentialManager.ts
@@ -58,23 +58,6 @@ export class DefaultCredentialManager extends AbstractCredentialManager {
   public static readonly SVC_NAME = "Zowe";
 
   /**
-   * Alternative services under which we will look for credentials.
-   * Do not change the order of the alternative services.
-   * When adding, removing or modifying the alternative services,
-   * double check the loadCredentials function.
-   */
-  private readonly ALTERNATIVE_SERVICES = ["@brightside/core", "@zowe/cli", "Zowe-Plugin"];
-
-  /**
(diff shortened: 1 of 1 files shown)


packages/security/src/DefaultCredentialManager.ts
```diff
@@ -114,12 +97,27 @@ export class DefaultCredentialManager extends AbstractCredentialManager {
...
+     */
+    if (this.service !== DefaultCredentialManager.SVC_NAME) {
+        this.allServices.push(DefaultCredentialManager.SVC_NAME);
     }
-    this.allServices.push(DefaultCredentialManager.SVC_NAME);
+
+    /* Previous services under which we will look for credentials.
+     * We dropped @brightside/core because we no longer support the
+     * lts-incremental version of the product.
+     */
```
t1m0thyj, line 116: Agree that it makes sense to drop @brightside/core 👍 

Should we also drop the logic below (lines 186-193) that migrated old @brightside/core property names (username -> user, pass -> password)? Perhaps the `loadHelper` defined on line 184 is unnecessary and `this.getCredentialHelper` can be directly called instead.

gejohnston: If the only case for conversion from username -> user, and pass -> password are in LTS-incremental, then I agree that it makes sense to remove them.

I will wait to let other reviewer confirm this, then I will add an update to this PR to remove that logic.

PR #475, opened 2020-12-08
Allow comments in our JSON config.

zFernand0:
Enhancements:
- Allow comments in any `config.json` files
- Normalize any config operation to the `Config.INDENT` value
- Allow for comments to be stored on the secure vault (kind of an Easter Egg 😋 )

Notes:
- Properties in the `secure` array are not stored in alphabetical order.
- Any array added to the `config.json` will be of type `CommentArray` which `extends Array`


t1m0thyj (approved):
LGTM, the thorough tests made it easy to review 🙂 

> * Allow for comments to be stored on the secure vault (kind of an Easter Egg 😋 )

IMO this is unnecessary, since the JSON stored in secure vault is always programmatically generated, and keeping it minified seems ideal since Windows limits the length of secure vault fields. But you did say it was an Easter egg 😋

packages/imperative/src/config/cmd/set-secure/set-secure.handler.ts
```diff
@@ -9,8 +9,9 @@
 *
 */
 
+import * as JSONC from "comment-json";
 import { ICommandHandler, IHandlerParameters } from "../../../../../cmd";
-import { IConfigOpts } from "../../../../../config";
+import { IConfigOpts, Config } from "../../../../../config";
```
awharn, line 14: Is Config used in this file?

awharn (approved):
Looks good to me, just one question/comment

packages/imperative/src/config/cmd/set-secure/set-secure.handler.ts
zFernand0: It's not anymore, and I believe the Config.INDENT property can be made private again.
Working on it!
Good catch!

awharn (approved):

t1m0thyj (approved):
Thanks for addressing comments 😁

PR #480, opened 2020-12-11
[json-config-next] Fixes for Config secure load

t1m0thyj:
This is hopefully an easier PR to review than #478 🙂 - will try to keep them smaller

* Don't throw error in `Config.load` when secure vault fails to initialize. We can prompt for values instead.
* Fix Keytar not being found in CLI node_modules folder. Added a test for this as it was a regression since #464.


awharn (approved):
LGTM 👍🏼

zFernand0 (approved):
Thanks for adding the test 👍

PR #482, opened 2020-12-14
[json-config-next] Add tests to support use case 3

awharn:
The secure and set-secure commands are already implemented, and integration tests have already been written. To finish out the use case 3 story, add handler tests for the `config secure` command. Handler tests for `set-secure` will be added in use case 4, as the command is being modified in that use case to support storing secure credentials insecurely in the JSON.


t1m0thyj (approved):
LGTM! Nice that a big blob in the handler code could be removed 😎

gejohnston (approved):
Looks good to me.

zFernand0 (approved):
LGTM!

PR #484, opened 2020-12-18
[json-config-next] Use Case 4

awharn:
Changes `config set-secure` to `config set`, and adds options to allow setting properties securely or insecurely.
Fixes issue where secure fields from other projects are overwritten in the OS credential vault during secure save
Adds tests around these changes and this use case


packages/security/src/DefaultCredentialManager.ts
```diff
@@ -100,24 +100,15 @@ export class DefaultCredentialManager extends AbstractCredentialManager {
...
-    this.allServices = [this.service];
+    this.allServices = [service || DefaultCredentialManager.SVC_NAME];
 
-    /* In our current implementation, this.service is always
-     * DefaultCredentialManager.SVC_NAME. We keep the logic below,
-     * in case we re-enable overrides in the future.
-     */
-    if (this.service !== DefaultCredentialManager.SVC_NAME) {
-        this.allServices.push(DefaultCredentialManager.SVC_NAME);
+    if (this.defaultService === DefaultCredentialManager.SVC_NAME) {
```
zFernand0, line 105: so if the user specified they own service in which to store credentials, we are not going to load any of the other services?

packages/config/src/Config.ts
```diff
@@ -526,15 +532,19 @@ export class Config {
                 }
             }
 
+            // Clear the entry and rebuild it
+            delete this._secure.configs[layer.path];
+
             // Create the entry to set the secure properties
             if (Object.keys(sp).length > 0) {
                 this._secure.configs[layer.path] = sp;
             }
```
zFernand0, line 541: So, this prevents `secureSave` from storing empty objects in the vault? (while at the same time clearing any empty objects already there?)

zFernand0 (approved):
PR LGTM. Just a couple of questions 😋

packages/config/src/Config.ts
awharn: Yes. The idea is to reduce the size of the object stored on the vault if there are no secure properties, and to delete any properties that were on the vault before if they were all removed.

gejohnston (approved):
Looks good.

PR #491, opened 2020-12-30
[json-config-next] Json config next tokens

awharn:
Adds token support for the single profile. Only uses the new config if it exists - otherwise uses old school profiles.


gejohnston (approved):
Looks ok to me.

PR #492, opened 2020-12-30
[json-config-next] Restore app settings and overrides for backward compatibility

t1m0thyj:
Tested both old-school profiles and new config.json profiles with and without SCS plugin.

If the SCS plugin is installed, secure values in `zowe.config.json` will be stored under the service name "Zowe-Plugin" instead of "Zowe". This seems ok since it is compatible with Imperative's built-in credential manager, and after SCS plugin is uninstalled the name will change to "Zowe".


gejohnston (approved):
Looks good.
Sorry you had to undo a bunch of changes that I made earlier.
At least the changes were just commented out, not completely deleted.

awharn (approved):

PR #493, opened 2020-12-31
[json-config-next] Add fix and enhancement

### LINKED ISSUE #487: zowe config init -h suggested improvements
In the description, `undefined.config.json` and `undefined.config.user.json` should be changed to `zowe.config.json` and `zowe.config.user.json`:
image

Also, let's consider an alias of `g` for `global` and changing `ci` with default of value of `false` to `prompt` with default value of `true`. We could mention in the description that setting this value to false is useful in continuous integration pipelines where such values are usually maintained in the CI tool and passed in via environment variables.

awharn:
Adds enhancement #486 
Fixes #487


gejohnston (approved):
Looks good to me.

t1m0thyj (approved):
LGTM 🙂

PR #499, opened 2021-01-06
[json-config-next] Fixes after restoring app settings and overrides

t1m0thyj:
If new config does not exist, always fall back to the old implementation of `OverridesLoader.loadCredentialManager`. This fixes an issue where after removing the SCS override, Imperative would fail to read old profiles with plain text credentials.


gejohnston (approved):
Looks good to me.

awharn (approved):
Looks good to me.

PR #500, opened 2021-01-06
[json-config-next] Fix #487

### LINKED ISSUE #487: zowe config init -h suggested improvements

awharn:
Actually fixes #487 this time, adding the -g alias to global and changing --ci to --prompt.


t1m0thyj (approved):
LGTM 😺

PR #502, opened 2021-01-07
Sync with/Next (Add Daemon Mode Stuff)

dkelosky:


__tests__/__integration__/imperative/package.json
```diff
@@ -45,7 +45,7 @@
     "js-yaml": "3.13.1",
     "jsonfile": "4.0.0",
     "jsonschema": "1.1.1",
-    "keytar": "7.1.0",
+    "keytar": "6.0.1",
```
t1m0thyj, line 48: Curious why the downgrade of Keytar package?

packages/cmd/src/CommandProcessor.ts
```diff
@@ -370,6 +365,15 @@ export class CommandProcessor {
...
         // Build the response object, base args object, and the entire array of options for this command
         // Assume that the command succeed, it will be marked otherwise under the appropriate failure conditions
+        if (params.arguments.dcd) {
+            // NOTE(Kelosky): we adjust `cwd` and do not restore it, so that multiple simultaneous requests from the same
+            // directory will operate without unexpected chdir taking place.  Multiple simultaneous requests from different
+            // directories may cause unpredictable results
+            process.chdir(params.arguments.dcd as string)
+
+            // reinit config for daemon client directory
+            ImperativeConfig.instance.config = await Config.load(ImperativeConfig.instance.rootCommandName, ImperativeConfig.instance.config.opts);
```
t1m0thyj, line 375: This makes sense for now. 🙂 I'm wondering if in the future we'll want a more efficient way to alter the project directory of the Config class - assuming that global directory never changes.

packages/config/src/Config.ts
```diff
@@ -50,7 +51,7 @@ export class Config {
     private _vault: IConfigVault;
     private _secure: IConfigSecure;
 
-    private constructor() { }
+    private constructor(public opts?: IConfigOpts) { }
```
t1m0thyj, line 54: Curious why `opts` was added here, it seems unused since constructor is empty?

t1m0thyj (approved):
LGTM, left comments in a few places that I'm curious about 🙂

packages/cmd/src/CommandProcessor.ts
t1m0thyj: > a more efficient way to alter the project directory of the Config class

Having this would also be helpful for fixing #506

__tests__/__integration__/imperative/package.json
dkelosky: likely a mistake on my part - ill change this

packages/cmd/src/CommandProcessor.ts
dkelosky: im still thinking about this - but the i think there is a flaw in this logic somewhere ... i'll keep testing but i might need to fix a bug here

packages/config/src/Config.ts
dkelosky: it's empty but still has an assignment built in.  it's used on line 375 in packages/cmd/src/CommandProcessor.ts

PR #510, opened 2021-01-10
Changes to Allow Prompting the Daemon Client

dkelosky:
- Defaults all prompting to be the same timeout (instead of 30s and 15m)
- adds `prompt` API to IHandlerResponseApi
- uses `response.console.prompt` in various handlers instead of calling the prompt method directly
- `prompt` routes to to CliUtils prompting or daemon mode prompting depending on execution environment


packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
```diff
@@ -111,8 +111,9 @@ describe("Configuration Initialization command handler", () => {
         expect(setSchemaSpy).toHaveBeenCalledWith(expectedSchemaObjectNoBase);
 
         expect(promptWithTimeoutSpy).toHaveBeenCalledTimes(1);
+        // Prompting for secure property
         // tslint:disable-next-line: no-magic-numbers
-        expect(promptWithTimeoutSpy).toHaveBeenCalledWith(expect.stringContaining("blank to skip:"), true, 900); // Prompting for secure property
+        expect(promptWithTimeoutSpy).toHaveBeenCalledWith(expect.stringContaining("blank to skip:"), true, undefined);
```
t1m0thyj, line 116: "// tslint:disable-next-line: no-magic-numbers" no longer needed - this also applies to other lines below in the file

t1m0thyj (commented):
Changes LGTM 🙂

The Jenkins build appears to be failing because integration tests time out. There is a test in `imperative.test.cli.auth.login.fruit.integration.test.ts` that purposely waits until a CLI prompt times out. This test used to take ~30 sec to run which was reasonable, but now takes 15 min since the default timeout length was increased. Not sure how we want to handle this.

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
dkelosky: good point - in hindsight I'm not sure I get the point of why we had any of these in here.  I see `4` is used in some of the tests, and I don't know the significance of `4` - so, it's a magic number.  Likewise with other rule exclusions.

I'll try to change all of these

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
```diff
@@ -221,7 +222,8 @@ export abstract class BaseAuthHandler implements ICommandHandler {
 
         const sessCfgWithCreds = await ConnectionPropsForSessCfg.addPropsOrPrompt<ISession>(
             sessCfg, params.arguments,
-            { requestToken: false }
+            { requestToken: false },
+            params
         );
```
zFernand0, line 227: isn't this a bit strange?
why are we passing `params.arguments` and `params` on the same function?
I know we want to maintain backwards compatibility, but isn't this creating technical debt?

packages/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -186,9 +190,12 @@ export class ConnectionPropsForSessCfg {
...
                 while (answer === undefined) {
-                    answer = await CliUtils.promptWithTimeout(
-                        "Enter the port number for your service: "
-                    );
+                    const text = "Enter the port number for your service: "
+                    if (parms) {
+                        answer = await parms.response.console.prompt(text);
+                    } else {
+                        answer = await CliUtils.promptWithTimeout(text);
+                    }
```
zFernand0, line 198: Having a helper function would be nice (to avoid duplication and reduce maintenance overhead) 😋 
Not a required change though. Just being annoying : )

packages/utilities/src/CliUtils.ts
```diff
@@ -396,8 +398,8 @@ export class CliUtils {
      * @param message - The message to display to the user e.g. "Please enter password:"
      * @returns value - the value entered by the user
      */
+    // TODO(Kelosky): prompt* prompting
```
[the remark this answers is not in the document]
zFernand0: probably not? 😋

zFernand0 (commented):
Looks ok to me. 
Left a few comments! 😋

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
dkelosky: id rather change it, not have the technical debt, and break the api - if you're onboarding ill do it 😄

packages/rest/src/session/ConnectionPropsForSessCfg.ts
dkelosky: not annoying - good point

dkelosky: fixed

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
dkelosky: I'm not sure how to proceed here.  

This is an SDK method but has a lot of Integration around CLI behavior.  

It will conditionally prompt or call back for input depending on options.  

It's probably a little awkward to use already since it accepts command arguments (which may or may not be applicable).  

Adding an extra full handler object as an optional cleans things up for our internal usage but bloats the API method signature. 

What do you think is the best way to proceed?

dkelosky: @zFernand0 - how about changing the method signature to pass handler object as optional in the options object. 

Looking at the code more closely and thinking about this, the options object is currently how we control behavior between other SDK clients and the CLI itself.

packages/rest/src/session/doc/IOptionsForAddConnProps.ts
```diff
@@ -53,4 +54,12 @@ export interface IOptionsForAddConnProps {
...
      */
     getValuesBack? (properties: string[]): { [key: string]: any };
+
+    /**
+     * Specifies full command paramter object if running in a CLI environment (versus) as
+     * a Node.js SDK
+     * @type {IHandlerParameters}
+     * @memberof IOptionsForAddConnProps
+     */
+    parms?: IHandlerParameters
```
zFernand0, line 64: 👍

PR #526, opened 2021-01-15
Support "Content-Encoding: gzip" in REST responses

### LINKED ISSUE #318: Request to implement Accept-encoding: gzip header
`AbstractRestClient.performRest()` method should introduce an option (or maybe just do it by default) to include request header `Accept-encoding: gzip deflate` or something similar.

It should also deal with the response, when it comes in such format by checking 
`content-encoding: gzip` or `content-encoding: deflate`.

This can be done by using `zlib` library and pipe-ing response to it, something like this:

```typescript
let stream: http.IncomingMessage | zlib.Unzip = response;
// if compressed, pipe it to uncompression function
if (response.headers['content-encoding'] === 'gzip' || response.headers['content-encoding'] === 'deflate') {
	stream = zlib.createUnzip();
	response.pipe(stream);
}

stream.on('data', chunk => {
	data.push(chunk);
	size += chunk.length;
});

```

Currently it's not possible to do with the functionality provided by imperative, because we don't have access to response headers (or at least I don't know how) therefore not able to see if the response is compressed.
We either have to implement the whole HTTP request, or use default setup which means, the times for download from mainframe are bigger. 

Just for comparison on Endevor Webservice, when using the default method for list elements, it will take over 1 minute for 8k elements, while with this implementation it will take something around 15-20 seconds.
*Note: times can vary, depending on mainframe performance*

t1m0thyj:
Prerequisite for https://github.com/zowe/zowe-cli/pull/906. Also resolves #318.

Note: SonarQube reports 7 code smells, but this PR is not responsible for any of them AFAICT.


awharn (approved):

gejohnston (approved):
Looks good.

zFernand0 (approved):

dankox:
Hi, I'm kinda curious, if we have decompression implemented on top of zowe response, does this affect us? Will it actually decompress and we have to disable our decompression or this needs to be setup separately?

I didn't really go thru the code and didn't confirm, but I felt like asking it might be the fastest way to find out. Thanks for the info.

PR #529, opened 2021-01-19
Json config next deprecated

gejohnston:
**Added a general mechanism to deprecate commands.**

- New property is available in a command definition

  A new property, named deprecatedReplacement, can be added to the command definition of any command or command group to identify that the command has been deprecated. The value of the property is used as the text that is placed after the ‘Recommended replacement:’ phrase in a deprecation message. For example, the property below is now specified on the command definition of the 'zowe profiles list' command.

      deprecatedReplacement: “The 'config list' command”

- Deprecated message in command output

  For the example above, the following deprecated message will be automatically displayed:

      > zdev profiles list zosmf-profiles
      zosmf_gene (default)
      zosmf_gene_basePath
      zosmf_gene_with_everything
      MyNewProfile

      Warning: The command 'profiles list' is deprecated.
      Recommended replacement: The 'config list' command

- Deprecated help text

  A deprecated message and deprecated indicators will automatically be placed in the help text for deprecated commands and command-groups.

  ```
  zdev profiles --help

  DESCRIPTION
    Create and manage configuration profiles

    Warning: This group has been deprecated.
    Recommended replacement: The 'config init' command

  USAGE
    zowe profiles <group>
    Where <group> is one of the following:

  GROUPS
    create | cre      Create new configuration profiles deprecated)
    delete | rm       Delete existing profiles (deprecated)
    list | ls         List existing profiles (deprecated)
    set-default | set Set which profiles are loaded by default (deprecated)
    update | upd      Update existing profiles (deprecated)
    validate | val    Test the validity of a profile (deprecated)
  ```

**Deprecated all of the zowe profiles comands**

In response to the new Zowe V2 team configuration, all of the 'zowe profiles' commands are deprecated. The set of profiles commands are show in the example above. They will continue to work, but will display a deprecated message.


dkelosky:
This works for me (even under "daemon" mode):
image

Perhaps this was already discussed or too late to change, but I wondered if the `Warning` text here should be another color:
image

dkelosky (approved):

packages/cmd/src/help/DefaultHelpGenerator.ts
```diff
@@ -210,13 +211,15 @@ export class DefaultHelpGenerator extends AbstractHelpGenerator {
...
+                summaryText += command.summary || command.description;
 
-                // Mark with the experimental tag if necessary
-                if (command.experimental) {
+                if (command.deprecatedReplacement) {
+                    // Mark with the deprecated tag
+                    summaryText += this.grey(" (deprecated)");
+                } else if (command.experimental) {
+                    // Mark with the experimental tag
                     summaryText += this.grey("(experimental) ");
```
t1m0thyj, line 221: Should this be changed to `" (experimental)"` since it seems to be a suffix now rather than a prefix? (related to above comment)

gejohnston:
About the color of deprecated warning text in help description:

Mike asked that we do not call excessive attention to deprecated commands with bright colors. A new color will also require testing on on all platforms.

I think that leaves us with the same grey color used for the "(deprecated)" indicator.

I will make the deprecated warning text grey.

t1m0thyj (commented):
LGTM, thanks for adding thorough tests 🙂 Just one comment

__tests__/src/example_clis/with_experimental/ExperimentalExample.spec.ts
```diff
@@ -22,14 +22,14 @@ describe("We should provide the ability to set commands as experimental", functi
     it("should show the text 'experimental' when listing experimental commands",
         function () {
             T.findExpectedOutputInCommand(cliBin,
-                ["--help"], ["(experimental) this command is experimental"],
+                ["--help"], ["this command is experimental(experimental)"],
```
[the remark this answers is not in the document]
gejohnston: No space should be in front of (experimental).

packages/cmd/src/help/DefaultHelpGenerator.ts
gejohnston: Yes a space should come before (experimental).

PR #532, opened 2021-01-26
Don't show secure values from config in plain text

### LINKED ISSUE #505: [cli] config list shows password in plain text

t1m0thyj:
Fixes #505. The output of "config list" command now replaces any secure values with the text `(secure value)`.

Also fixes an issue where optional Keytar dependency isn't loaded, when new config JSON file doesn't exist.

Added unit tests that were missing for "config list", "config profiles", and "config schema" command handlers.

### DIFF
--- a/packages/imperative/__tests__/config/cmd/profiles/profiles.handler.test.ts
+++ b/packages/imperative/__tests__/config/cmd/profiles/profiles.handler.test.ts
@@ -71,7 +71,7 @@ const configProps: IConfig = {
     secure: []
 };
 
-describe("Configuration List command handler", () => {
+describe("Configuration Profiles command handler", () => {
     const configMock = jest.fn();
 
     beforeAll(() => {
--- a/packages/imperative/__tests__/config/cmd/profiles/profiles.handler.test.ts
+++ b/packages/imperative/__tests__/config/cmd/profiles/profiles.handler.test.ts
@@ -51,15 +51,15 @@ const configProps: IConfig = {
                 incoming: {
                     type: "imap",
                     properties: {
-                        host: "imap.gmail.com",
-                        port: 993
+                        host: "fakeHost",
+                        port: 143
                     }
                 },
                 outgoing: {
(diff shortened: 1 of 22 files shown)


zFernand0 (approved):
LGTM!

dkelosky (approved):
😭

packages/imperative/__tests__/config/cmd/profiles/profiles.handler.test.ts
```diff
@@ -0,0 +1,101 @@
...
+            },
+            properties: {}
+        }
+    },
+    defaults: {},
+    plugins: [],
+    secure: []
+};
+
+describe("Configuration List command handler", () => {
```
gejohnston, line 74: A minor detail, but this looks like the description copied from the list command handler. Should it say 'profile command handler'?

gejohnston (approved):
It was good of you to add tests for previously ignored commands.
I had one minor observation, but things generally look fine.

t1m0thyj:
> Also fixes an issue where optional Keytar dependency isn't loaded, when new config JSON file doesn't exist.

I've realized there is a problem both before and after this fix 😢 
Before: "config init" fails to store credentials securely, because no new config exists yet to trigger loading the CredentialManager
After: "profiles create" fails to load/store plain text credentials, because Keytar is now a dependency of Zowe CLI

t1m0thyj:
> > Also fixes an issue where optional Keytar dependency isn't loaded, when new config JSON file doesn't exist.
> 
> I've realized there is a problem both before and after this fix 😢
> **Before:** "config init" fails to store credentials securely, because no new config exists yet to trigger loading the KeytarCredentialManager
> **After:** old school profiles fail to load/store plain text credentials, because Keytar is now a dependency of Zowe CLI and KeytarCredentialManager always loads

Fixed in https://github.com/zowe/imperative/pull/532/commits/314cf59e94bd1bc8a4ffa7a3fd3d48c0828883ce, merging

Issue #533, opened 2021-01-29
[next] Unable to perform profile I/O operations in Zowe SDKs


t1m0thyj:
```
[2021/01/29 10:01:91] [INFO] [Logger.ts:239] An error occurred collecting all configurations from the profile root directory "C:\Users\Timothy\.zowe\profiles". Please supply the configurations on the profile manager constructor parameters OR initialize the profile manager environment. Details: Cannot read property 'exists' of undefined
```

PR #534, opened 2021-01-29
Don't require "config" to be initialized outside of CLI-only code

t1m0thyj:
Fixes "Unable to perform profile I/O operations in Zowe SDKs" (#533)


zFernand0 (approved):
LGTM! 
Clever solution! 😋

gejohnston (approved):
I had to look up the use of ? in this context.  It makes for a nice simple piece of code. I will probably use it in future situations.

PR #538, opened 2021-02-04
Support plugin installation in NPM v7 environment

### LINKED ISSUE #457: Don't automatically install plugin peer dependencies
Imperative plugins are expected to define the CLI and Imperative as peer dependencies (as is done in the [zowe-cli-sample-plugin repo](https://github.com/zowe/zowe-cli-sample-plugin/blob/314550a5326a7e984fc5ffcab945241d33cbda2c/package.json#L43-L46)).

This has worked fine so far. At install time NPM warns if peer dependencies are missing, which is useful for SDK users running `npm install`, and harmless (although perhaps annoying 😛) for CLI users running `zowe plugins install`.

In NPM v7, a [breaking change](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/) has been introduced to install peer dependencies automatically if they're missing. This will likely cause problems, since when `zowe plugins install` runs a 2nd copy of the CLI and Imperative will get installed inside `~/.zowe/plugins`. 

To prevent this from happening, we can turn off the automatic peer dep installation by specifying [`--omit=peer`](https://docs.npmjs.com/cli/v7/commands/npm-install#omitting-dependency-types) on the NPM install command issued by Imperative. Before doing this, need to make sure that passing this flag doesn't break older versions of NPM.

t1m0thyj:
**Changes**
* Disables new behavior of NPM v7 that automatically installs peer dependencies. This is a breaking change in NPM v7 that would impact Imperative plugin install if we don't handle it. Fixes #457.
* Removes code that scrapes command output of `npm install` to find "&lt;packageName&gt;@&lt;packageVersion&gt;". Instead, uses [pacote](https://www.npmjs.com/package/pacote) package (a dependency of NPM itself) to fetch name and version of plugin package.
* Updates plugin-related test suites so they pass in both NPM v6 and NPM v7 environments.

**Notes**
* The `install` and `update` methods for plugins are now asynchronous - they return `Promise<string>` instead of `string`. This change was necessary in order to call the asynchronous `pacote.manifest` method. This should not be a breaking change since AFAICT the affected code is not accessible in any APIs of the published Imperative package.
* The command output returned by the `install` and `update` methods is no longer being parsed. This hopefully makes Imperative more resilient to future NPM changes.
* I've tested that the version of "pacote" used in this PR can handle just about any type of package specifier that you throw at it, to fetch the name and version:
  * Package on NPM registry (e.g., `@zowe/imperative@next`)
  * Local or hosted TGZ (e.g., [imperative-4.10-2.tgz](https://zowe.jfrog.io/artifactory/npm-local-release/%40zowe/imperative/-/%40zowe/imperative-4.10.2.tgz))
  * GitHub repo (e.g., `github:zowe/imperative#next`)
  * Local folder (e.g., `.` for current directory)
* I've tested that all Zowe/Brightside plugins install successfully in an environment with NPM v7 and the CLI built from this PR.


gejohnston (approved):
Looks good.

zFernand0 (approved):
LGTM!
Should we mention keytar was updated for node 15 support? 
No need to worry too much since Node 15 is not LTS 😋

t1m0thyj:
> Should we mention keytar was updated for node 15 support?
> No need to worry too much since Node 15 is not LTS 😋

For Imperative I don't think we need to mention it - Keytar is just a dev dep. For `@zowe/cli@next` we should probably update Keytar and mention this 🙂

PR #540, opened 2021-02-08
Update changelog in pipeline before deploy stage

t1m0thyj:
Version number in changelog should be updated before we deploy a new package 😸 This fix was made in the CLI repo a while ago, but we forgot to port it to Imperative 😿


zFernand0 (approved):
LGTM! 😋

PR #541, opened 2021-02-11
Next refactor config

gejohnston:
Refactor the Config class
- Move api functions out of Config.ts into their own modules in an api subdirectory.
- Adjust the callers of Config functions.
- Merge new functionality from the next branch into the refactored files.
- Fix automated tests.
- Add doc to the functions.


packages/config/src/Config.ts
```diff
@@ -17,46 +17,88 @@ import * as findUp from "find-up";
...
+     * App name used in config filenames (e.g., *my_cli*.config.json)
+     * @internal
+     */
+    public _app: string;
+
+    /**
+     * List to store each of the config layers enumerated in `layers` enum
+     * @internal
+     */
+    public _layers: IConfigLayer[];
```
dkelosky, line 61: right or wrong, most of our member variables in the code base start with `m`.  I know other standards recommend an `_` prefix.  

given that most of the code base is `m`, should this be used here, if at minimum, for aesthetics ?

gejohnston: I am quite willing to make this change.

I do think that we should establish what our standards are before people write code.

I do not feel strongly about this particular naming, since I did not create these variable names.

dkelosky: Perhaps others have a stronger opinion - @zFernand0 & @t1m0thyj?

Ideally our standards would be enforced in lint rules.  Usually most seem to pattern their code with the rest of the code base, but, that's not universally agreed upon either 😄 .

zFernand0: No strong opinions here.
I say let's pick one and open an issue to enforce the rule for internal variables (public or private). 😋

dkelosky: `m` is most common and would require less changes over all than adopting a `_` standard.

t1m0thyj: No strong opinions here either.
Whichever we pick, perhaps we should investigate whether we can set up a linting rule to enforce it? I guess that's the same as what @zFernand0 is suggesting 😋

gejohnston: I agree that we should change this subset of files, instead of changing all other files.

I will undertake the change from _var to mVar now.

t1m0thyj (approved):
LGTM! Thanks for refactoring to remove usage of `_` - I think that was a good decision 🙂

t1m0thyj:
If you want the SonarQube stage to pass in Jenkins pipeline, we can merge #542 first which fixes that issue. Or we can merge this PR first ignoring the failing build for now, I'm fine with either order of merging.

gejohnston:
We can merge #542 or this PR in either order, depending on which is ready first.
I looked over #542 and saw nothing that would create a conflict for this PR.

PR #551, opened 2021-02-25
Fix plugin install from private registry

t1m0thyj:
Fixes https://github.com/zowe/zowe-cli/issues/958

The "pacote" npm SDK doesn't handle authenticating to NPM registries. So this was failing for private registries:
`await pacote.manifest(npmPackage, { registry });`

We could attempt to load credentials from `.npmrc` and provide them to "pacote", but such an approach seems undesirable.

For packages being installed from a registry, this PR bypasses using "pacote" to determine package name and version. We can extract the package name from the `npm install` command, and use it to find where the package was installed in the node_modules folder.

For packages not being installed from a registry (e.g., from a local/remote TGZ, a local directory, or a Git URL), we are unable to extract the package name from the `npm install` command, and can continue using "pacote" to fetch the name and version.


gejohnston (approved):

zFernand0 (approved):
LGTM

awharn (approved):

PR #554, opened 2021-03-01
Interactive Selection Implementation

zFernand0:
Added support for interactive menu selection

### DIFF
--- a/packages/cmd/src/response/CommandResponse.ts
+++ b/packages/cmd/src/response/CommandResponse.ts
@@ -634,9 +634,13 @@ export class CommandResponse implements ICommandResponseApi {
                         throw new ImperativeError({msg: "Menu should contain at least one element"});
                     }
 
+                    /**
+                     * Converts an object[] into a single line menu using the keys as headers
+                     * @param iMenu Menu to be converted to a singleLineMenu (string[]) menu
+                     * @returns single line menu
+                     */
                     const convertInputMenu = (iMenu: object[]): [string[], string] => {
                         const _menu: string[] = outer.format.formatOutput({
-                            // Fields:  owner phase subsystem phase-name job-correlator type url jobid class files-url jobname status retcode
                             fields: opts?.fields,
                             output: iMenu,
                             format: "table",
--- a/packages/cmd/src/response/CommandResponse.ts
+++ b/packages/cmd/src/response/CommandResponse.ts
@@ -290,16 +290,17 @@ export class CommandResponse implements ICommandResponseApi {
                 }
 
                 /**
-                 * Formats and prints the data/output passed. The handler dictates (via the ICommandOutputFormat params)
+                 * Formats the data/output passed. The handler dictates (via the ICommandOutputFormat params)
                  * the default output format. However, if the user has specified any of the response-format options,
                  * those take precedence over the defaults.
                  * @private
                  * @param {ICommandOutputFormat} params - the command format output parameters (see interface for details)
                  * @param {CommandResponse} response - the command response object
                  * @param {Arguments} args - the arguments passed on the command line by the user
+                 * @returns the formatted output to be printed to the console.
                  * @memberof CommandProcessor
                  */
-                private formatOutput(params: ICommandOutputFormat): any {
+                public formatOutput(params: ICommandOutputFormat): any {
 
                     // If a single filter is specified, save the field the data was extracted from
                     const extractedFrom = (params.fields != null && params.fields.length === 1 && typeof params.output !== "string") ?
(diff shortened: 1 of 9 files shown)


packages/cmd/src/CommandProcessor.ts
```diff
@@ -371,7 +371,9 @@ export class CommandProcessor {
             process.chdir(params.arguments.dcd as string)
 
             // reinit config for daemon client directory
-            ImperativeConfig.instance.config = await Config.load(ImperativeConfig.instance.rootCommandName, ImperativeConfig.instance.config.opts);
+            const newOpts = ImperativeConfig.instance.config?.opts || {};
+            if (newOpts.vault == null) newOpts.vault = ImperativeConfig.instance.config?.mVault;
+            ImperativeConfig.instance.config = await Config.load(ImperativeConfig.instance.rootCommandName, newOpts);
```
zFernand0, line 376: This is the issue I was trying to explain in previous standups where the Daemon always prompts for credentials (even after doing `zdevx config secure --global`

dkelosky: thanks - i thought i had fixed something very similar to this before, but apparently not 😄

packages/utilities/src/CliUtils.ts
```diff
@@ -534,6 +535,35 @@ export class CliUtils {
         return answerToReturn;
     }
 
+    public static async interactiveSelection(menu: string[], opts?: IInteractiveOptions): Promise<number> {
+        return new Promise<number>(async (resolve, reject) => {
+            // TODO(Fernando) - Move terminal code out of the API
```
dkelosky, line 540: i think this `TODO` is done 🥳

packages/cmd/src/CommandProcessor.ts
t1m0thyj: I thought so as well... this code seemed to fix it in the past:
https://github.com/zowe/imperative/blob/66f7b630943570faffc3016700e239584b49da6b/packages/config/src/api/ConfigSecure.ts#L32-L34
But at least you got it working 😛

packages/utilities/src/CliUtils.ts
```diff
@@ -534,6 +535,35 @@ export class CliUtils {
...
+                        reject("No input selected!");
+                    }
+                });
+                const singleColumnMenu = require("util").promisify(term.singleColumnMenu);
+                const userInteraction = await singleColumnMenu.call(term, menu, {cancelable: true});
+                term.grabInput(false);
+                resolve(userInteraction.selectedIndex + 1);
+            } catch(err) {
+                term.grabInput(false);
+                reject("Oh oh! Something went wrong!\nDetails:\n" + err.message);
```
t1m0thyj, line 562: "Oh oh" -> "Uh oh"

Also same question as above, should an ImperativeError be generated here?

packages/utilities/src/CliUtils.ts
```diff
@@ -534,6 +535,35 @@ export class CliUtils {
...
+            try {
+                term.grabInput(true);
+                const cursor = await term.getCursorLocation();
+                term.on("key", (name: string) => {
+                    if (name === "CTRL_C" || name === "ESCAPE") {
+                        if (name === "CTRL_C") {
+                            term.moveTo(1, cursor.y + menu.length + (opts?.header != null ? 1 : 0) + 1);
+                        }
+                        term.grabInput(false);
+                        reject("No input selected!");
```
t1m0thyj, line 553: Should an ImperativeError be generated here?

t1m0thyj (commented):
This is an awesome feature! Hope to see it used in other CLI commands soon. 🙂

packages/cmd/src/CommandProcessor.ts
zFernand0: 🤔 Now I'm confused.
Maybe I should remove the hack of creating the `newOpts` object?

packages/cmd/src/response/CommandResponse.ts
```diff
@@ -625,9 +626,44 @@ export class CommandResponse implements ICommandResponseApi {
...
+                public interactiveSelection(menu: string[], opts?: IInteractiveOptions): Promise<number> {
+                    if (outer.mStream) {
+                        return new Promise<number>((resolve) => {
+                            // build interactive header
+                            const daemonHeaders = DaemonUtils.buildHeader({interactive: true});
+
+                            const obj: any = {}
+                            if (opts?.header) obj.header = opts.header;
+                            obj.menu = menu;
+                            outer.writeStream(JSON.stringify(obj));
```
zFernand0, line 644: Should we create `x-zowe-daemon-data` (and maybe `x-zowe-daemon-data-length`) instead of pushing strings before the headers?

dkelosky: this seems like a cleaner way to handle between client and daemon - perhaps this can be done before `next` becomes `latest` ?

dkelosky:
I tried this out and it seemed pretty straight forward to use:

interactive test

A couple ideas:
- can you help with formatting?  in the gif, i was struggling to get things to align, and i think other api users would as well
- it might be cool if the api could accept an array of objects where the key became the headers

zFernand0:
Thanks for trying it Dan!

`can you help with formatting?`
To help with the alignment, I believe I've exposes some of the console.format private functions (like formatOutput and such)
Feel free to check out the `interactive-entries-2 branches 😋 

`it might be cool if the api could accept an array of objects where the key became the headers`
Cool idea. One question for you is how to handle missing keys, like: 
`[ { header1: 123, header2: 456 }, { header1: 234, header3: 789} ]`

dkelosky:
> To help with the alignment, I believe I've exposes some of the console.format private functions (like formatOutput and such)
Feel free to check out the `interactive-entries-2 branches 😋

😢 ok I'll go searching for it 

> Cool idea. One question for you is how to handle missing keys, like:
[ { header1: 123, header2: 456 }, { header1: 234, header3: 789} ]

ImperativeError!

packages/cmd/src/CommandProcessor.ts
dkelosky: i think this change is needed - im pretty confused, but i started experiencing the problem @zFernand0 had.  this code change fixed the problem (the continuous need to respond to credential prompting)

zFernand0:
> > To help with the alignment, I believe I've exposes some of the console.format private functions (like formatOutput and such)
> > Feel free to check out the `interactive-entries-2 branches 😋
> 
> 😢 ok I'll go searching for it
> 
> > Cool idea. One question for you is how to handle missing keys, like:
> > [ { header1: 123, header2: 456 }, { header1: 234, header3: 789} ]
> 
> ImperativeError!

Hey @dkelosky,
I removed the interactive-entries-2 branches. however `object[]` is now supported 😋 
Here is an example of how to use the API both ways 😋 
https://github.com/zowe/zowe-cli/blob/225ef68779869483757ffa69a56683c4ee0d275d/packages/cli/src/zosjobs/view/interactive-jobs/InteractiveJobs.handler.ts#L36-L57

dkelosky:
Very cool feature!  Thanks for sharing an example - it looks clean.

Also, I never knew the `??` existed before!

packages/cmd/src/response/CommandResponse.ts
zFernand0: This should be resolved by https://github.com/zowe/imperative/pull/681

zFernand0:
I believe that there are too many changes required for this PR to become ready for review.
It might be best to close this and create a new one when/if there's any interest.
Also, this could be moved to a plug-in that introduces interaction and animations to the terminal

PR #561, opened 2021-03-09
Update Next with 4.11.1-4.12.0

awharn:
Also changes 4.12.0 to have decoding enabled by default


awharn:
History is borked, creating new branch

PR #567, opened 2021-03-12
Prof info restore env

gejohnston:
Restores the skeleton to implement overriding values with environment variables.

This is a small change with low impact. I will merge this into profInfo-for-next at close of business Fri Mar 12, unless objections are raised.

### DIFF
--- a/packages/config/src/ProfileInfo.ts
+++ b/packages/config/src/ProfileInfo.ts
@@ -216,7 +216,7 @@ export class ProfileInfo {
                 defaultProfile.profName = loadedProfile.name;
                 defaultProfile.profLoc = {
                     locType: ProfLocType.OLD_PROFILE,
-                    osLoc: nodeJsPath.resolve(profRootDir + "/" + profileType + "/" +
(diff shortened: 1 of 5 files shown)


t1m0thyj (approved):

packages/config/src/ProfileInfo.ts
```diff
@@ -17,6 +17,7 @@ import { IProfAttrs } from "./doc/IProfAttrs";
 import { IProfArgAttrs } from "./doc/IProfArgAttrs";
 import { IProfLoc, ProfLocType } from "./doc/IProfLoc";
 import { IProfMergedArg } from "./doc/IProfMergedArg";
+import { IProfOpts } from "./doc/IProfOpts";
 
 // for team config functions
 import { Config } from "./config";
```
awharn, line 23: Failing to build because of file pathing error. `./config` should be `./Config`

gejohnston: Thanks for finding this. Something else always seemed more urgent than looking into pipeline errors, since it built and tested ok on my Windows laptop.

Issue #461, opened 2020-11-11
Random codecov diffs


mheuzey:
When opening #458, I noticed that the codecov report initially showed some lines losing coverage in a file that I had not touched (CommandResponse.ts). The diff disappeared and reappeared as I made minor adjustments to the PR (no code changes). Looking into this topic of random codecov diffs, I found a thread which suggested that the tests might be non-deterministic. So I ran the unit test for that file on my local machine a few times, and it does look like the coverage changes between runs.

![image](https://user-images.githubusercontent.com/32349430/98859690-8b5ba000-2430-11eb-85a3-6a6858bbab90.png)

I'm not an expert on tests, but I figured I would bring this up so someone with more knowledge could verify whether something in the test suite could be causing random coverage changes. I don't know how widespread this problem is, but looking over the last few PRs, I see that #456 and #435 both have the same coverage diff as mine, and both PRs made small changes that shouldn't have affected coverage.

JTonda:
We believe that updates to the codecov tools have eliminated this problem therefore we are closing this issue.

PR #572, opened 2021-03-23
Added headers[] option to TextUtils.getTable()

### LINKED ISSUE #369: TextUtils.getTable: accept headers[] as parameter
I would like to pass an array of headers to TextUtils.getTable, in order to guarantee column order. Per the [ECMAScript spec](https://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys), properties are enumerated in the following order:

1. Integer indices (if applicable), in ascending order.
2. Other string keys (if applicable), in property creation order.
3. Symbol keys (if applicable), in property creation order.

This means that in my table, if some of my column headers are simple numbers, they will get hoisted to the left of the table, since TextUtils.getTable builds the headers by calling Object.keys on the array of key-value pairs.

Another use case would be if I have built a full table of values, but don't always want to display the whole thing. By passing an array of headers, I can filter out which columns are displayed, if necessary.

mheuzey:
Added a headers[] parameter to TextUtils.getTable(). This gives the user more control over which headers to display, and in what order. Closes #369.

As a side note, TestUtils.getTable now has 7 parameters, 5 of which are optional. It might be a good idea to move to an ITableOptions object as a parameter. I held off on this for now, as I believe this would be a breaking change, but we might want to consider this for the next major version.

Signed-off-by: mheuzey <michael.heuzey@gmail.com>

### DIFF
--- a/packages/utilities/src/TextUtils.ts
+++ b/packages/utilities/src/TextUtils.ts
@@ -156,18 +156,12 @@ export class TextUtils {
      */
     public static getTable(objects: any[], primaryHighlightColor: string,
                            maxColumnWidth?: number, includeHeader: boolean = true, includeBorders: boolean = false,
-                           hardWrap: boolean = false, headers: string[] = []): string {
+                           hardWrap: boolean = false, headers?: string[]): string {
         const Table = require("cli-table3");
-        // for every property of every object in the array,
-        // make a header for the table
-        if (headers.length === 0) {
-            for (const obj of objects) {
-                for (const key of Object.keys(obj)) {
-                    if (headers.indexOf(key) === -1) {
-                        headers.push(key);
-                    }
-                }
-            }
+
+        // if the user did not specify which headers to use, build them from the object array
+        if (!headers) {
+            headers = this.buildHeaders(objects);
         }
         if (isNullOrUndefined(maxColumnWidth)) {
             maxColumnWidth = this.getRecommendedWidth() / headers.length;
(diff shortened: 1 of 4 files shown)


packages/utilities/src/TextUtils.ts
```diff
@@ -150,19 +150,22 @@ export class TextUtils {
...
      * @param includeBorders -  should the table have borders between the cells?
      * @param hardWrap - hard wrap the text within the width of the table cells (defaults to false)
+     * @param headers - specify which headers in which order to display. if omitted, loops through the rows
+     *        and adds object properties as headers in their enumeration order
      * @returns {string} the rendered table
      */
     public static getTable(objects: any[], primaryHighlightColor: string,
                            maxColumnWidth?: number, includeHeader: boolean = true, includeBorders: boolean = false,
-                           hardWrap: boolean = false): string {
+                           hardWrap: boolean = false, headers: string[] = []): string {
```
zFernand0, line 159: it'd be nice to have an interface for this function so that people don't have to `getTable(objects, "white", null, true, false, false, ["header1","header2"])`, and instead call it like 
```
getTable({
  objects,
  primaryHighlightColor: "white",
  includeHeader: true,
  headers: ["header1", "header2"]
})
```

zFernand0 (approved):
LGTM!
Should we care about the Cognitive Complexity issues in Sonar Cloud?

mheuzey:
It looks like 10 out of the 16 points are just from the one spot that builds the headers from the object properties, because of nesting. If you counted that spot as just 1 point, you would be down to 7 out of the allowed 15. The alternative way to placate SonarCloud would be to pull that code out into a separate buildHeaders() function, so that it gets its own allotment of points.

mheuzey:
The coverage for the file I updated has increased; the decrease in overall coverage is due to the same file I mentioned in #461 about fluctuating code coverage. I would say it's safe to ignore that one.

awharn (approved):

PR #573, opened 2021-03-24
Update mProgressApi and endBar to prevent output duplication

awharn:
When calling `mProgressApi`'s `endBar()`, the contents of the `stdout` and `stderr` buffers are printed to the console. Before now, calls to `endBar()` would output duplicate data on the console because the buffers are never cleared when their contents are printed to the console. This PR rectifies this behavior by only printing a subset of the buffer that was written after the progress bar started to console.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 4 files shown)


gejohnston (approved):
Looks good.

t1m0thyj (approved):
LGTM!

zFernand0 (approved):
🚀

PR #575, opened 2021-03-24
ProfInfoErr class into profInfo-for-next

gejohnston:
This PR adds the ProfInfoErr class.

The way that I envision ProfInfoErr being used is:
- All errors from the ProfileInfo class will throw a ProfInfoErr.
- We will always set the errorCode property for any error thrown in ProfileInfo.
- Consumers will test for "instanceof ProfInfoErr", and then test our errorCode.

The following two items work, but we could change them based on your review opinions.
1. The errorCode property is a string. I defined a constant for each known error, and chose to use a mnemonic phase as the value.

           MISSING_REQ_PROP: string = "MissingProp"

    We could choose smaller strings, maybe numbers? "1", "2", "3", etc. I know that our mainframe division likes IBM-style error codes (like IEB1234), but I hate them. I do not think that they are necessary in this context, and I like the mnemonic.

2. I added a property named itemsInError, which is an array of strings. It is not currently used anywhere. I added it to address the need that Andrew identified for listing multiple bad items with one thrown error. If that property does not work for the identified problem, it can be changed.

### DIFF
--- a/packages/config/src/ProfInfoErr.ts
+++ b/packages/config/src/ProfInfoErr.ts
@@ -79,7 +79,11 @@ export class ProfInfoErr extends ImperativeError {
     ) {
         super(impErrDetails, profErrParms);
         this.name = "ProfInfoErr";
-        this.mItemsInError = profErrParms?.itemsInError;
+
+        // make a shallow copy of the itemsInError array
+        if (profErrParms) {
+            this.mItemsInError = [...profErrParms.itemsInError];
+        }
     }
 
     /**
--- a/packages/config/src/ProfInfoErr.ts
+++ b/packages/config/src/ProfInfoErr.ts
@@ -10,7 +10,9 @@
 */
 
 // for imperative operations
-import { ImperativeError } from "../../error";
+import { IProfInfoErr } from "./doc/IProfInfoErr";
+import { ImperativeError, IImperativeError } from "../../error";
+import { IImperativeErrorParms } from "../../error/src/doc/IImperativeErrorParms";
 
 /**
  * This class is the error exception mechanism for the ProfileInfo API.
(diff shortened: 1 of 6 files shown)


packages/config/src/ProfInfoErr.ts
```diff
@@ -0,0 +1,95 @@
...
+     * @param impErrDetails
+     *        ImperativeError details and text (stack, messages, etc.)
+     *
+     * @param impErrParms
+     *        ImperativeError control parameters to indicate logging of node-report and more
+     */
+    constructor(
+        impErrDetails: IImperativeError,
+        profErrProps?: IProfInfoErr,
+        impErrParms?: IImperativeErrorParms
```
t1m0thyj, line 80: Rather than passing two key-value objects into the constructor, would it make sense for `IProfInfoErr` to extend the interface `IImperativeErrorParms` and pass only a single object?

t1m0thyj (approved):

packages/config/src/ProfInfoErr.ts
zFernand0: With no conflicting properties, I think it may make things a bit easier for the programmer 😋 

Not sure what the preferred word is between `props` and `parms` though

gejohnston: I thought about doing this. The reason I did not, was that you needed to specify the ImperativeError details in a second nested level of the structure. If you do not specify either of the second two arguments (which is probably most cases), you would supply the same structure to ProfInfoErr, as you would to ImperativeError.

That being said, since others feel that the combined structure is a better interface, I am all for it. I will make the changes and push an update to the PR.

gejohnston: Perhaps I misinterpreted the original comment. Were you suggesting keeping what ImperativeError calls "details" as the first argument to ProInfoErr, and only combining the second two arguments to ProfiInfoErr. That is, combining stuff specific to ProfileInfo and what ImperativeError calls its parms argument?

t1m0thyj: I was thinking to replace:
`export interface IProfInfoErr {`
with:
`export interface IProfInfoErr extends IImperativeErrorParms {`

As long as the properties added to `IProfInfoErr` do not conflict with the `IImperativeErrorParms` properties, there should be no additional nesting of properties.

packages/config/src/ProfileInfo.ts
```diff
@@ -464,10 +483,16 @@ export class ProfileInfo {
             }
 
             if (missingRequired.length > 0) {
-                throw new ImperativeError({ msg: "Missing required properties: " + missingRequired.join(", ") });
+                throw new ProfInfoErr({
+                    errorCode: ProfInfoErr.MISSING_REQ_PROP,
+                    msg: "Missing required properties: " + missingRequired.join(", ")
+                });
```
t1m0thyj, line 489: Perhaps this validation is unnecessary and we should leave it up to SDK users to throw an exception if options they deem to be required are missing?

This question isn't directly related to the PR but I wanted to raise it for discussion. If we think this validation is unnecessary, removing it doesn't need to be addressed in this PR - I can create an issue for it 🙂

gejohnston: Since the the array of missingArgs contains arguments with default values, perhaps we should check for both missingArgs.length > 0 and at least one missingArg has no value?

If default values were added to knownArgs, the consuming app could make an easy test for missingArgs.length > 0 and would not need our exception.

If we can concisely describe what we could give to the consumer, maybe we should ask the consumers what they would prefer in our design doc Github issue?

t1m0thyj: > If we can concisely describe what we could give to the consumer, maybe we should ask the consumers what they would prefer in our design doc Github issue?

Agree this might be a good idea. For context, the current implementation is that we check if the number of missingArgs in the profile schema's `required` property is greater than 0.

awharn (approved):

Issue #568, opened 2021-03-12
[next] Investigate removing the ConfigProfiles.load method


t1m0thyj:
I created the `ConfigProfiles.load` method when the "auth login" command was implemented to use team config. We needed a way to load profile info, which didn't exist yet.

After the profile info API is created (#556), it should do everything that `ConfigProfiles.load` does, and much more. We may want to remove the method if it is no longer necessary.

t1m0thyj:
Refactoring the "auth login" handler to use the new ProfileInfo API could be an opportunity to test the ProfileInfo API and better define its behavior for handling tokens.

Some relevant comments from #579:
- Can we accurately describe how a base profile's user, password, or token is merged with a service profile's user and password?
- Which arguments are returned by the ProfiInfo API?
- I am less interested in how it operates, and more interested in what the consumer should expect, for example:
    - Is a token NOT returned if a user & password are found in a service profile?
    - Do we return a token if the user & password for the service profile is inherited from a higher-level node in the JSON document?
    - Will we return all such arguments and expect the consumer to sort out token vs user & password?
- Do we need automated tests to confirm the expected behavior?

PR #579, opened 2021-03-25
[next] Secure credential handling for ProfileInfo API

### LINKED ISSUE #577: [next] Don't reinitialize Imperative logger in ProfileInfo
Zowe Explorer and internal VS Code extensions already call the Imperative method `Logger.initLogger` and configure it using log4jsconfig. When ProfileInfo initializes the logger again, it breaks the already initialized logger.

t1m0thyj:
* Added `ProfileCredentials` class to handle initialization of CredentialManager
* Added `requireKeytar` property to IProfOpts so that VS Code extensions can use a custom Keytar module
* Added `secure` property to IProfArgAttrs for secure arguments. When set to true, `argValue` is left undefined until `ProfileInfo.loadSecureArg` is called, so that loading of sensitive data is delayed. This may resolve https://github.com/zowe/zowe-cli/issues/957.
* Don't create new `ImperativeLogger` in ProfileInfo constructor if one already exists (fixes #577, thanks to @gejohnston for contributing this fix).


packages/config/src/doc/IProfOpts.ts
```diff
@@ -9,15 +9,35 @@
...
 /**
  * Options that will affect the behavior of the ProfileInfo class.
  * They are supplied on the ProfileInfo constructor.
  */
 export interface IProfOpts {
+    /**
+     * Specifies an already initialized Imperative logger that the ProfileInfo
+     * class should use. If omitted, a new logger will be instantiated.
+     */
+    imperativeLogger?: Logger;
```
gejohnston, line 23: I am certain that I added the following code to ProfileInfo.initImpUtils( ) in an attempt to avoid re-initializing logging without having to make our caller pass a logger on our constructor. Now however, I cannot find this code in any commit history.  Somehow I must have lost the code before making a commit. I know this code will initialize logging if it has not been initialized. Perhaps you can test if this code works properly in a test environment where a caller has already initialized logging? The "if" test is the only thing not currently committed to ProfileInfo-for-next
```
        // initialize logging
        if (LoggerManager.instance.isLoggerInit === false) {
            const loggingConfig = LoggingConfigurer.configureLogger(
                ImperativeConfig.instance.cliHome, ImperativeConfig.instance.loadedConfig
            );
            Logger.initLogger(loggingConfig);
        }
        this.mImpLogger = Logger.getImperativeLogger();
    }
```

gejohnston: To use the comment above a second required change is to import a module:
import { LoggerManager } from "../../logger/src/LoggerManager";

packages/config/src/ProfileInfo.ts
```diff
@@ -361,7 +364,8 @@ export class ProfileInfo {
     public mergeArgsForProfile(profile: IProfAttrs): IProfMergedArg {
         const mergedArgs: IProfMergedArg = {
             knownArgs: [],
-            missingArgs: []
+            missingArgs: [],
+            secureArgs: []
```
gejohnston, line 368: In the mergeArgsForXXX functions, knownArgs and missingArgs are mutually exclusive. Every argument belongs into one of those categories. Each secureArg will fall into one of the existing categories of known or missing.

- Do you think that the consumer will want a separate list of secure arguments?   

- Will the new "isSecure" property not be sufficient for the consumer to process an argument correctly from the known or missing args?

packages/config/src/ProfileInfo.ts
```diff
@@ -560,6 +576,31 @@ export class ProfileInfo {
         return this.mUsingTeamConfig;
     }
 
+    /**
+     * Load value of secure argument from the vault.
+     * @param arg Secure argument object
+     */
+    public loadSecureArg(arg: IProfArgAttrs): any {
```
gejohnston, line 583: During final cleanup of this PR, I ask that you add an example of using this public function to the pseudocode within the block comments for the ProfileInfo class.

gejohnston:
This is not a comment on the implementation of this PR, but my topic is closely related so it seems appropriate to hold a conversation here, even if the result is a separate issue or story.

- Can we accurately describe how a base profile's user, password, or token is merged with a service profile's user and password?
- Which arguments are returned by the ProfiInfo API?
- I am less interested in how it operates, and more interested in what the consumer should expect, for example:
    - Is a token NOT returned if a user & password are found in a service profile?
    - Do we return a token if the user & password for the service profile is inherited from a higher-level node in the JSON document?
    - Will we return all such arguments and expect the consumer to sort out token vs user & password?
- Do we need automated tests to confirm the expected behavior?

packages/config/src/doc/IProfOpts.ts
t1m0thyj: Thanks, I'll implement this and get rid of the unnecessary `imperativeLogger` property.

packages/config/src/ProfileInfo.ts
t1m0thyj: Both good points - `secureArgs` is likely unnecessary and can be removed.

> Do you think that the consumer will want a separate list of secure arguments?

I don't think it's a requirement, and keeping the properties of `IProfMergedArg` mutually exclusive is probably a good idea.

> Will the new "isSecure" property not be sufficient for the consumer to process an argument correctly from the known or missing args?

When testing these APIs in a VSCE, it seems simpler to handle secure args using only the `isSecure` property instead of having them in a separate array.

t1m0thyj: Perhaps I should rename `isSecure` to `secure`? Wondering if the word `is` in the property name is unnecessary.

gejohnston: I do not think there is any firm rule, but my observation is that "isXxx" tends to be the name of a function to determine some condition, but xxx is more often used as the name of a property.

t1m0thyj:
> This is not a comment on the implementation of this PR, but my topic is closely related so it seems appropriate to hold a conversation here, even if the result is a separate issue or story.
> 
> * Can we accurately describe how a base profile's user, password, or token is merged with a service profile's user and password?

It would be possible to determine if an argument came from a base profile by looking at its `jsonLoc`. If this is too inconvenient, perhaps we could add a property to `IProfArgAttrs` specifying the profile type from which an argument was loaded?

> * Which arguments are returned by the ProfiInfo API?

Fernando's PR raised some questions around this:
1. If there is no schema present, how will `mergeArgsFromProfile` know to "only process properties that were in a profile"? Without a schema, it doesn't know that "user" is a profile property and "response-format-json" is not.
2. When there is a schema present, should `mergeArgsForProfile` throw an error if non-profile properties like "response-format-json" get loaded into `knownArgs`?

> * I am less interested in how it operates, and more interested in what the consumer should expect, for example:
>   
>   * Is a token NOT returned if a user & password are found in a service profile?
>   * Do we return a token if the user & password for the service profile is inherited from a higher-level node in the JSON document?
>   * Will we return all such arguments and expect the consumer to sort out token vs user & password?

I believe we return all such arguments and leave it up to the consumer (or the `ConnectionPropsForSessCfg` API) to prioritize user/password over token when both are present.

> * Do we need automated tests to confirm the expected behavior?

The behavior of merging args in certain cases seems unclear at this point since we have questions around it, so I think adding tests would help to better define the behavior.

t1m0thyj:
I think the outstanding questions from this PR have been captured in https://github.com/zowe/imperative/issues/568#issuecomment-811366713 and #583.

Trying to figure out now why so many integration tests in this PR have started failing.

gejohnston (approved):
This looks good.
I think that all remaining concerns about credential handling are bigger than this story, and they should be handled in some future, prioritized, scheduled story.

zFernand0 (approved):
LGTM!

PR #580, opened 2021-03-26
Read environment Variables for ProfInfo APIs

zFernand0:
- Allow reading and overriding of missing arguments

### DIFF
--- a/packages/config/src/ProfileInfo.ts
+++ b/packages/config/src/ProfileInfo.ts
@@ -870,8 +870,7 @@ export class ProfileInfo {
         const envStart = envPrefix + "_OPT_";
         for (const key in process.env) {
             if (key.startsWith(envStart)) {
-                const value: string = process.env[key];
-                let argValue: any = value;
+                let argValue: any = process.env[key];
                 let dataType: any = typeof argValue;
                 let argName: string = CliUtils.getOptionFormat(key.substring(envStart.length).replace(/_/g, "-").toLowerCase()).camelCase;
 
@@ -882,25 +881,20 @@ export class ProfileInfo {
                         }
                     }
                 } else {
-                    if (value.toUpperCase() === "TRUE" || value.toUpperCase() === "FALSE") {
+                    if (argValue.toUpperCase() === "TRUE" || argValue.toUpperCase() === "FALSE") {
                         dataType = "boolean";
-                    } else if (!isNaN(+(value))) {
+                    } else if (!isNaN(+(argValue))) {
                         dataType = "number";
-                    } else if (/(^"{)(}"$)/g.exec(value)) {
-                        dataType = "object";
-                    } else if (/(^"\[)(\]"$)/g.exec(value)) {
-                        dataType = "array";
                     }
+                    // TODO: Look for option definition for argName to check if it's an array
                 }
 
                 if (dataType === "boolean") {
-                    argValue = value.toUpperCase() === "TRUE";
+                    argValue = argValue.toUpperCase() === "TRUE";
                 } else if (dataType === "number") {
-                    argValue = +(value);
+                    argValue = +(argValue);
                 } else if (dataType === "array") {
-
-                } else if (dataType === "object") {
-
+                    argValue = CliUtils.extractArrayFromEnvValue(argValue);
                 }
 
                 const tempArg: IProfArgAttrs = {
@@ -918,11 +912,5 @@ export class ProfileInfo {
                 }
             }
         }
-
-        if (this.mOverrideWithEnv === false) {
-            return;
-        }
-
-        // todo: get values from environment variables
     }
 }
--- a/packages/config/src/ProfileInfo.ts
+++ b/packages/config/src/ProfileInfo.ts
@@ -921,7 +921,7 @@ export class ProfileInfo {
                     if (knownArgsIndex < 0) {
                         mergedArgs.knownArgs.push(tempArg);
                     }
-                    if (missingArgsIndex >= 0) mergedArgs.missingArgs = mergedArgs.missingArgs.splice(missingArgsIndex, 1);
+                    if (missingArgsIndex >= 0) mergedArgs.missingArgs.splice(missingArgsIndex, 1);
                 }
             }
         }
--- a/packages/config/src/ProfileInfo.ts
+++ b/packages/config/src/ProfileInfo.ts
@@ -464,7 +464,7 @@ export class ProfileInfo {
             // overwrite with any values found in environment
             this.overrideWithEnv(mergedArgs, profSchema);
 
-            for (const tempArg of mergedArgs.missingArgs) {
+            for (const tempArg of mergedArgs.missingArgs || []) {
                 // Check if missing property is required
                 if (profSchema.required?.includes(tempArg.argName)) {
                     missingRequired.push(tempArg.argName);
@@ -920,10 +920,8 @@ export class ProfileInfo {
                 if (argNameFound || missingArgsIndex >= 0) {
                     if (knownArgsIndex < 0) {
                         mergedArgs.knownArgs.push(tempArg);
-                    } else {
-                        mergedArgs.knownArgs[knownArgsIndex] = tempArg;
                     }
-                    if (missingArgsIndex >= 0) delete mergedArgs.missingArgs[missingArgsIndex];
+                    if (missingArgsIndex >= 0) mergedArgs.missingArgs = mergedArgs.missingArgs.splice(missingArgsIndex, 1);
                 }
             }
         }
--- a/packages/config/src/ProfileInfo.ts
+++ b/packages/config/src/ProfileInfo.ts
@@ -917,7 +917,7 @@ export class ProfileInfo {
 
                 const missingArgsIndex = mergedArgs.missingArgs.findIndex((arg) => arg.argName === argName);
                 const knownArgsIndex = mergedArgs.knownArgs.findIndex((arg) => arg.argName === argName);
-                if (missingArgsIndex < 0 && argNameFound) {
+                if (missingArgsIndex < 0 || argNameFound) {
                     if (knownArgsIndex < 0) {
                         mergedArgs.knownArgs.push(tempArg);
                     } else {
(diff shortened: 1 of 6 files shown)


packages/config/src/ProfileInfo.ts
```diff
@@ -863,11 +865,57 @@ export class ProfileInfo {
...
+                    argName,
+                    argValue,
+                    dataType,
+                    argLoc: { locType: ProfLocType.ENV }
+                };
+
+                const argIndex = mergedArgs.knownArgs.findIndex((arg) => arg.argName === argName)
+                if (argIndex < 0) {
+                    mergedArgs.knownArgs.push(tempArg);
+                } else if (this.mOverrideWithEnv) {
```
gejohnston, line 915: Is there a reason for placing the test of mOverrideWithEnv at the end of this function? If we tested mOverrideWithEnv for FALSE at the beginning of the function, we could immediately return and not perform any of the environment operations.

What have I overlooked?

gejohnston (changes requested):
See if my suggested change makes sense.

packages/config/src/ProfileInfo.ts
zFernand0: I guess I made the assumption (probably a wrong one :yum:) that we may want to still load environment variables into `knownArgs` in case they were not specified by the user, or don't exist in a profile.

gejohnston: The mOverrideWithEnv variable was created because at least some of the VSCode extension teams said that they never want environment variables to override the values from the profiles.  Such teams will specify such a choice on the ProfileInfo constructor, and we will not use the environment to override any argument value for that instance of the ProfileInfo class.

gejohnston: My interpretation is that those GUI teams don't want the environment to supply a value even for an argument that is not specified anywhere. They feel that environment variables will be too confusing for  their VSCode users.

zFernand0: Thanks Gene. I think I've address these comments. 😋

gejohnston (approved):
Looks good to me.

packages/config/src/ProfileInfo.ts
```diff
@@ -863,11 +865,59 @@ export class ProfileInfo {
...
+                    for (const [propName, propObj] of Object.entries(profSchema.properties || {})) {
+                        if (argName === propName) {
+                            dataType = this.argDataType(propObj.type);
+                            argNameFound = true;
+                        }
+                    }
+                }
 
-        // todo: get values from environment variables
+                if (profSchema == null || !argNameFound) {
```
t1m0thyj, line 890: Is it desired behavior that I can set a property `ZOWE_OPT_SOME_ARG_NAME_THAT_DOES_NOT_EXIST`, and the value will get read as a known arg?

Just want to make sure we don't get bitten by a scenario like this:
```shell
export ZOWE_OPT_RECURSIVE=true
zowe files delete uss <dirname> -f
```

Oh crap, I deleted a whole directory recursively!

I know this is an odd scenario - why would someone set this environment variable? But it could be unexpected behavior if users expect only properties contained in the profile schema (host, user, password, etc) to be loaded from environment variables.

t1m0thyj: This may also be a broader discussion than just environment variables - I suppose users could also set `'recursive': true` in a config JSON file, which would result in a similar scenario.

gejohnston: I believe that our agreement with consumers of this API was that we would only process properties that were in a profile (or missing properties that were defined for that profile type). I think that we should extend that to environment variables.

I think would could achieve that limitation if some of this logic were grouped under the following test, which occurs earlier in function:
```
       if (profSchema != null) {
```
Inside there, we could test that the environment variable matches a missing arg.
I think that we would still want to process environment variables for the knownArgs even if we did not have a schema.

t1m0thyj: > I believe that our agreement with consumers of this API was that we would only process properties that were in a profile (or missing properties that were defined for that profile type). I think that we should extend that to environment variables.

I agree with this - but it's unclear to me how we will "only process properties that were in a profile" when loading arg values from a config JSON file? If there is no schema present, a user could define any args they wish in a team config profile (`response-format-json`, `help-web`, etc).

Another question - if there is a schema present, should `mergeArgsForProfile` throw an error when non-profile args (like `rfj`) are defined in a config JSON file? This validation isn't happening currently.

> I think that we would still want to process environment variables for the knownArgs even if we did not have a schema.

I think environment variables only need to be processed for missingArgs. The way knownArgs is currently implemented, all the args it contains have values already defined, so looking at environment variables (at the bottom of the order of precedence list) could be skipped.

zFernand0: I've made some changes to the implementation...
Please see comment below

zFernand0:
In an attempt to address all comments on the PR, I've made some changes to the current implementation.
As long as `mOverrideWithEnv === true`:
- properties found in the schema will be overwritten (in `knownArgs`)
- properties found in the `missingArgs` will be added or updated/overwritten with ENV values regardless of whether they are found in the schema.

packages/config/src/ProfileInfo.ts
gejohnston: I think the placement of default values into either missing or known args muddies the conversation. Currently default values are in missing, but our VSCE users could ask to change that. I think that these statements follow the order-of-precedence used by the CLI, and mostly conforms to Timothy's guidelines. 

1. A default value can be overridden by an env var. We only get default values if we can find a schema.
2. A missing value (which is not default) can also be overridden by an env var.  We only get missing values by finding a schema and finding properties that are required.
3. A known value should NOT be overridden by an env var. Just like the CLI, if the user places a value in a config file, the config file takes precedence over an env var. Sorry, I forgot this particular rule in earlier conversations.

zFernand0: hopefully this helps address the 3rd rule 😋 
https://github.com/zowe/imperative/blob/c352d576bc372b551358ca6aee1861674e32b32d/packages/config/__tests__/ProfileInfo.test.ts#L329-L334

packages/config/src/ProfileInfo.ts
```diff
@@ -39,6 +39,7 @@ import { CliUtils, ImperativeConfig } from "../../utilities";
 import { ImperativeError } from "../../error";
 import { ImperativeExpect } from "../../expect";
 import { Logger } from "../../logger";
+import { IProfArgAttrs } from "./doc/IProfArgAttrs";
 
 /**
  * This class provides functions to retrieve profile-related information.
```
dkelosky, line 45: Is the new api all wrapped up in `ProfileInfo` class ? I don't want to be picky on names, but two initial reactions:
- `Profile` commands are being deprecated, should new APIs be associated with a Profile name?
- `Info` seems like it provides "profile info" but the description says it provides functions to retrieve profile-related information. Is this the final name?

packages/config/src/ProfileInfo.ts
zFernand0: each property tries to address a different scenario.
- `host`: property from team config -> ENV var will not update it even if `mOverrideWithEnv === true`
- `RFH `: property from team config -> not found in schema, nor missing arguments (Not updated)
- `port`: property defined in schema with default value and not specified in team config
- `list`: property defined in schema but not specified in team config

packages/config/src/ProfileInfo.ts
```diff
@@ -920,10 +920,8 @@ export class ProfileInfo {
                 if (argNameFound || missingArgsIndex >= 0) {
                     if (knownArgsIndex < 0) {
                         mergedArgs.knownArgs.push(tempArg);
-                    } else {
-                        mergedArgs.knownArgs[knownArgsIndex] = tempArg;
                     }
-                    if (missingArgsIndex >= 0) delete mergedArgs.missingArgs[missingArgsIndex];
+                    if (missingArgsIndex >= 0) mergedArgs.missingArgs = mergedArgs.missingArgs.splice(missingArgsIndex, 1);
```
t1m0thyj, line 924: Should the left side of equal sign be removed?
`mergedArgs.missingArgs = mergedArgs.missingArgs.splice(missingArgsIndex, 1);` -> `mergedArgs.missingArgs.splice(missingArgsIndex, 1);`

I believe `Array.splice` returns the deleted item(s), not the remaining item(s) in the array.

packages/config/src/ProfileInfo.ts
t1m0thyj: Thanks for addressing and sorry for complicating this discussion by getting the behavior of `mergeArgsForProfile` involved 😢 I'll move that part of the discussion to #579, so this conversation can be resolved 🙂

packages/config/src/ProfileInfo.ts
zFernand0: `/facepalm` 🤦

zFernand0: I believe this is fixed, and enhanced the test so that this behavior is properly captured.

packages/config/src/ProfileInfo.ts
gejohnston: I was responsible for choosing the name ProfileInfo. Discussions with stakeholders made it clear that they wanted to limit retrieval of arguments to those arguments that were contained in profiles or defined within profile definitions. Even though questions about writing back continue to occur, our mandate was to read the information and merge any overriding values. Thus the name profile-information --> ProfileInfo. I just did not find a better name. I do not object to a better name. If we want to do that, I suggest that we approve and merge this PR, and open an issue for the name change to be completed before we deliver an @next version that can be used by VSCE validators.

t1m0thyj (approved):
LGTM 🙂

packages/config/src/ProfileInfo.ts
zFernand0: no worries! Thanks for the discussion! : )

packages/config/src/ProfileInfo.ts
zFernand0: please see https://github.com/zowe/imperative/issues/582 :)

PR #589, opened 2021-04-09
Change Override Credential Manager Service Name to Use DafaultCredentialManager Name

dkelosky:
Fix https://github.com/zowe/zowe-cli/issues/984

This adds `Zowe` to the list of service names to read from for SCS plugin (and any others that likely do not exist).  It allows the old SCS plugin to read credentials securely stored via DefaultCredentialManager.  

I think this is technically a breaking change since `@zowe/cli` will no longer be added to [`allServices`](https://github.com/zowe/zowe-cli-scs-plugin/blob/master/src/credentials/KeytarCredentialManager.ts#L50). 

Signed-off-by: Dan Kelosky <dkelosky@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Allow SCS to load new securely stored credentials. [#984](https://github.com/zowe/zowe-cli/issues/984)
+
 ## `5.0.0-next.202104071400`
 
 - Enhancement: Add the ProfileInfo API to provide the following functionality:
--- a/packages/cmd/src/CommandProcessor.ts
+++ b/packages/cmd/src/CommandProcessor.ts
@@ -371,9 +371,13 @@ export class CommandProcessor {
             process.chdir(params.arguments.dcd as string)
 
             // reinit config for daemon client directory
-            ImperativeConfig.instance.config = await Config.load(ImperativeConfig.instance.rootCommandName, ImperativeConfig.instance.config.opts);
+            const newOpts = ImperativeConfig.instance.config?.opts || {};
+
+            if (newOpts.vault == null) newOpts.vault = ImperativeConfig.instance.config?.mVault;
+            ImperativeConfig.instance.config = await Config.load(ImperativeConfig.instance.rootCommandName, newOpts);
             this.mConfig = ImperativeConfig.instance.config;
         }
+
         const prepareResponse = this.constructResponseObject(params);
         prepareResponse.succeeded();
 
--- a/packages/imperative/src/OverridesLoader.ts
+++ b/packages/imperative/src/OverridesLoader.ts
@@ -80,6 +80,8 @@ export class OverridesLoader {
   ): Promise<void> {
     const overrides: IImperativeOverrides = config.overrides;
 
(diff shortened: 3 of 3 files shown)


gejohnston:
I thought we were going to recommend that users who try the team-config should uninstall the SCS plugin. I think our policy is to NOT support the use of both team-config and old-profiles at the same time.

This change may make it less of an unexplained problem for users who do not follow our recommendations, which is good.

Does the embedded SCS and the old SCS plugin store content in a different format within the vault, or does the SCS just store any blob that it is given? I ask that question to ask my next question - Have you confirmed that when you change back to the team-config directory, everything still works as expected?

dkelosky:
> I think our policy is to NOT support the use of both team-config and old-profiles at the same time.

Sorry, I didn't realize this.  If that is our policy, I think that would eliminate this need for this PR.  In practice, I still need both old and new profiles for things like `zowe-api-dev` which seems to rely on [old profiles](https://github.com/zowe/zowe-cli/issues/982).  

> Does the embedded SCS and the old SCS plugin store content in a different format within the vault, or does the SCS just store any blob that it is given?

@t1m0thyj - do you know this answer?

> Have you confirmed that when you change back to the team-config directory, everything still works as expected?

Yes, I confirmed this with and without daemon mode.

gejohnston (approved):
Since the team-config continues to work when the user switches back from old-style profiles, I think that this change will be beneficial.

t1m0thyj:
> > I think our policy is to NOT support the use of both team-config and old-profiles at the same time.
> 
> Sorry, I didn't realize this. If that is our policy, I think that would eliminate this need for this PR. In practice, I still need both old and new profiles for things like `zowe-api-dev` which seems to rely on [old profiles](https://github.com/zowe/zowe-cli/issues/982).

Although we may not officially support the simultaneous use of old and new config, I see no harm in making them compatible with each other, as long as there are no negative implications of doing so.

> > Does the embedded SCS and the old SCS plugin store content in a different format within the vault, or does the SCS just store any blob that it is given?
> 
> @t1m0thyj - do you know this answer?

I believe the SCS just stores any blob that it is given, but it's missing a few enhancements that have been made to the Imperative DefaultCredentialManager in the next branch, such as supporting storage of secure values longer than 2,560 bytes on Windows. Perhaps this is a concern associated with this PR - if secure values for Zowe CLI on my Windows machine exceed 2,560 bytes, then the SCS will fail to load them.

dkelosky:
Thanks @t1m0thyj - as discussed on our call this morning, maybe the size limit is acceptable since running with both team config and scs is not recommended.

zFernand0:
> I think this is technically a breaking change since `@zowe/cli` will no longer be added to [`allServices`](https://github.com/zowe/zowe-cli-scs-plugin/blob/master/src/credentials/KeytarCredentialManager.ts#L50). 

Thanks for using `next` as the base for doing this.
Also, thanks for fixing the annoying prompt with the strange `newOpts` object 😋

zFernand0 (approved):

dkelosky:
> Also, thanks for fixing the annoying prompt with the strange newOpts object

I _thought_ I copied that code from you, but I must have tweaked it in testing 😄

t1m0thyj (approved):
LGTM! Seems like a clever way to resolve the issue, and also fixes https://github.com/zowe/zowe-cli/issues/936 👍

awharn (approved):
LGTM! 😄

PR #604, opened 2021-04-29
Node 16 - Next

awharn:
Updates a dev dependency to better support node 16 on Next


awharn:
Going to merge the change from master to next

PR #610, opened 2021-05-18
Enhance ProfileInfo API

gejohnston:
Add new functions to the ProfileInfo API.

-  The following function converts a IProfAttrs structure into a IProfileLoaded structure, which is supplied to many other functions in ZoweExplorer.
```
    public static ProfileInfo.profAttrsToProfLoaded(
  ​     profAttrs: IProfAttrs,
  ​     dfltProfLoadedVals?: any
    ): IProfileLoaded
```

-  The following function creates a session from an array of IProfArgAttrs.
```
    public static ProfileInfo.createSession(
        profArgs: IProfArgAttrs[],
        connOpts: IOptionsForAddConnProps = {}
    ): Session
```

- Refactored ConnectionPropsForSessCfg.addPropsOrPrompt to call a new function named
  ConnectionPropsForSessCfg.resolveSessCfgProps, which can now also be called by ProfileInfo.createSession.

- Added a new option named getSecureVals to the mergeArgsForXXX( ) functions. That option will automatically get the actual values for secure arguments, instead of just placeholders, which require apps to call a second function loadSecureArg( ). Here is the signature of one function with the new option:
```
public mergeArgsForProfile(
        profile: IProfAttrs,
        mergeOpts: IProfMergeArgOpts = {getSecureVals: false}
    ): IProfMergedArg
```


packages/config/src/ProfileInfo.ts
```diff
@@ -666,6 +788,49 @@ export class ProfileInfo {
...
+     *
+     * @returns A session that can be used to connect to zosmf;
+     */
+    private static initZosmfSessCfg(profArgs: IProfArgAttrs[]): ISession {
+        const sessCfg: any = {};
+
+        // the set of names of arguments in IProfArgAttrs used in ISession
+        const profArgNames = [
+            "host", "port", "user", "password", "rejectUnauthorized",
+            "protocol","basePath", "tokenType", "tokenValue", "authtoken"
```
t1m0thyj, line 807: `"authtoken"` -> `"authToken"`

packages/config/src/ProfileInfo.ts
```diff
@@ -359,6 +363,41 @@ export class ProfileInfo {
...
+     *      Options that alter our actions. See IOptionsForAddConnProps.
+     *      The connOpts parameter need not be supplied.
+     *      Default properties may added to any supplied connOpts.
+     *      The only option values used by this function are:
+     *          connOpts.requestToken
+     *          connOpts.defaultTokenType
+     *
+     * @returns A session that can be used to connect to zosmf;
+     */
+    public static createZosmfSession(
```
t1m0thyj, line 385: It may be confusing to call this method `createZosmfSession`, since it doesn't use the ZosmfSession type (defined in https://github.com/zowe/zowe-cli/blob/master/packages/zosmf/src/ZosmfSession.ts), and nothing about this method seems specific to z/OSMF.

packages/config/src/ProfileInfo.ts
```diff
@@ -666,6 +788,49 @@ export class ProfileInfo {
...
+    /**
+     * Initialize a zosmf session configuration object with the arguments
+     * from profArgs
+     *
+     * @param profArgs
+     *      An array of profile argument attributes.
+     *
+     * @returns A session that can be used to connect to zosmf;
+     */
+    private static initZosmfSessCfg(profArgs: IProfArgAttrs[]): ISession {
```
t1m0thyj, line 801: So far I think ZosmfSession and code that is specific to z/OSMF sessions has lived in the `@zowe/cli` repo. For this `initZosmfSessCfg` method and the `createZosmfSession` method above:
* If they are specific to z/OSMF sessions, do they belong in a Zowe SDK instead of Imperative?
* If they are not specific to z/OSMF sessions, should we rename these methods in Imperative?

The latter seems true to me - that these methods are not specific to z/OSMF sessions. They could be used for any session that connects to a `protocol://host:port`, and supports basic and token auth.

Perhaps this method could be renamed to `initSessCfg` to make it more generic, so that developers could feel free to use it for other types of sessions: CICS, DB2, etc.

packages/config/src/ProfileInfo.ts
dkelosky: good point - I guess similarly there is `initZosmfSessCfg`.  It seems like z/OSMF related topics would belong in Zowe CLI instead of our generic imperative infrastructure piece.

packages/config/src/ProfileInfo.ts
gejohnston: Good catch. This might have been very annoying to debug when a problem occurred.

packages/config/src/ProfileInfo.ts
gejohnston: Because I was working closely with the ZoweExplorer team, I think I had zosmf on the brain. I renamed the functions to createSession and initSessCfg. While createSession just works for zosmf, if a calling app needed additional properties for a different session type, the app could create a thin layer that coerces the session returned by createSession into the session of choice and add the required properties to that session.

t1m0thyj (approved):
LGTM!

zFernand0 (approved):

PR #615, opened 2021-06-07
[next] Team config overrides

t1m0thyj:
This PR restores support for overrides, which were temporarily removed in the "next" release. This is a test implementation that is subject to change after team discussion, so I've opened as a draft PR.

The following config which used to reside in `~/.zowe/settings/imperative.json` can now go directly in a zowe.config.json file:
```json
{
  "overrides": {
    "CredentialManager": "<name of plugin that provides custom credential manager>"
  }
}
```

TODO:
* [x] Resolve conflicts with next
* [x] Make "overrides" property optional and hidden by default
* [x] Revert unintended breaking changes (see comment below)
* [ ] Add tests if needed
* [ ] Handle switching between overrides in daemon instance


packages/imperative/src/OverridesLoader.ts
```diff
@@ -74,30 +44,33 @@ export class OverridesLoader {
...
+    const overrides: IImperativeOverrides = config.overrides || {};
 
     const ZOWE_CLI_PACKAGE_NAME = `@zowe/cli`;
 
     // The manager display name used to populate the "managed by" fields in profiles
-    const displayName: string = (
-        overrides.CredentialManager != null
+    // App settings is not configured - use the CLI display name OR the package name as the manager name
+    let displayName: string = config.productDisplayName || config.name;
+    if (overrides.CredentialManager != null
```
zFernand0, line 58: Thanks for simplifying the long one-liner! 😋

packages/imperative/src/OverridesLoader.ts
```diff
@@ -35,37 +35,7 @@ export class OverridesLoader {
...
-  /**
-   * Load the baked-in zowe CredentialManager and initialize it.
-   * If we need to reinstate 3rd party overrides, delete this function and
-   * rename loadCredentialManagerOld.
-   *
-   * @internal
-   * @param {IImperativeConfig} config - the current {@link Imperative#loadedConfig}
-   * @param {any} packageJson - the current package.json
-   */
-  public static async loadCredentialManager(
```
zFernand0, line 50: 👍 Thanks for combining the two functions `loadCredentialManager` and `loadCredentialManagerOld` 😋

packages/imperative/src/OverridesLoader.ts
```diff
@@ -110,7 +83,7 @@ export class OverridesLoader {
         displayName,
 
         // zowe cli will always add `Zowe` to it's list of service names
-        service: config?.name === ZOWE_CLI_PACKAGE_NAME ? DefaultCredentialManager.SVC_NAME : config?.name,
+        service: config?.credentialServiceName || (config.name !== ZOWE_CLI_PACKAGE_NAME ? config.name : null),
```
zFernand0, line 86: I noticed `overrides` was added to `IConfig`, but I couldn't find `credentialServiceName`.
Should it be added?

packages/config/src/doc/IConfig.ts
```diff
@@ -9,11 +9,13 @@
...
 import { IConfigProfile } from "./IConfigProfile";
 
 export interface IConfig {
     $schema?: string;
     defaults: { [key: string]: string };
     profiles: { [key: string]: IConfigProfile };
     plugins: string[];
-}
\ No newline at end of file
+    overrides: IImperativeOverrides;
```
zFernand0, line 20: 👍  +  😕 (see previous comment)

zFernand0 (commented):
Code changes look great!
I have not tested the functionality yet 😞 
Will approve once we have a quick discussion about it. 😋

packages/imperative/src/OverridesLoader.ts
t1m0thyj: `credentialServiceName` is a property that already existed in the IImperativeConfig interface: https://github.com/zowe/imperative/blob/next/packages/imperative/src/doc/IImperativeConfig.ts#L316

packages/imperative/src/OverridesLoader.ts
t1m0thyj: Something we'll probably need to discuss as a team: I'm not sure if combining them was totally a good idea. We may need to undo some of these changes to split out some logic again. 

Currently in the "next" branch, in the case that team config files do not exist, we maintain our old behavior of enabling secure credentials only when the CredentialManager setting is defined.

The implementation in this PR is that secure by default always rules. Which means if you have old school profiles stored in plain text, updating to Zowe v2 would break you. And there's no longer a `zowe scs update` command to easily unbreak you, since we recommend uninstalling the SCS when upgrading to Zowe v2.

t1m0thyj:
Closing since the team decided against adding the "overrides" object to zowe.config.json.

Will open another PR that enables the `~/.zowe/settings/imperative.json` file when Zowe v2 config is active.

PR #618, opened 2021-06-14
AbstractRestClient.request() is O(n2)

mheuzey:
### The Problem

When `AbstractRestClient.request()` receives data chunks, it continuously creates new Buffers to hold the data.

https://github.com/zowe/imperative/blob/83511b7d04ab203cfee5d60db3958b954a8cade8/packages/rest/src/client/AbstractRestClient.ts#L588-L595

This means that as the size of the response increases, it takes exponentially longer to store the data, making this function O(n<sup>2</sup>).

### The Solution

The proposed change will store the incoming chunks in an array, and then do a single concatenate at the end. This way, each chunk is written only once during data transfer, and then once more at the end of the response, reducing the time complexity to O(2n) = O(n).

Note that at the time of concatenation, 2x the size of the response will be used in memory, as the response will be written both in the final Buffer, and the chunks array (until garbage collection occurs). This is no worse than the current scenario though, as the response is stored in both the final Buffer as well as the previous Buffer.

### Alternative

An alternative solution would be to initially allocate the Buffer based on the information in the `content-length` HTTP header, and then write to the Buffer with `this.mData.fill(respData)`. This would cut the required memory in half, as each chunk is kept only until it is added to the Buffer, then discarded. It also would not require a lengthy write after the last chunk is received. However, there are several issues that would need to be checked with this method:

- If no `content-length` header is sent, the Buffer cannot be preallocated
- If a massive `content-length` is declared by a malicious agent, it could cause unneeded memory allocation, potentially crashing the system
- If the received chunks do not match the declared length, it could result in either:
  - Wasted time spent clearing memory if `Buffer.alloc()` was used to overallocate
  - Potentially sensitive memory being shown if `Buffer.allocUnsafe()` was used to overallocate
  - A new Buffer being created and filled if the chunks exceed the declared length

Altogether, the proposed solution is simpler and would require fewer code changes.

### Thoughts

I haven't done rigorous benchmarking, but I ran a few requests, and came up with the below data

image


mheuzey:
I opened this as draft since making requests is at the core of what Imperative does, and I wanted to give people a chance to bring up any issues or test the changes. All the tests are passing, so I'll give it a couple more days, and if no one raises any issues, I'll convert it to regular PR.

t1m0thyj:
@mheuzey Thanks for contributing this performance improvement!

Discussed with the CLI team since this is a pretty high-impact change in Imperative that would affect all REST API calls. We don't foresee any problems or potential breaking changes caused by this PR. 🙂 

Could you please update this branch with the latest changes from master, and add an entry to the changelog for this PR (see [changelog update guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md#changelog-update-guidelines))?

zFernand0 (approved):
🚀

mheuzey:
@t1m0thyj Sure, I can do that. I hadn't made a Changelog entry because this is more of a backend change, and I wasn't sure if that got called out.

mheuzey:
I thought editing in Github's editor would be simpler than editing locally and pushing, but apparently DCO doesn't like GitHub’s verified signature...

zFernand0:
if this build succeeds: https://wash.zowe.org:8443/job/imperative/job/PR-625/1/
then we should add a note to the contributing guidelines about branch names with special characters.
Thanks @mheuzey

zFernand0:
Seems like this worked 😋 
We will need to make a quick update to https://github.com/Zowe/zowe-cli/blob/master/CONTRIBUTING.md#sign-all-of-your-git-commits

PR #622, opened 2021-06-21
Update config arguments

awharn:
Replaces --user with --user-config (--uc) and --global with --global-config (--gc) due to problems that arise when trying to create a config through a session to a config service, such as the API Mediation Layer.

Thanks for the help @t1m0thyj


zFernand0 (commented):
Code looks great.
Testing in a few minutes 😋

t1m0thyj:
Normally I'm not sure if we care about changelog updates for PRs whose destination is the "next" branch. But since this is a breaking change in option names for some of the "zowe config" commands, should we document this in the changelog?

zFernand0:
> Normally I'm not sure if we care about changelog updates for PRs whose destination is the "next" branch. But since this is a breaking change in option names for some of the "zowe config" commands, should we document this in the changelog?

It will only be considered a breaking change until next moves to zowe-v2-lts.
Then, it can be removed from the changelog (or at least omitted from the list of breaking changes), right?
😋

gejohnston (approved):
Looks good.

t1m0thyj (approved):
LGTM!

zFernand0 (approved):

PR #623, opened 2021-06-21
Add IApimlSvcAttrs to identify a service in APIML.

gejohnston:
Add the IApimlSvcAttrs interface to identify a service in APIML and include such property, named apimlConnLookup, in the IImperativeConfig interface.


packages/imperative/src/doc/IApimlSvcAttrs.ts
```diff
@@ -0,0 +1,71 @@
...
+     * version of the REST API. This attribute is defined by the REST service
+     * programmer. An example is "api/v1". If no gatewayUrl is supplied, but
+     * multiple versions of the REST API exists in APIML, the first instance
+     * of the apiId returned by APIML will be used. You might choose to not
+     * specify a gatewayURL, if your client app supports all available versions
+     * of the associated REST service.
+     * @type {string}
+     * @memberof IApimlSvcAttrs
+     */
+    gatewayUrl?: string;
```
dkelosky, line 49: a topic that comes up time-to-time is what goes in imperative versus zowe cli.  since imperative is a generic cli infrastructure, should this be in the zowe cli code base?

t1m0thyj:
LGTM!

t1m0thyj (approved):
LGTM! Sorry for accidentally closing 😢

packages/imperative/src/doc/IApimlSvcAttrs.ts
gejohnston: Yes. The team discussed this very question in a standup meeting. I believe you were on vacation at the time. This was our rationale:

- We want a new structure (IApimlSvcAttrs) to enable plugins to tie themselves to an APIML service. We can then ask APIML for the configuration data for the plugin to connect to that service.
    - IImperativeConfig is what each plugin uses to define itself and its options.
    - Thus, we want the new IApimlSvcAttrs structure to be added as a property to IImperativeConfig.
    - IImperativeConfig could not easily import IApimlSvcAttrs from zowe-cli.
- If IApimlSvcAttrs was instead added to Zowe-cli, then zowe-cli would have to extend IImperativeConfig and then add the IApimlSvcAttrs to that new extended interface.
    - All plugins would have to be modified to define themselves using the new extended Zowe-CLI interface instead of IImperativeConfig. That seemed like a big change for every plugin.
- Even though IApimlSvcAttrs is not complely generic CLI functionality (ie, non-mainframe, and non-APIML), it is optional. It is also only an interface, and no logic in imperative performs programmatic actions on the properties of IApimlSvcAttrs. We decided that these limited charactistics of IApimlSvcAttrs made it acceptable for inclusion in Imperative.

dkelosky: im sorry you had to rehash that, but thank you for the explanation @gejohnston

PR #625, opened 2021-06-28
AbstractRestClient.request is order n2

mheuzey:
Same as #618, but without parenthesis.


zFernand0 (approved):
Thanks for doing it! 😋

t1m0thyj (approved):

PR #627, opened 2021-07-07
Add the framework for a config auto-init command

awharn:
Adds the base handler and framework for a `config auto-init` command.
Adds relevant tests and snapshots.
Modifies layers.merge() to allow for a dry run that doesn't modify the in memory config layer, returning a clone.
Modifies findSecure() to make the interface public for external credential masking during auto-init dry run.
Exposes all of the missing interfaces for REST client related classes and functions.

Known issues:
 - layers.merge() does not entirely support JSON Comments. Current behavior places priority on incoming comments, and existing JSON objects. A solution could not be found to place priority for both on existing comments/JSON.


awharn:
Whoops, looks like VSCode stopped signing my comments when it updated. 😢

packages/imperative/src/doc/IImperativeAutoInitCommandConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+*
+*/
+
+import { ICommandProfileAutoInitConfig } from "../../../cmd/src/doc/profiles/definition/ICommandProfileAutoInitConfig";
+
+export interface IImperativeAutoInitCommandConfig {
+    /**
+     * Command definition for "config auto-init" command
+     */
+    autoInitConfig?: ICommandProfileAutoInitConfig;
```
t1m0thyj, line 18: Do we expect we may add more properties to this interface in the future?
If not, would it make sense to flatten the object structure to remove an unnecessary level?

Nested (current): `configAutoInitCommandConfig: {
    autoInitConfig: {...}
}`
Flattened (alternative): `configAutoInitCommandConfig: {...}`

Sorry this is probably a dumb question 😢

packages/imperative/__tests__/Imperative.test.ts
```diff
@@ -624,8 +636,42 @@ describe("Imperative", () => {
...
+                        }
+                    };
+                })
+            });
+
+            const autoGenCmdTree = Imperative.addAutoGeneratedCommands(JSON.parse(JSON.stringify(mockRootCmdTree)));
+            expect(mocks.CompleteAutoInitCommandBuilder.getAutoInitCommand).toHaveBeenCalledTimes(1);
+            // Expect 2 command groups added (auth, config, and profiles)
+            expect(autoGenCmdTree.children.length).toBe(1);
+            expect(autoGenCmdTree.children[0].children.length).toBe(1);
```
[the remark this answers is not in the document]
t1m0thyj: Sorry, I missed that the test is already doing this 🙂

packages/imperative/src/doc/IImperativeAutoInitCommandConfig.ts
awharn: I don't think we are planning on adding anything else in the near future.

__tests__/__integration__/imperative/src/imperative.ts
```diff
@@ -75,6 +75,21 @@ export const config: IImperativeConfig = {
...
+    configAutoInitCommandConfig: {
+        autoInitConfig: {
+            handler: __dirname + "/cli/config/FruitAutoInitHandler",
+            provider: "Fruit Manager",
+            autoInit: {
+                options: [
+                    hostOption,
+                    portOption,
+                    userOption,
+                    passwordOption
```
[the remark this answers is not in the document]
awharn: That sounds like a good idea! 😄

t1m0thyj (approved):
LGTM! Not sure if we should update the changelog, or add "no-changelog" label to make Jenkins happy? 🙂

tslint.json
```diff
@@ -21,7 +21,8 @@
       -1,
       0,
       1,
-      2
+      2,
+      3
```
zFernand0, line 25: are we going to go back and remove all the `tslint:disable-next-line: no-magic-number`? 
Or just wait for the tslint to eslint formal migration?

packages/rest/index.ts
```diff
@@ -11,6 +11,10 @@
 
 export * from "./src/client/types/HTTPVerb";
 export * from "./src/client/doc/IHeaderContent";
+export * from "./src/client/doc/IHTTPSOptions";
+export * from "./src/client/doc/IOptionsFullResponse";
+export * from "./src/client/doc/IRestClientError";
+export * from "./src/client/doc/IRestClientResponse";
```
zFernand0, line 17: Thank you!

packages/imperative/src/config/cmd/auto-init/handlers/BaseAutoInitHandler.ts
```diff
@@ -0,0 +1,170 @@
...
+import stripAnsi from "strip-ansi";
+import * as open from "open";
+import * as JSONC from "comment-json";
+import * as lodash from "lodash";
+import { ImperativeConfig, TextUtils } from "../../../../../../utilities";
+
+/**
+ * This class is used by the auto init command handler as the base class for its implementation.
+ */
+export abstract class BaseAutoInitHandler implements ICommandHandler {
```
zFernand0, line 25: I was wondering if BaseAutoInitHandler and BaseAuthHandler should've shared a common handler, but talked myself out of it.
While they do have some basic structure + abstract properties and function in common, it may be more trouble than what's worth.
We may reconsider this if/when we get a 3rd BaseHandler that required the same abstract properties and functions.

zFernand0 (approved):
LGTM

tslint.json
awharn: I think it might be best to just wait for the migration.

gejohnston (approved):
I have not been very involved with the decision-making, but the logic looks good to me. My comments are all superficial.

packages/imperative/src/config/cmd/auto-init/builders/CompleteAutoInitCommandBuilder.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Generate a complete command for automatic initialization of a user configuration
+ */
+export class CompleteAutoInitCommandBuilder {
+    /**
+     * Get the complete auth group of commands
+     * @param {ICommandProfileAutoInitConfig} autoInitConfig - mapping of profile types to auto init configs
+     * @param {Logger} logger - logger to use in the builder classes
+     * @param {IImperativeAutoInitCommandConfig} autoInitCommandConfig - config that allows command definitions to be overridden
```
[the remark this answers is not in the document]
awharn: It is not used - when I removed the extra layer per Timothy's earlier comment, I removed that parameter as it wasn't necessary. I guess I forgot to remove it there. Oops. 😅

PR #635, opened 2021-07-21
Fix validate command not returning an error code

awharn:
Allows the `validate plugins` command to return an error code and display an error if validation finds problems. Planned to enable this functionality by default in V2, but useful for upcoming scripting in V1 (i.e. extended web-help generation).

Note: Update Zowe CLI to add `--no-fail-on-error` in the post-install script validatePlugins.js once the fail-on-error option is true by default in V2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- BugFix: Return with an error code when the `plugins validate` command is run and plugins have errors. Disable this behavior with `--no-fail-on-error` option. Also adds `--fail-on-warning` option to return with an error code when plugins have warnings. [#463](https://github.com/zowe/imperative/issues/463)
+- Enhancement: The `plugins validate` command will return an error code when plugins have errors if the new `--fail-on-error` option is specified. Also adds `--fail-on-warning` option to return with an error code when plugins have warnings. [#463](https://github.com/zowe/imperative/issues/463)
 
 ## `4.13.4`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Return with an error code when the `plugins validate` command is run and plugins have errors. Disable this behavior with `--no-fail-on-error` option. Also adds `--fail-on-warning` option to return with an error code when plugins have warnings. [#463](https://github.com/zowe/imperative/issues/463)
 ## `4.13.4`
 
 - BugFix: Added missing periods at the end of command group descriptions for consistency. [#55](https://github.com/zowe/imperative/issues/55)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the Imperative package will be documented in this file.
 ## Recent Changes
 
 - BugFix: Return with an error code when the `plugins validate` command is run and plugins have errors. Disable this behavior with `--no-fail-on-error` option. Also adds `--fail-on-warning` option to return with an error code when plugins have warnings. [#463](https://github.com/zowe/imperative/issues/463)
+
 ## `4.13.4`
 
 - BugFix: Added missing periods at the end of command group descriptions for consistency. [#55](https://github.com/zowe/imperative/issues/55)
--- a/__tests__/src/packages/imperative/plugins/suites/ValidatePlugin.ts
+++ b/__tests__/src/packages/imperative/plugins/suites/ValidatePlugin.ts
@@ -345,7 +345,7 @@ describe("Validate plugin", () => {
                 expect(result.stdout).toContain("The plugin's configuration does not contain an 'imperative.rootCommandDescription' property.");
             });
 
-            it("missing rootCommandDescription property - warning", () => {
+            it("missing rootCommandDescription property - error", () => {
                 const testPlugin = "missing_rootCommandDescription";
                 const fullPluginPath = join(testPluginDir, "error_plugins", testPlugin);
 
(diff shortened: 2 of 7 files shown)


awharn:
Note: The security hotspot already existed - I just added a try/catch block

packages/imperative/src/plugins/cmd/validate/validate.handler.ts
```diff
@@ -71,10 +75,17 @@ export default class ValidateHandler implements ICommandHandler {
...
       } else {
-        this.displayPluginIssues(pluginName, params.response);
+        err = this.displayPluginIssues(pluginName, params.response, failOnWarning);
       }
     }
+
+    if (err === true && params.arguments.failOnError) {
+      params.response.console.log("\n");
+      params.response.console.error(TextUtils.chalk.red("Problems detected during plugin validation. Please check above for more information."));
+      params.response.data.setExitCode(1);
```
zFernand0, line 87: Should we eventually implement different exit codes, based on the type of error?
For example:
- Imperative APIs (e.g.  1 - 99)
- CLI APIs (e.g. 100 - 199)
- HTTP error codes (e.g. 300 +)

packages/imperative/src/plugins/cmd/validate/validate.definition.ts
```diff
@@ -44,5 +62,9 @@ export const validateDefinition: ICommandDefinition = {
       description: "Validate all installed plug-ins",
       options    : ""
     },
+    {
+      description: "Validate a plug-in named my-plugin, and treat warnings as errors",
+      options    : "my-plugin --fow"
```
zFernand0, line 67: Shouldn't the example always use the full name for options instead of aliases?

zFernand0 (commented):
Behavior seems consistent, and we have way more control over the exit code of the command with the `--foe` and `--fow` options.
LGTM so far!

packages/imperative/src/plugins/cmd/validate/validate.definition.ts
awharn: I don't know how consistent we are with that rule - https://github.com/zowe/zowe-cli/blob/master/packages/cli/src/zostso/start/address-space/AddressSpace.definition.ts

But, I will change it. 😃

packages/imperative/src/plugins/cmd/validate/validate.handler.ts
awharn: That'd be nice, but would probably require a good bit of re-work.

zFernand0: More like a lot! 😋

packages/imperative/src/plugins/cmd/validate/validate.definition.ts
zFernand0: Thanks for changing it.
Maybe we should open an issue about this one too

packages/imperative/src/plugins/cmd/validate/validate.handler.ts
zFernand0: https://github.com/zowe/imperative/issues/636

packages/imperative/src/plugins/cmd/validate/validate.definition.ts
zFernand0: https://github.com/zowe/imperative/issues/637

awharn:
Updating the PR to not implement this new functionality by default. Plan to enable the fail-on-error option by default in V2, just in case of unexpected breaking changes.

__tests__/src/packages/imperative/plugins/suites/ValidatePlugin.ts
```diff
@@ -221,6 +345,24 @@ describe("Validate plugin", () => {
                 expect(result.stdout).toContain("The plugin's configuration does not contain an 'imperative.rootCommandDescription' property.");
             });
 
+            it("missing rootCommandDescription property - warning", () => {
```
t1m0thyj, line 348: This test looks like it may be for an error rather than a warning.

packages/imperative/src/plugins/cmd/validate/validate.definition.ts
```diff
@@ -35,6 +35,24 @@ export const validateDefinition: ICommandDefinition = {
...
+      description: "Enables throwing an error and setting an error code if plugin validation detects an error",
+      required: false,
+      defaultValue: false
+    },
+    {
+      name: "fail-on-warning",
+      aliases: ["fow"],
+      type: "boolean",
+      description: "Treat validation warnings as errors. Requires fail-on-error.",
+      required: false,
```
t1m0thyj, line 52: Should this have defaultValue of false? I guess that's the default behavior so not necessarily required. Just curious because fail-on-error has defaultValue of false defined 🙂

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Return with an error code when the `plugins validate` command is run and plugins have errors. Disable this behavior with `--no-fail-on-error` option. Also adds `--fail-on-warning` option to return with an error code when plugins have warnings. [#463](https://github.com/zowe/imperative/issues/463)
```
t1m0thyj, line 7: Should the changelog be updated since `--fail-on-error` is no longer the default for zowe-v1-lts?

t1m0thyj (approved):
LGTM, left a few questions 😛

packages/imperative/src/plugins/cmd/validate/validate.definition.ts
awharn: I did not define it because of the implicit requirement of the other value. I can remove the default for the other if you like, I just added it because we are planning to make it true on next.

CHANGELOG.md
awharn: Done. 😄

__tests__/src/packages/imperative/plugins/suites/ValidatePlugin.ts
awharn: Yep, you're right. Good catch.

packages/imperative/src/plugins/cmd/validate/validate.definition.ts
zFernand0: I'm in favor of consistency 😅 
either add it for both, or remove it for both 😋

zFernand0 (approved):
LGTM! 😋

PR #641, opened 2021-07-26
[next] Fix prompts that didn't work in daemon mode

t1m0thyj:
* Fix prompts for `prompt*` values so that they work in daemon mode
* Deprecate `CliUtils.promptForInput` and `CliUtils.promptWithTimeout`
* Replace them with `CliUtils.readPrompt` which uses the same "read" module as `npm login`
* Add options to never timeout if `secToWait` is 0 (default: 600), or customize the `maskChar` (default: "*")
* Remove "node-pty" dev dependency which is no longer needed since we can pipe to stdin with `echo`
* Note: There is a known issue with the "read" package: https://github.com/npm/read/issues/26


awharn:
Is this change going to be a problem? Would we maybe want to investigate a change for more secure prompting, where we don't output anything instead of masking with *'s? I just ask because:

1) Masking still gives away the length
2) Some of us have hit the problem in VSCode where masked values, in the integrated terminal, appear in plain text for a split second.

I know that 2 isn't really our problem, and the user can disable it in VSCode. Some sites might have concerns with 1, though.

packages/utilities/src/CliUtils.ts
```diff
@@ -528,9 +530,11 @@ export class CliUtils {
         }
 
         // Ensure that we use a reasonable timeout
-        const maxSecToWait = 900; // 15 minute max
```
dkelosky, line 531: i thought the 15 minute wait came out of some parking lot discussion on this topic... i forget the details and don't have an email.  i'm just sharing my own recollection

dkelosky (approved):

dkelosky:
@awharn has good comments.  we do use `*` in other secure prompts, e.g. `zowe config init` or when information is missing:
```
{
    "profiles": {
        "lpar1": {
            "properties": {
                "host": "some.host.com"
            },
            "profiles": {
                "zosmf": {
                    "type": "zosmf",
                    "properties": {
                        "port": 1443
                    }
                }
            }
        }
    },
    "defaults": {
        "zosmf": "lpar1.zosmf"
    },
    "plugins": []
}
```

Then `zowe jobs list jobs`

packages/utilities/src/CliUtils.ts
t1m0thyj: I remember we discussed what the default timeout should be (10 min), but I guess I forgot the discussion about the maximum timeout 😢 IMO 15 minutes seems to be kind of an arbitrary limit, but I can change it back if you'd like 🙂

gejohnston:
Regarding stars or nothing: Some literature claims that "shoulder surfing" is less of a problem than brute force attacks. Other literature claims that unix displayed no characters for passwords because it was easier to put the terminal into no-echo mode than to replace the characters with stars.

Personally I like the confirmation of stars, but I do not feel strongly that it must be that way.

gejohnston (approved):
By consolidating into one Zowe function, it should be easier to make changes for consistent behavior in many of our prompts.

awharn:
I'm willing to leave it as stars. Really, it doesn't matter to me, and just wanted to discuss the security implications with the team before settling, making sure everyone was on the same page. Turns out the VSCode issue is particular to my environment, when you use a VSCode remote over SSH and use the integrated terminal. Setting `Local Echo Latency Threshold` to `-1` solves that problem, if anyone else hits that issue (VSCode doesn't know that the character is supposed to be hidden).

awharn (approved):

zFernand0:
Code changes LGTM! 😋

awharn (approved):

t1m0thyj:
Awaiting feedback from UX about whether we should consistently mask secure values with `*`, or show nothing.

t1m0thyj:
Reopened the PR since the team unanimously decided against showing `*` characters in secure prompts, and the PR has been updated to reflect this.

gejohnston (approved):

zFernand0 (approved):

awharn (approved):

PR #621, opened 2021-06-17
Fix Markdown special characters not being escaped in web help

t1m0thyj:
Fixes #620 and #323

* Fixed inconsistent behavior in `DefaultHelpGenerator` where Markdown special characters weren't always escaped in descriptions
* Fixed URL autolinking which stopped working in web help, by passing `linkify` option to markdown-it
* Fixed command descriptions not being detected if they begin with non-alphanumeric character (e.g., in the sample plugin)
* Fixed subgroups not being displayed under their own heading (e.g., in the @broadcom/caview-for-zowe-cli plugin)
* Updated Jenkinsfile to use npm@7 so that package-lock will stay at lockfile version 2 rather than getting reverted to version 1

PR #649, opened 2021-07-28
Fix character escaping behavior in web help

t1m0thyj:
* Fix `escapeMarkdown` being called twice for positional descriptions, causing characters to be double escaped
  * This was a regression introduced in #621
  * Thanks to @awharn for helping to diagnose 🙂 
* Fix backslash character not being escaped, causing it to disappear at some places in web help
* Remove hack to "Remove backslash escapes from URLs" that used to be necessary to work around "marked" bug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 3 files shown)


zFernand0 (approved):
LGTM! :yum:

awharn (approved):

Issue #648, opened 2021-07-28
[next] Remove authToken and replace with tokenType/tokenValue


t1m0thyj:
Currently, the next branch is in a halfway state where CLI help and old-school profiles use separate args tokenType and tokenValue, while team config uses a single arg authToken.

For v2 LTS, we decided to use tokenType and tokenValue everywhere.

For v3 LTS, we can implement https://github.com/zowe/imperative/issues/514

PR #654, opened 2021-08-05
[next] Revert authToken back to tokenType and tokenValue

t1m0thyj:
This is part of the fix for https://github.com/zowe/imperative/issues/648.

After this PR is merged and consumed in Zowe CLI, more tests will need to be updated there.


awharn (approved):

gejohnston (approved):

zFernand0 (commented):
Should we add a changelog since this is a next-only breaking change?
Changes LGTM though

zFernand0 (approved):
LGTM!

PR #655, opened 2021-08-06
Improve Command Suggestions

awharn:
- Uses a more performant package for levenshtein suggestions
- Adds command aliases to the suggestions
- Add suggestions on arguments for the second and third command groups


package.json
```diff
@@ -55,13 +55,13 @@
     "cli-table3": "0.5.1",
     "dataobject-parser": "1.2.1",
     "deepmerge": "3.0.0",
+    "fastest-levenshtein": "^1.0.12",
```
zFernand0, line 58: Maybe remove the `^` caret?

zFernand0 (commented):
Code changes look good
Wondering about the commented `buildResponseObject` function.
Also, do we have issues for the `TODO` items around `getFullCommandName`? 
(if not, should we create them?)

t1m0thyj (approved):
LGTM. Having aliases added to the list that Levenshtein checks should be much better 🙂

zFernand0 (approved):

PR #658, opened 2021-08-16
[next] Update next with latest

awharn:
Updates next to contain newest content on Master branch. Also includes two enhancements:

- Modifies `zowe plugins validate` to return RC 1 and display an error on a validation error (breaking)
- Modifies levenshtein logic to support Daemon mode


package.json
```diff
@@ -88,10 +88,10 @@
     "yargs": "15.3.1"
   },
   "devDependencies": {
-    "@types/bootstrap": "^4.3.1",
-    "@types/clipboard": "^2.0.1",
+    "@types/find-up": "^2.1.1",
```
t1m0thyj, line 91: Is the addition of `@types/find-up` intentional here?

t1m0thyj (approved):
LGTM! Just one comment.

package.json
awharn: Nope, it kinda just slipped in. There were a lot of merge conflicts.

gejohnston (approved):

PR #659, opened 2021-08-18
[next] Migrate from TSLint to ESLint

t1m0thyj:
ℹ️ This PR is easier to review with whitespace changes hidden. It will still be a big diff though 😢


packages/imperative/src/plugins/utilities/PluginIssues.ts
```diff
@@ -19,194 +19,194 @@ import { readFileSync } from "jsonfile";
...
+     *
+     * @param {string} pluginName - The name of the plugin
+     *
+     * @param {string} issueSevs - An array of issue severities.
+     *
+     * @returns {boolean} - True if any plugin issues have any of the severities.
+     *                      False otherwise.
+     */
+    public doesPluginHaveIssueSev(pluginName: string, issueSevs: IssueSeverity[]): boolean {
+        if (Object.prototype.hasOwnProperty.call(this.pluginIssues, pluginName)) {
```
gejohnston, line 110: Just curious - what is the ESLINT rationale for using the Object.prototype form of hasOwnProperties?

gejohnston (approved):
No issues. Just a couple of questions.

packages/imperative/src/plugins/utilities/PluginIssues.ts
t1m0thyj: I think the reason they give is that calling `x.hasOwnProperty` directly on an object could be vulnerable to someone maliciously overriding the behavior of `hasOwnProperty` for the object `x`. I don't know how much we care about this, but it wasn't too hard to "fix" so I followed the ESLint recommendation 🙂 For more details about their rationale, see https://eslint.org/docs/rules/no-prototype-builtins.

packages/cmd/__tests__/profiles/CommandProfileLoader.test.ts
```diff
@@ -27,626 +27,597 @@ const PROFILE_BANANA_TYPE: string = "banana";
...
-        .loadProfiles({_: undefined, $0: undefined});
-    } catch (e) {
-      error = e;
-    }
-    expect(error).toBeDefined();
-    expect(error instanceof ImperativeError).toBe(true);
-    expect(error.message).toMatchSnapshot();
-  });
-
-  it("should percolate the load error to the caller", async () => {
```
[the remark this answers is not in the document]
t1m0thyj: The ESLint Jest plugin has a [no-identical-title](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-identical-title.md) lint rule that complains if multiple tests in the same suite have an identical title. That's how I found this, and performing a diff between these two tests showed they were exactly identical so I removed one of them: https://github.com/zowe/imperative/blob/next/packages/cmd/__tests__/profiles/CommandProfileLoader.test.ts#L232-L288

awharn (approved):
These changes look good to me. 😄

PR #662, opened 2021-09-01
[next] Don't prompt for token value in "config secure"

t1m0thyj:
Instead of prompting users directly for token value, log in to an authentication service like API ML to obtain the token.

This implements the first part of https://github.com/zowe/zowe-cli/issues/923:
> 1. Standardize on `zowe config secure`
>     > prompt for a username/password combo to get the token

### DIFF
--- a/packages/imperative/src/config/cmd/secure/secure.handler.ts
+++ b/packages/imperative/src/config/cmd/secure/secure.handler.ts
@@ -10,11 +10,16 @@
 */
 
 import { ICommandHandler, IHandlerParameters } from "../../../../../cmd";
+import { ICommandHandlerRequire } from "../../../../../cmd/src/doc/handler/ICommandHandlerRequire";
+import { ICommandProfileAuthConfig } from "../../../../../cmd/src/doc/profiles/definition/ICommandProfileAuthConfig";
+import { Config } from "../../../../../config";
 import { secureSaveError } from "../../../../../config/src/ConfigUtils";
+import { ConnectionPropsForSessCfg, ISession, Session } from "../../../../../rest";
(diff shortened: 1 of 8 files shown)


packages/imperative/src/config/cmd/secure/secure.handler.ts
```diff
@@ -50,7 +69,69 @@ export default class SecureHandler implements ICommandHandler {
...
+        }
+
+        const authConfigs: ICommandProfileAuthConfig[] = [];
+        ImperativeConfig.instance.loadedConfig.profiles.forEach((profCfg) => {
+            if (profCfg.type === profileType && profCfg.authConfig != null) {
+                authConfigs.push(...profCfg.authConfig);
+            }
+        });
+
+        for (const authConfig of authConfigs) {
```
gejohnston, line 109: Just a question: Is the for loop used to handle the case where multiple APIML servers might be configured?

gejohnston (approved):
Looks good.

packages/imperative/src/config/cmd/secure/secure.handler.ts
t1m0thyj: This implementation supports only one APIML login per team config file. So if a file contains multiple `tokenValue` properties associated with profiles that have `tokenType=apimlAuthenticationToken`, only the first occurrence would attempt to log in to APIML.

Imperative supports multiple auth configs which is the reason for this for loop, although Zowe CLI has only one auth config for base profiles.

zFernand0 (approved):
LGTM!!! 🐇

awharn (approved):

PR #664, opened 2021-09-08
Update tar in package lock

awharn:
TAR 6.1.8 has vulnerabilities that are detected by npm audit. Update package-lock to version 6.1.11

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18316,9 +18316,9 @@
       }
     },
     "node_modules/tar": {
-      "version": "6.1.8",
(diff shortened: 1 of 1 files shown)


gejohnston (approved):

zFernand0 (approved):

awharn:
Test failures on Node 16 seem to be because 16.9.0 changed how the user is warned when referencing a property of an undefined object. Updates will be needed, and some snapshots will need to be removed and replaced with regex matches. This, however, is not a result of the changes made in this PR, and I believe can be done at a later time.

PR #668, opened 2021-09-21
Add `zowe config` group examples in Help

namanpatel1112:
Resolves https://github.com/zowe/zowe-cli/issues/1061

### DIFF
--- a/packages/imperative/src/config/cmd/list/list.definition.ts
+++ b/packages/imperative/src/config/cmd/list/list.definition.ts
@@ -42,6 +42,14 @@ export const listDefinition: ICommandDefinition = {
     ],
     examples: [
         {
+            description: "List property names for a specified config property.",
+            options: `"default"`
+        },
+        {
+            description: "List only root level property names for a specified config property.",
+            options: `"default" --root`
+        },
+        {
             description: "List config properties by separating them by their respective config file locations.",
             options: "--locations"
         },
@@ -50,7 +58,7 @@ export const listDefinition: ICommandDefinition = {
             options: "--root"
         },
         {
-            description: "List only the root level configuration properties by separating them by their respective config file locations..",
+            description: "List only the root level configuration properties by separating them by their respective config file locations.",
             options: "--locations --root"
         }
     ]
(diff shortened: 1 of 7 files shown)


namanpatel1112:
I added examples for the `zowe config` group of commands. But, I cannot see the details about the failing jenkins check.

t1m0thyj:
> I added examples for the `zowe config` group of commands. But, I cannot see the details about the failing jenkins check.

The changelog check is failing. Not sure why the GH workflow didn't run though, it should also have detected the missing changelog 🤔

awharn:
> 
> 
> > I added examples for the `zowe config` group of commands. But, I cannot see the details about the failing jenkins check.
> 
> The changelog check is failing. Not sure why the GH workflow didn't run though, it should also have detected the missing changelog 🤔

It doesn't seem to be configured to run against PRs that target `next`, just `master`.

awharn (approved):

packages/imperative/src/config/cmd/list/list.definition.ts
```diff
@@ -39,5 +39,27 @@ export const listDefinition: ICommandDefinition = {
                 "For example, specify in addition to '--locations' to get a list of config file paths only.",
             type: "boolean"
         }
+    ],
+    examples: [
+        {
+            description: "List property names for a specified config property.",
+            options: `"default"`
```
t1m0thyj, line 46: Should this be plural "defaults" to match the property name in zowe.config.json?

packages/imperative/src/config/cmd/list/list.definition.ts
```diff
@@ -39,5 +39,27 @@ export const listDefinition: ICommandDefinition = {
...
+        },
+        {
+            description: "List only root level property names for a specified config property.",
+            options: `"default" --root`
```
t1m0thyj, line 50: Same question as above, should the property name be plural?

packages/imperative/src/config/cmd/list/list.definition.ts
namanpatel1112: Right, I'll make it plural.

t1m0thyj (approved):

PR #671, opened 2021-09-27
[next] Introduce Config Update-Schemas Command

zFernand0:
Fixes https://github.com/zowe/zowe-cli/issues/1059

⚠️ This PR introduces **Next-only** breaking changes :warning:

Example output:
image


package.json
```diff
@@ -57,6 +57,7 @@
     "comment-json": "4.1.0",
     "dataobject-parser": "1.2.1",
     "deepmerge": "4.2.2",
+    "fast-glob": "^3.2.7",
```
zFernand0, line 60: Introduced a new dependency for performance reasons when `depth` is specified.
For more information on the other methods we tried, see the comments in the `packages/config/src/ConfigSchema.ts` file

packages/config/src/Config.ts
```diff
@@ -9,7 +9,7 @@
 *
 */
 
-import * as node_path from "path";
+import * as path from "path";
 import * as os from "os";
 import * as fs from "fs";
```
zFernand0, line 14: I believe I changed this to be able to have a line that's is almost 150 characters long 😋 
I can revert it back (and maybe change all other packages we import from node to `node_<package>` (e.g. `node_os`, `node_fs`, ...)

packages/config/src/Config.ts
```diff
@@ -319,6 +319,29 @@ export class Config {
...
+        const layer = this.layerActive();
+        const schemaUri = layer.properties.$schema ?? null;
+        if (schemaUri == null) {
+            return null;
+        }
+
+        const schemaFilePath = path.resolve(
+            schemaUri.startsWith("file://") ? fileURLToPath(schemaUri) :
+                schemaUri.startsWith("./") ? path.join(path.dirname(layer.path), schemaUri) : schemaUri);
+        const isSchemaLocal = fs.existsSync(schemaFilePath) || path.isAbsolute(schemaUri);
```
zFernand0, line 336: Checking to see if the path.isAbsolute may be redundant. 
Will find out when writing unit tests 😋

packages/config/src/ConfigSchema.ts
```diff
@@ -28,13 +35,30 @@ export class ConfigSchema {
...
+        $schema: "URL",
+        $version: "Version",
+        properties: {
+            defaults: "Default Definitions",
+            explainedParentKey: "Properties",
+            ignoredKeys: null
+        },
+        explainedParentKey: "Schema",
+        ignoredKeys: null
+    };
```
zFernand0, line 53: This may not be super necessary.
Mostly for a nice `<app>.log` output. 
Another method would be to use `util.inspect`

packages/config/src/ConfigSchema.ts
```diff
@@ -175,17 +199,140 @@ export class ConfigSchema {
 
     /**
      * Loads Imperative profile schema objects from a schema JSON file.
-     * @param schemaJson The schema JSON for config
+     * @param schema The schema JSON for config
      */
-    public static loadProfileSchemas(schemaJson: IConfigSchema): IProfileTypeConfiguration[] {
-        const patternName = Object.keys(schemaJson.properties.profiles.patternProperties)[0];
+    public static loadSchema(schema: IConfigSchema): IProfileTypeConfiguration[] {
```
zFernand0, line 204: This is the `next-only` breaking change mentioned in the PR comments.
Not really needed, but I thought it was nice to be consistent with the API names 😋

packages/imperative/src/config/cmd/auto-init/handlers/BaseAutoInitHandler.ts
```diff
@@ -153,17 +153,15 @@ export abstract class BaseAutoInitHandler implements ICommandHandler {
...
                 await ImperativeConfig.instance.config.save(false);
             }
         } else {
             // Merge, generate schema, and save
             ImperativeConfig.instance.config.api.layers.merge(profileConfig);
             if (ImperativeConfig.instance.config.api.layers.get().properties.$schema == null) {
                 // TODO What condition should we use to decide whether to (re)generate schema?
-                const schema = ConfigSchema.buildSchema(ImperativeConfig.instance.loadedConfig.profiles);
-                ImperativeConfig.instance.config.setSchema(schema);
+                ConfigSchema.updateSchema();
```
zFernand0, line 164: After creating the `ConfigSchema.updateSchema` API, all occurrences of `buildSchema + setSchema` were updated to use it. 😋

packages/imperative/src/plugins/utilities/npm-interface/install.ts
```diff
@@ -124,6 +126,11 @@ export async function install(packageLocation: string, registry: string, install
             spaces: 2
         });
 
+        iConsole.debug(`Checking for global team configuration files to update.`);
+        if (ImperativeConfig.instance.config.layers.filter((layer) => layer.global && layer.exists).length > 0) {
+            ConfigSchema.updateSchema({layer: "global"});
+        }
```
zFernand0, line 132: This is the first part of https://github.com/zowe/zowe-cli/issues/1059#issuecomment-909466415. :yum:

zFernand0 (commented):
Left a few comments that may help in reviewing the implementation. 😋 
Feel free to request any changes! 🚀

packages/imperative/src/config/cmd/update-schemas/update-schemas.definition.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { join } from "path";
+import { ICommandDefinition } from "../../../../../cmd";
+
+export const updateSchemasDefinition: ICommandDefinition = {
+    name: "update-schemas",
+    aliases: ["us"],
+    type: "command",
+    summary: "update schema files",
+    description: "Update schema files by looking up the directory structure.",
```
gejohnston, line 20: While we plan to automatically call this when installing a plugin, if we want customers to run this command directly, we may need a more complete description of what they would accomplish with this command. I don't know if we should try to explain some of the edge cases. It might be too wordy and confusing,

gejohnston (approved):
Looks good. I added a couple of suggestions. You should use your own judgement about what to actually do.

packages/imperative/src/config/cmd/update-schemas/update-schemas.definition.ts
```diff
@@ -0,0 +1,31 @@
...
+    handler: join(__dirname, "update-schemas.handler"),
+    positionals: [],
+    options: [
+        {
+            name: "depth",
+            description: "Specifies how many levels down the directory structure should the schemas be updated.",
+            type: "number",
+            defaultValue: 0
```
zFernand0, line 28: Should we restrict this option to only positive values, or should we allow negative numbers?
Quick explanation of what we **could** allow:
- Positive depth: Limits how far down the directory structure we go
- Negative depth: Limits how far up the directory structure we go

gejohnston (approved):
Looks good.

packages/imperative/src/config/cmd/update-schemas/update-schemas.handler.ts
```diff
@@ -0,0 +1,50 @@
...
+        const updatedPaths = ConfigSchema.updateSchema({ layer: "all", depth: params.arguments.depth });
+
+        // output to terminal
+        params.response.data.setObj(updatedPaths);
+        params.response.console.log("Configuration files found: " + Object.keys(updatedPaths).length.toString());
+        if (Object.keys(updatedPaths).length > 0) {
+            const outputObj: any = {};
+            for (const schema in updatedPaths) {
+                if (Object.prototype.hasOwnProperty.call(updatedPaths, schema)) {
+                    const schemaInfo = updatedPaths[schema];
```
t1m0thyj, line 38: Curious if this one-liner would work to do the same thing? 😛 
`for (const [schema, schemaInfo] of Object.entries(updatedPaths)) {`

packages/config/src/Config.ts
```diff
@@ -319,6 +320,32 @@ export class Config {
...
+            return {
+                original: null,
+                resolved: null,
+                local: false,
+            };
+        }
+
+        const schemaFilePath = path.resolve(
+            schemaUri.startsWith("file://") ? fileURLToPath(schemaUri) : // else if : )
+                schemaUri.startsWith("./") ? path.join(path.dirname(layer.path), schemaUri) : schemaUri);
```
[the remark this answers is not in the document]
t1m0thyj: This may already be addressed by https://github.com/zowe/zowe-cli/issues/1163 🙂

packages/config/src/Config.ts
```diff
@@ -434,16 +461,17 @@ export class Config {
...
-        layer.properties = { $schema: schemaUri, ...layer.properties };
+        if (layer.properties.$schema == null) {
+            layer.properties = JSONC.parse(JSONC.stringify({ ...{ $schema: schemaUri }, ...layer.properties }, null, ConfigConstants.INDENT));
+        }
 
-        if (schemaObj != null) {
-            const filePath = node_path.resolve(node_path.dirname(layer.path), schemaUri);
-            fs.writeFileSync(filePath, JSONC.stringify(schemaObj, null, ConfigConstants.INDENT));
+        const schemaInfo = this.getSchemaInfo();
+        if (schemaObj != null && (schemaInfo.local || schemaInfo.original.startsWith("./"))) {
```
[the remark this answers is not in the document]
t1m0thyj: I'm particularly curious about this, because the import handler uses different logic for determining whether a config is local:
`const isConfigLocal = fs.existsSync(configFilePath) || path.isAbsolute(params.arguments.location);`

In theory, the `path.isAbsolute` check should accurately determine whether or not a file is local, even if it doesn't exist.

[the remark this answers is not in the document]
zFernand0: > Is it because the schema may be local, even though it doesn't exist on disk yet?

That's was the reason.

> the `path.isAbsolute` check should accurately determine whether or not an absolute URL/path is local, even if it doesn't exist

That's right, however relative paths that do not exists will also return `local == false`
Then we could just resolve it prior to checking it, right? ✅ 
But what happens to URLs?
`path.isAbsolute(path.resolve("http:/localhost:12345/zowe.schema.json"))`
`path.isAbsolute("/root/gh/zowe/imperative/http:/localhost:12345/zowe.schema.json")`
Results: `true`

TL;DR: No good way to workaround it other than
```
let isUrl = true; 
try { new URL(params.arguments.location) } catch (_) { isUrl = false; }
```

packages/config/src/__mocks__/Config.ts
```diff
@@ -0,0 +1,57 @@
+/*
```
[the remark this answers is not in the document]
zFernand0: I can move them around 😋 
Should we also move `imperative/src/__mocks__` and `utilities/src/__mocks__` ?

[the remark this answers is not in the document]
zFernand0: IDK if this will be confusing to developers though...

```typescript
// jest mocks
jest.mock("../../../../utilities/ImperativeConfig");
jest.doMock("../../../../utilities/ImperativeConfig");

// requires
let {ImperativeConfig} = require("../../../../utilities/src/ImperativeConfig");
// imports
({ImperativeConfig} = await import("../../../../utilities/src/ImperativeConfig"));
```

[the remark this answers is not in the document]
t1m0thyj: It should be fine, I was concerned about having mocks inside the "src" folder but didn't realize we already do this elsewhere and exclude all "__mocks__" in our tsconfig.json 👍

packages/config/src/Config.ts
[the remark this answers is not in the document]
zFernand0: Handled URLs as mentioned in the comment above.
Q: Should the `isUrl` be a utility function on Imperative?

packages/config/__tests__/Config.api.test.ts
```diff
@@ -361,6 +361,82 @@ describe("Config API tests", () => {
...
+                expect(await validateExists(path.dirname(fakePath), undefined)).toBe(false);
+                expect(await validateExists(path.dirname(filePathProjectUserConfig), undefined)).toBe(false);
+                expect(await validateExists(path.dirname(filePathProjectConfig), undefined)).toBe(false);
+                expect(await validateExists(path.dirname(filePathAppUserConfig), undefined)).toBe(false);
+                expect(await validateExists(path.dirname(filePathAppConfig), undefined)).toBe(false);
+            });
+
+            it("should determine if a User Config layer exists", async () => {
+                expect(await validateExists(path.dirname(filePathProjectUserConfig), true)).toBe(true);
+                expect(await validateExists(path.dirname(filePathProjectConfig), true)).toBe(true);
```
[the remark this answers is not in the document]
zFernand0: I might change this test a little bit to simulate all the possible scenarios that layerExists should cover. 😋

packages/utilities/src/JsUtils.ts
```diff
@@ -20,14 +22,19 @@ export class JsUtils {
...
+
+    // __________________________________________________________________________
+    /**
+     * Is the supplied string a URL.
+     *
+     * @param {string} urlString - The string to test.
+     *
+     * @returns {boolean} - True if it is a URL. False otherwise.
+     */
+    public static isUrl(urlString: string): boolean {
```
t1m0thyj, line 36: After discussing this IRL, should `urlString` be wrapped in `path.posix.normalize`, so that drive letters in Windows paths don't get evaluated as protocols?

zFernand0: Checking for `new URL("C:\\Users...").origin !== "null"` seems to work 😋

packages/config/src/Config.ts
```diff
@@ -319,6 +321,30 @@ export class Config {
...
+                resolved: null,
+                local: false,
+            };
+        }
+
+        const tempSchema = originalSchema.startsWith("file://") ? fileURLToPath(originalSchema) : originalSchema;
+        const schemaFilePath = path.resolve( tempSchema.startsWith("./") ? path.join(path.dirname(layer.path), tempSchema) : tempSchema);
+        return {
+            original: originalSchema,
+            resolved: schemaFilePath,
```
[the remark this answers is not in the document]
zFernand0: Thanks.
Even though it's not used at the moment, it'd be good for it to be meaningful for programmatic purposes. :yum:

PR #672, opened 2021-09-29
[next] Change default log level to WARN

### LINKED ISSUE #634: [next] Consider changing default log level
Currently the default log level is INFO, which makes the log files really big, and can hurt performance because writing to the logs isn't buffered.

For Zowe v2 LTS, perhaps we should change the default log level to WARN? This should still show important error information.

t1m0thyj:
Resolves #634. This is a breaking change.

When the environment variable `NODE_ENV=development` is set:
* Restores the old logging behavior (default log level of DEBUG)
* Disables web help caching (so that it always gets regenerated)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Breaking: Changed default log level from DEBUG to WARN for Imperative logger and app logger to reduce the volume of logs written to disk. [#634](https://github.com/zowe/imperative/issues/634)
+
 ## `5.0.0-next.202109281439`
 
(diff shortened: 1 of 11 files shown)


zFernand0:
Really cool idea! 😋

gejohnston (approved):

zFernand0 (approved):
This is awesome!
Should we add something to the conformance criteria about this? (for plugin developers maybe)

t1m0thyj:
Are we going to leave `NODE_ENV=development` as an "Easter egg", or should we document it somewhere? I'm thinking probably the latter. Since the features are intended for developers only, maybe it would make sense to document them in the `docs` folder of CLI repo: https://github.com/zowe/zowe-cli/tree/next/docs

awharn (approved):

PR #677, opened 2021-10-11
Make profiles available right after installing the plugin

zFernand0:
Signed-off-by: zFernand0 <fernando.rijocedeno@broadcom.com>

### DIFF
--- a/__tests__/src/packages/imperative/plugins/suites/InstallingPlugins.ts
+++ b/__tests__/src/packages/imperative/plugins/suites/InstallingPlugins.ts
@@ -15,7 +15,7 @@
 
 import * as fs from "fs";
 import * as T from "../../../../../src/TestUtil";
-import { join } from "path";
+import { join, resolve } from "path";
 import { TEST_REGISTRY } from "../../../../../__src__/TestConstants";
 import { execSync, SpawnSyncReturns } from "child_process";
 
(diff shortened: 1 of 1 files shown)


gejohnston (approved):
You had to add a lot of lower-level function changes to get enough information to do the one extra step that you wanted during plugin-install. Nice job working through all of that.

t1m0thyj (approved):
LGTM! Clever way to eliminate a `.bind` 👍 

I'm ok with the breaking change to the public method `PluginManagementFacility.requirePluginModuleCallback` since this PR is targeting the "next" branch 🙂

Since this is a breaking change, should it be mentioned in the changelog?

packages/config/src/Config.ts
```diff
@@ -294,31 +296,31 @@ export class Config {
      * App name used in config filenames (e.g., *my_cli*.config.json)
      */
     public get appName(): string {
-        return this.mApp;
+        return path.isAbsolute(this.mApp) ? ImperativeConfig.instance.loadedConfig?.name ?? path.basename(this.mApp) : this.mApp;
```
zFernand0, line 299: Not sure if this is a real problem or it's just something that will only happen during integration testing.

When calling the a CLI using a method like `node --require ts-node/register <absolute-path-to-CLI.ts> config init`, some Config APIs may have issues saving the config and schema files.

I'm more than happy to revert this change back if we think that it's something people shouldn't be doing. 😋  🦆

packages/config/src/Config.ts
```diff
@@ -294,31 +296,31 @@ export class Config {
...
+        return `${this.appName}${Config.END_OF_USER_CONFIG}`;
     }
 
     // _______________________________________________________________________
     /**
      * Filename used for config schema JSON files
      */
     public get schemaName(): string {
-        return `${this.mApp}.schema.json`;
+        return `${this.appName}.schema.json`;
```
zFernand0, line 323: Just changed to use the enhanced getter

zFernand0:
Will wait for #674 to be merged 😋

PR #680, opened 2021-10-25
Add `--overwrite` option for `config init` command

namanpatel1112:
Resolves https://github.com/zowe/zowe-cli/issues/1036


zFernand0:
Some of these changes could be included in the #685 issue 😋

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
```diff
@@ -145,6 +145,64 @@ describe("Configuration Initialization command handler", () => {
...
+        expect(setSchemaSpy).toHaveBeenCalledWith(expectedSchemaObjectNoBase);
+
+        expect(promptWithTimeoutSpy).toHaveBeenCalledTimes(1);
+        // Prompting for secure property
+        expect(promptWithTimeoutSpy).toHaveBeenCalledWith(expect.stringContaining("blank to skip:"), {"hideText": true});
+
+        expect(writeFileSyncSpy).toHaveBeenCalledTimes(2);
+        // Schema
+        expect(writeFileSyncSpy).toHaveBeenNthCalledWith(1, fakeSchemaPath, JSON.stringify(expectedSchemaObjectNoBase, null, ConfigConstants.INDENT));
+        // expect(writeFileSyncSpy).toHaveBeenNthCalledWith(2, fakeProjPath, JSON.stringify(compObj, null, ConfigConstants.INDENT)); // Config
```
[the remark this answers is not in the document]
t1m0thyj: @namanpatel1112 The only difference I see between the expected and received object which is causing the test fail, is that the received object is missing the `$schema` property.

I believe the problem lies in the `initWithSchema` method. On line 78, the `$schema` property is added to the config JSON. Then on line 91, the config JSON is replaced with a new object that is missing the `$schema` property.
https://github.com/zowe/imperative/blob/05a05fdfac602175c5329b333368f68cb857b213/packages/imperative/src/config/cmd/init/init.handler.ts#L76-L93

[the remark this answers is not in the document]
namanpatel1112: The tests passed. I didn't have to swap the order of the writeFileSync expects. If I swapped them, the tests were failing. 🤔

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
```diff
@@ -23,6 +23,7 @@ import * as lodash from "lodash";
 import * as fs from "fs";
 import * as os from "os";
 import { CredentialManagerFactory } from "../../../../../security";
+import { param } from "jquery";
```
awharn, line 26: Could you please remove this unused import?

packages/imperative/src/config/cmd/init/init.handler.ts
```diff
@@ -42,14 +42,14 @@ export default class InitHandler implements ICommandHandler {
...
-        await this.initWithSchema(config, params.arguments.userConfig);
+        await this.initWithSchema(config, params.arguments.userConfig, params.arguments.overwrite);
 
         if (params.arguments.prompt !== false && !CredentialManagerFactory.initialized && config.api.secure.secureFields().length > 0) {
             const warning = secureSaveError();
             params.response.console.log(TextUtils.chalk.yellow("Warning:\n") +
                 `${warning.message} Skipped prompting for credentials.\n\n${warning.additionalDetails}\n`);
         }
-
+
```
awharn, line 52: Can you please remove the trailing space on this line?

packages/imperative/src/config/cmd/init/init.definition.ts
```diff
@@ -46,20 +46,35 @@ export const initDefinition: ICommandDefinition = {
             description: "Prompt for secure values. Useful for disabling prompting in CI environments.",
             type: "boolean",
             defaultValue: true
+        },
+        {
```
awharn, line 50: I believe that we historically have the "ow" alias for overwrite operations. Would it make sense to add that alias here?

packages/imperative/src/config/cmd/init/init.definition.ts
```diff
@@ -46,20 +46,35 @@ export const initDefinition: ICommandDefinition = {
...
+            name: "overwrite",
+            description: "Replace existing config files instead of merging the new changes.",
+            type: "boolean",
+            defaultValue: false,
+            conflictsWith: ["dry-run"]
```
awharn, line 55: With overwrite commands, we often have a [confirmation option](https://github.com/zowe/imperative/blob/next/packages/imperative/src/config/cmd/auto-init/AutoInitConstants.ts#L81) to prevent people from invoking the option accidentally. In most places, this option is called "for-sure", with the alias "fs" for short. Would it make sense to add to this command as well? It could be potentially time consuming to create a configuration file again if overwrite is used.

awharn (commented):
I have a couple of suggested changes and questions.

zFernand0:
I would like to suggest one more thing.

Could we group the `--overwrite` tests (and probably the `--dry-run` tests as well) in their own `describe( )`?

The reason for asking for this is to make the `diff` a little easier to read through when reviewing the PR.

That said, I don't believe having a separate `describe( )` for each new test we add is necessarily the right thing to do either.

Personally, I don't feel too strongly as we could use tools other than `git diff` to review.

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
namanpatel1112: Removed it

packages/imperative/src/config/cmd/init/init.definition.ts
namanpatel1112: Good point, I added a for-sure flag to go with it. :)

namanpatel1112:
> I would like to suggest one more thing.
> 
> Could we group the `--overwrite` tests (and probably the `--dry-run` tests as well) in their own `describe( )`?
> 
> The reason for asking for this is to make the `diff` a little easier to read through when reviewing the PR.
> 
> That said, I don't believe having a separate `describe( )` for each new test we add is necessarily the right thing to do either.
> 
> Personally, I don't feel too strongly as we could use tools other than `git diff` to review.

So, should I leave them as they are right now? I found this diff checker: https://www.diffchecker.com/
But I am not sure if you had other tool(s) in mind.

packages/imperative/src/config/cmd/init/init.definition.ts
```diff
@@ -76,6 +92,14 @@ export const initDefinition: ICommandDefinition = {
         {
             description: "Do a dry run of initializing the user config files and do not prompt for secure values.",
             options: "--user-config --prompt false --dry-run"
+        },
+        {
+            description: "Overwrite any existing global config files",
+            options: "--global-config --overwrite"
```
awharn, line 98: Should this, and the example below, also include the --for-sure flag?

awharn (commented):
One more small comment. Once this is fixed, this PR has my approval. 😃

packages/imperative/src/config/cmd/init/init.definition.ts
namanpatel1112: Yes, I added it for both examples. Thanks for pointing it out.

awharn (approved):

awharn (commented):
I just approved, but noticed that a lot of tests started failing because it states dry-run and overwrite are both specified. It looks like because both of those mutually exclusive options have default values, it believes they are both specified. It might be best to remove the default values on overwrite and dry-run.

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
```diff
@@ -634,7 +944,6 @@ describe("Configuration Initialization command handler", () => {
         (params.response.console as any).prompt = promptWithTimeoutSpy;
         writeFileSyncSpy.mockImplementation(); // Don't actually write files
 
-        jest.spyOn(process, "cwd").mockReturnValueOnce(__dirname);
```
t1m0thyj, line 637: Curious about the reason for this change? I think the `jest.resetAllMocks` call in the beforeEach block should mean no side effects of mocks are left behind for subsequent tests.

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
```diff
@@ -683,7 +992,6 @@ describe("Configuration Initialization command handler", () => {
         // initForDryRun
         const initForDryRunSpy = jest.spyOn(handler as any, "initForDryRun");
 
-        jest.spyOn(process, "cwd").mockReturnValueOnce(__dirname);
```
t1m0thyj, line 686: Again, curious about this change? 🙂

packages/imperative/src/config/cmd/init/init.handler.ts
```diff
@@ -120,19 +121,23 @@ export default class InitHandler implements ICommandHandler {
      * @param config Config object to be populated
      * @param user If true, properties will be left empty for user config
      */
-    private async initWithSchema(config: Config, user: boolean): Promise<void> {
-        // Build the schema and write it to disk
-        ConfigSchema.updateSchema();
-
+    private async initWithSchema(config: Config, user: boolean, overwrite: boolean, forSure: boolean): Promise<void> {
```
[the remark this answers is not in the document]
namanpatel1112: I made the requested change.

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
namanpatel1112: Good catch, this wasn't supposed to be removed. I added it back.

packages/imperative/__tests__/config/cmd/init/init.handler.test.ts
namanpatel1112: Again, I removed it unintentionally. I added it back.

t1m0thyj (approved):
LGTM!

zFernand0 (approved):

PR #688, opened 2021-11-04
Move config load into try/catch block when --dcd specified

awharn:
Fixes Daemon error when an incorrect zowe.config.json file is supplied.
Updates Config.load API to have a `@throws` in the JS Doc.

Fixes zowe/zowe-cli#943
Fixes zowe/zowe-cli#1190

### DIFF
--- a/packages/imperative/__tests__/Imperative.test.ts
+++ b/packages/imperative/__tests__/Imperative.test.ts
@@ -235,7 +235,7 @@ describe("Imperative", () => {
                 expect(ConfigManagementFacility.instance.init).toHaveBeenCalledTimes(1);
             });
 
-            it("should surface failures if daemonMode is specified", async () => {
+            it("should surface failures if daemonMode is not specified", async () => {
                 jest.spyOn(mocks.Config, "load").mockRejectedValueOnce(new ImperativeError({msg: "Config error"})).mockResolvedValue({});
                 let error;
                 try {
--- a/packages/imperative/__tests__/Imperative.test.ts
+++ b/packages/imperative/__tests__/Imperative.test.ts
@@ -18,6 +18,7 @@ import { IConfigLogging } from "../../logger";
 import { IImperativeEnvironmentalVariableSettings } from "..";
 import { ICommandDefinition } from "../../cmd/src/doc/ICommandDefinition";
 import * as yargs from "yargs";
+import { ImperativeError } from "../../error/src/ImperativeError";
 
 describe("Imperative", () => {
     const mainModule = process.mainModule;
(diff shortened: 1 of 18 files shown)


gejohnston (approved):

packages/imperative/src/ConfigurationLoader.ts
```diff
@@ -55,7 +55,9 @@ export class ConfigurationLoader {
         // override the config with the content of the module specified
         if (config.configurationModule != null) {
             try {
+                const daemonMode = config.daemonMode;
                 config = callerFileRequirer(config.configurationModule);
+                config.daemonMode = daemonMode;
```
dkelosky, line 60: it looks like `daemonMode` piece is intentionally not overridden - is that correct?

packages/imperative/src/doc/IImperativeConfig.ts
```diff
@@ -326,4 +326,12 @@ export interface IImperativeConfig {
...
      */
     apimlConnLookup?: IApimlSvcAttrs[];
+
+    /**
+     * If Imperative should run in Daemon mode
+     * This should only be specified for CLIs
+     * @type {boolean}
+     * @memberof IImperativeConfig
+     */
+    daemonMode?: boolean;
```
dkelosky, line 336: the description of this is that it "should" run in daemon mode, but presumably it is possible to not run in daemon mode.  is that correct?  will zowe cli set this in a PR for that repo?

packages/imperative/src/ConfigurationLoader.ts
awharn: That is correct. We decided we wanted to keep daemonMode because it is not always in use, which means we couldn't set it to be true in the Imperative configuration in the CLI. The CLI, at startup, will search for `--daemon` in its list of arguments, and if the argument is present, daemonMode is set to true. This will suppress any Config related errors (bad secure storage mechanism, invalid JSON, etc) at server startup time only.

packages/imperative/src/doc/IImperativeConfig.ts
awharn: As stated in my previous comment, this option allows Imperative to initialize while suppressing configuration errors in Daemon mode. It should not be true when daemon mode is not running, because the configuration is never reloaded outside of when the --dcd option is specified (which is only true in Daemon mode).

dkelosky: ok thanks, if `--daemon` is found at startup, `daemonMode` will be set to `true` - is that the correct summary?

awharn: That is correct.

packages/imperative/__tests__/Imperative.test.ts
```diff
@@ -233,6 +234,35 @@ describe("Imperative", () => {
 
                 expect(ConfigManagementFacility.instance.init).toHaveBeenCalledTimes(1);
             });
+
+            it("should surface failures if daemonMode is specified", async () => {
```
t1m0thyj, line 238: Should this say `if daemonMode is not specified`?

zFernand0 (approved):
This PR LGTM! 😋 
Thanks @awharn for clarifying a few things IRL

packages/imperative/__tests__/Imperative.test.ts
awharn: Yeah...

awharn: Fixed.

t1m0thyj (approved):

gejohnston:
My attempts to run the zowe executable still does not report a syntax error when the daemon must be started on windows.

Maybe I did not rebuild correctly. Maybe you are not done with changes yet.

Confirm the behavior before merging as follows:

-  Download the current zowe executable
    - Windows: https://github.com/zowe/zowe-cli/suites/4278364423/artifacts/111371871
    - Linux: https://github.com/zowe/zowe-cli/suites/4278364423/artifacts/111371869

- Put a syntax error in zowe.config.json

- Use that execuable in a command like this (which shows the fix not working):

```
.\zowe.exe --version
Starting a background process to increase performance ...
The Zowe daemon that we started is not running on host = 127.0.0.1 with port = 4000.
Command used to start the Zowe daemon was:
    C:\Program Files\nodejs\zowe.cmd --daemon
Terminating.
```

dkelosky (approved):

PR #690, opened 2021-11-11
[next] Make credentials secure by default for v1 profiles

t1m0thyj:
Resolves zowe/zowe-cli#1128 by changing the default CredentialManager setting in `~/.zowe/settings/imperative.json` to the package name of the host CLI (`@zowe/cli`). Supported values for CredentialManager are:
* `CredentialManager = false` - Store credentials in plain text. This was the default setting in Zowe v1 LTS and is preserved for backwards compatibility. This setting is ignored for v2 profiles which are always secure.
* `CredentialManager = '@zowe/cli'` - Store credentials using built-in credential manager. This PR makes this the default setting in Zowe v2 LTS. In v1 profiles, values of secure fields will show up as `managed by Zowe CLI`.
* `CredentialManager = 'XXX'` - Store credentials using a custom credential manager. This was only supported for v1 profiles previously, but is also supported for v2 profiles now.
---
**TODO: Test cases**
* [x] Clean install of vNext, store credentials in v1 profile
* [x] Clean install of vNext, store credentials in v2 profile
* [x] Zowe v1 with CredentialManager false, update to vNext, load/store credentials in v1 profile
* [x] Zowe v1 with CredentialManager SCS, update to vNext, load/store credentials in v1 profile
* [x] Zowe v1 with CredentialManager SCS, update to vNext, uninstall SCS [1], load/store credentials in v1 profile
* [x] Zowe v1 with CredentialManager false, update to vNext, store credentials in v2 profile
* [x] Zowe v1 with CredentialManager SCS, update to vNext, store credentials in v2 profile
* [x] Zowe v1 with CredentialManager SCS, update to vNext, uninstall SCS [1], store credentials in v2 profile
* [x] Clean install of vNext in environment that does not support Keytar, fail to store secure credentials in v1 profile
* [x] In same environment, remove CredentialManager setting, store plain text credentials in v1 profile
* [x] Clean install of vNext in environment that does not support Keytar, fail to store secure credentials in v2 profile
* [x] In same environment, remove CredentialManager setting, fail to store secure credentials in v2 profile
* [x] Clean install of vNext in environment that does not support Keytar, bypass secure credentials in v2 profile with env vars
* [x] Repeat all of the above tests using daemon mode (will do after PR has been reviewed 😋)
  * All cases pass except for https://github.com/zowe/zowe-cli/issues/1215
* [x] Zowe v1 with CredentialManager SCS, update to vNext w/ daemon, switch from folder w/o team config to folder with it
  * Ensure no regression of https://github.com/zowe/zowe-cli/issues/984

[1] Per [Zowe docs](https://docs.zowe.org/stable/user-guide/cli-install-cli-next/), uninstall SCS plug-in, and delete ~/.zowe/settings/imperative.json where it is set as the CredentialManager. The planned migration utility should eventually automate this process.

### DIFF
--- a/packages/imperative/__tests__/OverridesLoader.test.ts
+++ b/packages/imperative/__tests__/OverridesLoader.test.ts
@@ -16,14 +16,19 @@ jest.mock("../../utilities/src/ImperativeConfig");
 
 import { OverridesLoader } from "../src/OverridesLoader";
 import { CredentialManagerFactory, AbstractCredentialManager } from "../../security";
-
 import * as path from "path";
-import { ImperativeConfig } from "../..";
+import { ImperativeConfig, Logger } from "../..";
+import { AppSettings } from "../../settings";
 
 const TEST_MANAGER_NAME = "test manager";
 
 describe("OverridesLoader", () => {
     const mainModule = process.mainModule;
+    const mockCredMgrInitialized = jest.fn().mockReturnValue(false);
+
+    beforeAll(() => {
+        Object.defineProperty(CredentialManagerFactory, "initialized", { get: mockCredMgrInitialized });
+    });
 
     beforeEach(() => {
         jest.restoreAllMocks();
@@ -57,7 +62,7 @@ describe("OverridesLoader", () => {
             expect(CredentialManagerFactory.initialize).toHaveBeenCalledTimes(0);
         });
 
-        it("should not set a credential manager if there are no overrides and keytar is present in dependencies", async () => {
+        it("should not set a credential manager for CLI if there are no overrides and keytar is present in dependencies", async () => {
             const config: IImperativeConfig = {
                 name: "ABCD",
                 overrides: {},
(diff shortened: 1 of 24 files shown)


awharn (approved):
This looks good to me. 👍

packages/imperative/__tests__/OverridesLoader.test.ts
```diff
@@ -57,7 +62,58 @@ describe("OverridesLoader", () => {
...
+                name: "host-package",
+                dependencies: {
+                    keytar: "1.0"
+                }
+            };
+
+            jest.spyOn(AppSettings, "initialized", "get").mockReturnValue(true);
+            jest.spyOn(AppSettings, "instance", "get").mockReturnValue({
+                getNamespace: jest.fn()
+            } as any);
```
zFernand0, line 83: Will this cause problems to new tests added that depend on the `AppSettings.instance` if this mocked implementation isn't restored after the test ?

zFernand0 (approved):
LGTM! 😋

gejohnston (approved):
Looks good.

packages/imperative/__tests__/OverridesLoader.test.ts
t1m0thyj: I believe the `jest.restoreAllMocks` in the afterAll block should handle this, by resetting all `spyOn` mocks including this one to their original implementation.

t1m0thyj:
**New Changes** (to fix the test cases)
* Added `ConfigSecure.loadFailed`, which command handlers can use to check if secure credential storage is working. Previously they were checking `CredentialManagerFactory.initialized` which could be true even if secure credentials failed to load.
* Added `Config.reload`, which is called by CommandProcessor in daemon mode to reload config layers for the current project directory. Previously it was reinitializing config entirely which could fail if the CredentialManager failed to load.

gejohnston (approved):
Looks good.

zFernand0 (approved):
LGTM! 🦆 
Seems like we got a good starting point for the `zowe daemon reload` command with `Config.reload` 😋

awharn (approved):

PR #691, opened 2021-11-11
Next mask logs

zFernand0:
Second half of https://github.com/zowe/zowe-cli/issues/1106

This PR includes:
- Environmental Variable to control whether console output get's masked or not
- Deprecates the Imperative function: `envVariablePrefix`
- Masks sensitive data on `imperative.log` and `<app>.log` for levels above `TRACE`
- Masks console output (by default) with the exception of `user`, `password`, `tokenValue`, and `keyPassphrase`.
  - Those values are assumed to be masked already (when needed)
  - We wanted to prevent exposing possible sensitive data by excluding said fields

To Do:
- [x] Fix existing tests
- [x] Add new unit tests
- [x] Add new integration tests

---
UPDATE:
List of integration tests to add:
- [x] `imperative.log` and `zowe.log` should mask sensitive content regardless of the ENV variable for log levels `>= DEBUG`
  - ALL
  - TRACE
  - DEBUG
  - INFO
  - WARN
  - ERROR
  - FATAL
  - MARK (Do we need this - 6b58845d2461f2b3013de209130e1a3027ad1390)
  - OFF
    Full log level list: https://github.com/zowe/imperative/blob/master/packages/logger/src/Logger.ts#L32
- [x] Console output should be masked only if the ENV (`<APP>_APP_MASK_OUTPUT`) variable is `!== FALSE`
- [x] `--rfj` console output should be masked if the ENV variable is `!== FALSE`
- [x] Fields designated as `secure` in the schema object will be ignored
  - [x] Schema objects from plugins should also be excluded
  - [x] Plugins can have multiple `optionDefinitions` on a single secure profile property
- [x] Nested secure fields should follow the same rules as above
- [x] ~V1 profiles should follow the same rules as above~


packages/imperative/src/Imperative.ts
```diff
@@ -14,51 +14,54 @@
...
+import { ICommandDefinition } from "../../cmd/src/doc/ICommandDefinition";
+import { ICommandProfileTypeConfiguration } from "../../cmd/src/doc/profiles/definition/ICommandProfileTypeConfiguration";
+import { ICommandResponseParms } from "../../cmd/src/doc/response/parms/ICommandResponseParms";
+import { IHelpGenerator } from "../../cmd/src/help/doc/IHelpGenerator";
+import { IHelpGeneratorFactory } from "../../cmd/src/help/doc/IHelpGeneratorFactory";
+import { IHelpGeneratorParms } from "../../cmd/src/help/doc/IHelpGeneratorParms";
+import { IYargsResponse } from "../../cmd/src/yargs/doc/IYargsResponse";
+import { WebHelpManager } from "../../cmd/src/help/WebHelpManager";
+import { YargsConfigurer } from "../../cmd/src/yargs/YargsConfigurer";
+import { YargsDefiner } from "../../cmd/src/yargs/YargsDefiner";
```
zFernand0, line 47: Please excuse my import fixing spree.
I was trying to debug a circular dependency.
This lead me to deprecate an Imperative function (`envVariablePrefix`)

packages/logger/__tests__/LoggerUtils.test.ts
```diff
@@ -49,7 +51,88 @@ describe("LoggerUtils tests", () => {
...
+                    api: {
+                        layers: { get: layersGet },
+                        secure: { secureFields }
+                    }
+                }
+            };
+        });
+
+        describe("should NOT censor", () => {
+            it("data if we are using old profiles", () => {
```
gejohnston, line 83: Do we not mask because there is no "secure" array in old profiles? There was some discussion about using meta data from the profiles to determine if a property in the profile definition was defined as secure. Do we not bother with that meta data for old profiles because it would only be a half-way solution?

gejohnston (approved):
No requested changes. Just some questions to help my understanding.

gejohnston (changes requested):
Oops. One requested change. The typo in the CHANGELOG should be fixed since customers read that.

packages/imperative/__tests__/Imperative.test.ts
```diff
@@ -34,7 +34,7 @@ describe("Imperative", () => {
             jest.doMock("../src/ConfigurationLoader");
             jest.doMock("../src/ConfigurationValidator");
             jest.doMock("../src/help/ImperativeHelpGeneratorFactory");
-            jest.doMock("../../utilities/src/ImperativeConfig");
+            jest.doMock("../../utilities/src/ImperativeConfig", ()=> require("../../utilities/src/__mocks__/ImperativeConfig"));
```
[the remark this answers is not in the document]
zFernand0: There is nothing preventing us from using the recommended pattern (in fact, there is an ESLint rule for that).
This was done while trying to debug some circular dependencies.
Will revert in the next commit. Thank you!

packages/imperative/__tests__/Imperative.test.ts
```diff
@@ -356,6 +357,10 @@ describe("Imperative", () => {
                     envConfig = mocks.EnvironmentalVariableSettings.read(Imperative.envVariablePrefix);
                 });
 
+                it("should know what the envVariablePrefix is [DEPRECATED]", async () => {
```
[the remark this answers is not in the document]
zFernand0: Besides from specifying the `envVariablePrefix` as part of the application's configuration object (in `imperative.ts`), I don't think we allowed another way to set this property since there was no `setter` defined for it.
Any existing non-standard workarounds they may be doing today to redefine it, e.g. `(Imperative as any).envVariablePrefix = xyz`, should still work even after the deprecation (and even after the removal).

packages/logger/src/LoggerUtils.ts
```diff
@@ -112,14 +113,19 @@ export class LoggerUtils {
...
-        // - meta (imperative config object from plugins)
-        // - schema.json
+        let specialValues = ["user", "password", "tokenValue", "keyPassphrase"];
 
-        for (const profile of LoggerUtils.profiles) {
+        const getPropertyNames = (prop: ICommandProfileProperty): string[] => {
+            const ret: string[] = [];
+            ret.push(prop.optionDefinition?.name);
+            prop.optionDefinitions?.map(opDef => ret.push(opDef.name));
+            return ret;
```
gejohnston, line 122: Can ret be an empty array during this return statement?
Do you expect that an empty array will be false, since the object will exist even if it has no elements?
Or, is the intention that this return statement should always be true?

gejohnston (commented):
The return statement that I commented on might be correct. I am just not sure what results you intended to achieve.

packages/logger/src/LoggerUtils.ts
zFernand0: I should probably add some JSDoc for it, but the intention is for it to return a combined list of options names for a given `ICommandProfileProperty`.
The expected outcome is an array of strings which results in either `["optionName"]`, or `[undefined, "optionA", "optionB", ...]` which will later get unified (`lodash.union`) with the `specialValues` array.

__tests__/config/integration/jest.config.json
```diff
@@ -24,5 +24,5 @@
         "unit",
         "__system__"
     ],
-    "testRegex": "(test|spec)\\.(ts)"
+    "testRegex": "__integration__.*(test|spec)\\.(ts)"
```
t1m0thyj, line 27: Is this one-line change all that was needed to stop running our unit tests twice? 😋

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added an environment variable to control whether or not sensitive data will be masked in the console output.<br/>
+    This behavior excludes any TRACE lavel logs for both, Imperative.log and AppName.log.<br/>
```
t1m0thyj, line 8: "lavel" -> "level"

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
```
t1m0thyj, line 5: Should we also document the next breaking change for `envVariablePrefix`?

packages/logger/__tests__/LoggerUtils.test.ts
```diff
@@ -49,7 +51,88 @@ describe("LoggerUtils tests", () => {
     });
 
     it("Should not hide notSecret operand", () => {
-        const data = LoggerUtils.censorYargsArguments({_: [], $0: "test", notSecret: "canSeeMe"});
+        const data = LoggerUtils.censorYargsArguments({ _: [], $0: "test", notSecret: "canSeeMe" });
         expect(data.notSecret).not.toContain(LoggerUtils.CENSOR_RESPONSE);
     });
+
+    describe("censoreRawData", () => {
```
awharn, line 58: censoreRawData -> censorRawData

awharn (commented):
Just a few small changes/comments.

__tests__/config/integration/jest.config.json
zFernand0: it was enough to get most integration tests running.
For instance there are some very old integration tests that are not contained in a `__integration__` folder
Ex: `__tests__/src/example_clis/with_experimental/ExperimentalExample.spec.ts`

Should we move them around so that they get picked up as well ?

CHANGELOG.md
zFernand0: It is not a breaking change though.
People can still use `Imperative.envVariablePrefix`, it's just deprecated 😋

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+
+- Enhancement: Added an environment variable to control whether or not sensitive data will be masked in the console output.<br/>
```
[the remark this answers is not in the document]
zFernand0: It's just so that the GH version looks a little nicer instead of having to add bullet points to the description or make the description of the enhancement a code block

gejohnston (approved):

__tests__/config/integration/jest.config.json
t1m0thyj: Looks like this was done in https://github.com/zowe/imperative/pull/697? 😋

packages/logger/src/Logger.ts
```diff
@@ -154,14 +153,6 @@ export class Logger {
      */
     private initStatus: boolean;
 
-    private static mProfileSchemas = new Map<string, IProfileSchema>();
-    public static get profileSchemas(): Map<string, IProfileSchema> {
-        return this.mProfileSchemas;
-    }
-    public static set profileSchemas(schemas: Map<string, IProfileSchema>) {
-        this.mProfileSchemas = schemas;
-    }
```
t1m0thyj: Are these ok to remove without deprecating them first?

t1m0thyj: Not sure why, the GitHub diff of new changes I hadn't seen before made it look like these methods were removed.

packages/logger/src/LoggerUtils.ts
zFernand0: JSDoc was added

__tests__/config/integration/jest.config.json
zFernand0: yes.
I'm hesitant to remove it as it may cause a merge conflict with #697
😢

packages/logger/__tests__/LoggerUtils.test.ts
zFernand0: > Do we not bother with that meta data for old profiles because it would only be a half-way solution?

`¯\_(ツ)_/¯`
I don't recall the exact details of the discussion, however I'm more than happy to open a new issue and finalize any details around the yaml-metadata.

zFernand0: Created https://github.com/zowe/imperative/issues/706

awharn (approved):

PR #692, opened 2021-11-22
[next] Improve error message when Keytar fails to load

### LINKED ISSUE #27: Keytar not Installed error masks actual error.
Receiving 'Keytar not Installed' error while debugging.  Set a breakpoint and discovered the actual error is due to keytar being compiled against an older version of Node.js. Actual error needs to be returned to the user.

The module '\\?\C:\Users\<user>\IdeaProjects\brightside\node_modules\keytar\build\Release\keytar.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 57. This version of Node.js requires
NODE_MODULE_VERSION 64. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).

t1m0thyj:
Resolves #27 by including additional details in the error message for v1 profiles


awharn (approved):

zFernand0 (approved):
Nice and simple fix! 😋

packages/config/src/ConfigUtils.ts
```diff
@@ -46,16 +46,25 @@ export function getActiveProfileName(profileType: string, cmdArguments: ICommand
         defaultProfileName || profileType;
 }
 
+/**
+ * Describe the solutions we suggest for credential manager errors.
+ * @param displayName Display name of the CLI or credential manager
+ * @returns Additional details to store on the error object
+ */
+export function secureErrorDetails(displayName: string): string {
```
awharn, line 54: Maybe we want the credential manager to provide these details instead of hardcoding them. If we use something other  then keytar in the future, the information provided here might not be accurate.

t1m0thyj: Two solutions that come to mind for this:
1. Add a condition to the else clause added in CliProfileManager above, that only provides these details `if (CredentialManager.manager instanceof DefaultCredentialManager === true)`. And create a GitHub issue suggesting that we provide different error details if we use something other than keytar in the future.
2. Add an optional `possibleSolutions` array to the AbstractCredentialManager class that can be defined by CredentialManager implementations. For DefaultCredentialManager, items 1 and 2 could be defined in this array.

Suggestions? I'm leaning towards the first one, but that's probably because I'm lazy 🙂

awharn: I kind of like the second one because it doesn't sound terrible to implement. But that's just me. We could definitely do the first one though, it's up to you. I just didn't want to approve this and forget, because if we move to something else, perhaps it won't have prereqs on Linux, perhaps it won't rely on the OS, or perhaps something goes wrong with another OS.

gejohnston (approved):
I agree that option 2 for providing error details would be a good improvement. Otherwise, everything looks good.

gejohnston (approved):
Looks good.

zFernand0 (approved):
LGTM! 😋

PR #695, opened 2021-11-29
Add system info utility

awharn:
Adds a utility to pass back an object with the architecture and platform of the system.
  - Should help with determining what version of the Zowe Daemon binary to install


packages/utilities/__tests__/ProcessUtils.test.ts
```diff
@@ -102,4 +102,120 @@ describe("ProcessUtils tests", () => {
...
     });
 
+    describe("getBasicSystemInfo", () => {
+        let realPlatform: string;
+        let realArch: string;
+
+        beforeAll(() => {
+            realPlatform = process.platform;
+            realArch = process.arch;
+        })
```
zFernand0, line 112: tiny lint warning

zFernand0 (commented):
PR LGTM!
Should this be added to master/zowe-v1-lts ?

awharn:
> PR LGTM! Should this be added to master/zowe-v1-lts ?

I don't know that it needs to be, but we can - this is kind of intended just for helping to determine basic OS info so that we can attempt to install the daemon. I really don't like messing with `process`, so I thought a little helper like this in Imperative would be a good way to go, and we can expand upon it in the future if we need to.

packages/utilities/__tests__/ProcessUtils.test.ts
awharn: Thanks.

t1m0thyj (approved):

zFernand0 (approved):
LGTM! : )

PR #698, opened 2021-12-07
[next] Add "config convert-profiles" command

t1m0thyj:
⚠️ Do not merge until #702 has been merged

```
$ zowe config convert-profiles
Detected 1 obsolete plug-in(s) and 27 old profile(s) to convert from Zowe v1 to v2.

Are you sure you want to continue? [y/N]: y

Removed obsolete plug-in: @zowe/secure-credential-store-for-zowe-cli
Converted endevor profiles: t***
Converted endevor-location profiles: d**
Converted jclcheck profiles: apiml, c***, d***, m******, x***, x***
Converted XXX profiles: apiml, c***
Converted ssh profiles: c***, c***, d***, m******, t***, x***, x***
Converted tso profiles: izuacct
Converted zftp profiles: x***, x***
Converted zosmf profiles: c***, c***, d***, d***, m******, t***, t***_ase

Your new profiles have been saved to C:\Users\Timothy\.zowe\zowe.config.json.
Run "zowe config edit --global-config" to open this file in your default editor.

Your old profiles have been moved to C:\Users\Timothy\.zowe\profiles-old.
Run "zowe config convert-profiles --delete" if you want to completely remove them.
```

PR #708, opened 2021-12-16
Add delete option to convert-profiles

awharn:
Adds the delete option to convert-profiles.
Adds functionality to delete old secure properties from the credential store.


packages/imperative/src/config/cmd/convert-profiles/convert-profiles.handler.ts
```diff
@@ -11,6 +11,8 @@
```
t1m0thyj, line 11: Do we want to disable the following check if `--delete` flag is true?
https://github.com/zowe/imperative/blob/9261c337222e2c9dcc40222988c1228f1975a62d/packages/imperative/src/config/cmd/convert-profiles/convert-profiles.handler.ts#L63-L66

Currently, it appears that if someone has already run `zowe config convert-profiles`, then they can't run `zowe config convert-profiles --delete` afterwards, since they have 0 active v1 profiles and this code block will exit the handler.

packages/imperative/src/config/cmd/convert-profiles/convert-profiles.handler.ts
```diff
@@ -192,4 +223,39 @@ export default class ConvertProfilesHandler implements ICommandHandler {
             uninstallPlugin(pluginName);
         }
     }
+
+    private async checkKeytarAvailable(): Promise<boolean> {
+        let success: boolean = undefined;
+        try {
+            await keytar.findCredentials('@zowe/cli');
```
t1m0thyj, line 230: I know the value here doesn't really matter, but would it make sense to use `this.ZOWE_CLI_PACKAGE_NAME` instead of the hardcoded string `@zowe/cli`?

packages/imperative/__tests__/config/cmd/convert-profiles/convert-profiles.handler.test.ts
```diff
@@ -189,6 +191,188 @@ describe("Configuration Convert Profiles command handler", () => {
...
+                fruit: ["apple", "coconut", "banana"]
+            },
+            profilesFailed: []
+        });
+        const updateSchemaSpy = jest.spyOn(ConfigSchema, "updateSchema").mockReturnValueOnce(undefined);
+        jest.spyOn(fs, "renameSync").mockReturnValueOnce();
+        jest.spyOn(keytar, "findCredentials").mockResolvedValue([
+            {account: "testAcct", password: "testPassword"},
+            {account: "secure_config_props", password: "testPassword"},
+            {account: "secure_config_props_1", password: "testPassword"}
```
t1m0thyj, line 254: The DefaultCredentialManager uses a hyphen before the number at the end like `secure_config_props-1`. Should we update this account string to match?

CHANGELOG.md
```diff
@@ -2,8 +2,13 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added `delete` option to `config convert-profiles` command.
+
 ## `5.0.0-next.202112151934`
 
+- Enhancement: Added `config convert-profiles` command that converts v1 profiles to team config. [zowe/zowe-cli#896](https://github.com/zowe/zowe-cli/issues/896)
```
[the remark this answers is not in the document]
awharn: Yeah, this is actually what Git did when I merged things in. I'll fix it when #698 is merged in.

packages/imperative/src/config/cmd/convert-profiles/convert-profiles.handler.ts
awharn: I might just add an if statement around the return.

packages/imperative/__tests__/config/cmd/convert-profiles/convert-profiles.handler.test.ts
awharn: We can, it will still match, but better to be more realistic in the tests.

packages/imperative/src/config/cmd/convert-profiles/convert-profiles.handler.ts
awharn: Yes. Should we use that above as well?

gejohnston (approved):

packages/imperative/src/config/cmd/convert-profiles/convert-profiles.handler.ts
t1m0thyj: No, in the `knownServices` array, I think it makes sense for all the strings to be hardcoded, since they should remain the same even if ZOWE_CLI_PACKAGE_NAME ever changes.

t1m0thyj (approved):
LGTM! Thanks for also addressing https://github.com/zowe/imperative/issues/711 😛

packages/imperative/__tests__/config/cmd/convert-profiles/convert-profiles.handler.test.ts
```diff
@@ -237,6 +236,52 @@ describe("Configuration Convert Profiles command handler", () => {
...
+        });
+
+        const handler = new ConvertProfilesHandler();
+        jest.spyOn(handler as any, "getObsoletePlugins").mockReturnValueOnce([]);
+        jest.spyOn(handler as any, "getOldProfileCount").mockReturnValueOnce(3);
+        const findOldSecurePropsSpy = jest.spyOn(handler as any, "findOldSecureProps");
+        const deleteOldSecurePropsSpy = jest.spyOn(handler as any, "deleteOldSecureProps");
+
+        const params = getIHandlerParametersObject();
+        (params.response.console.prompt as any).mockResolvedValueOnce("y").mockResolvedValueOnce("y");
```
t1m0thyj, line 265: Should there be another unit test that verifies profiles are converted, but not deleted, with the following prompt inputs?
`(params.response.console.prompt as any).mockResolvedValueOnce("y").mockResolvedValueOnce("n");`

t1m0thyj (approved):

t1m0thyj (changes requested):
Sorry that I keep on thinking of more edge cases 😢 

For the scenario that profiles have already been converted but not deleted, should we add an integration test that runs `config convert-profiles --delete`? It could verify that:
* Old profiles in "profiles-old" directory are deleted, even though no old profiles were found in "profiles" directory to convert.
* Only one prompt (the 2nd one) should be printed because the 1st one does not apply.

t1m0thyj (approved):
LGTM! Thanks for finding and fixing edge cases 🙂

PR #712, opened 2021-12-20
[next] Fix inconsistent error message in daemon mode

t1m0thyj:
Fixes https://github.com/zowe/zowe-cli/issues/1081 by building `failedCommandDefinition` every time a command fails. Previously, the object was only built once when Imperative initialized, so it had a hardcoded command line of "zowe --daemon" rather than the invalid command.


awharn (approved):

zFernand0 (approved):
LGTM!

gejohnston (approved):

PR #714, opened 2021-12-22
[next] 2-way JSON communication for daemon mode

t1m0thyj:
Implement JSON communication for data sent from client-to-server, similar to what already exists for server-to-client.

The format of the JSON body is documented in the corresponding PR https://github.com/zowe/zowe-cli/pull/1244.

The JSON object was added as a property that gets passed to CommandProcessor: `ICommandProcessorParms.daemonResponse`. This makes the `--dcd` argument unnecessary so it was removed.

### DIFF
--- a/packages/cmd/src/CommandProcessor.ts
+++ b/packages/cmd/src/CommandProcessor.ts
@@ -43,6 +43,8 @@ import { CliUtils } from "../../utilities/src/CliUtils";
 import { WebHelpManager } from "./help/WebHelpManager";
 import { ICommandProfile } from "./doc/profiles/definition/ICommandProfile";
 import { Config } from "../../config/src/Config";
+import { IDaemonResponse } from "../../utilities/src/doc/IDaemonResponse";
+
 /**
  * The command processor for imperative - accepts the command definition for the command being issued (and a pre-built)
  * response object and validates syntax, loads profiles, instantiates handlers, & invokes the handlers.
@@ -129,6 +131,13 @@ export class CommandProcessor {
      * @memberof CommandProcessor
      */
     private mConfig: Config;
+    /**
+     * Daemon response object containing process info.
+     * @private
+     * @type {IDaemonResponse}
+     * @memberof CommandProcessor
+     */
+    private mDaemonResponse?: IDaemonResponse;
 
     /**
      * Creates an instance of CommandProcessor.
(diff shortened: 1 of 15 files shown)


packages/imperative/src/Imperative.ts
```diff
@@ -365,7 +365,7 @@ export class Imperative {
      * Parse command line arguments and issue the user's specified command
      * @returns {Imperative} this, for chaining syntax
      */
-    public static parse(args?: string, context?: IYargsContext): Imperative {
+    public static parse(args?: string | string[], context?: IYargsContext): Imperative {
```
zFernand0, line 368: This may be considered a breaking change for TS CLI developers.
As an example: VScode's `TreeItem.label` property changed from `string` to `string | TreeItemLabel` and some extensions had to adjust to the change eventually.

packages/cmd/src/CommandProcessor.ts
```diff
@@ -384,11 +394,24 @@ export class CommandProcessor {
...
                 // directories may cause unpredictable results
-                process.chdir(params.arguments.dcd as string);
+                if (this.mDaemonResponse.cwd != null) {
+                    process.chdir(this.mDaemonResponse.cwd);
+                }
+
+                // Delete environment variables that start with CLI prefix
+                for (const envVarName of Object.keys(process.env)) {
+                    if (envVarName.startsWith(`${this.mEnvVariablePrefix}_`)) {
+                        delete process.env[envVarName];
```
zFernand0, line 408: Sounds a little scary to remove all CLI prefixed ENV vars from the process.
What happens when multiple users are making use of the same daemon process ?
What if the user had Zowe's prefix (`ZOWE_`) in ENV variables that affected a specific plugin behavior ?

packages/utilities/src/doc/IDaemonResponse.ts
```diff
@@ -9,25 +9,37 @@
...
-     * Content is reply from daemon
-     * @type {string}
-     * @memberof IDaemonResponse
+     * Stdin text received from the daemon client.
+     * This is used for plain text stdin data like replies to prompts.
      */
-    reply?: string;
-}
\ No newline at end of file
+    stdin?: string;
```
zFernand0, line 44: Next Only breaking change, right ?

zFernand0 (commented):
Implementation seems sound. 🎵😋

gejohnston (approved):

packages/cmd/src/CommandProcessor.ts
t1m0thyj: Mentioned this as a limitation on the Zowe CLI PR:
> The way environment variables and stdin data are handled will need to be enhanced in the future when we support multitenancy. While worth noting, this is not a regression since the old implementation also didn't support multitenancy.

Do you think the current implementation is acceptable for now? Our codebase already makes extensive use of the `process` module for "cwd", "env", and "stdin", all of which may break if multiple users make use of the same daemon process.

zFernand0: I think it's fine for the case of multitenancy.
I'm curious how this may affect plugins that may depend on a ZOWE_ ENV variable that we may be deleting here 😋

zFernand0: Thanks for clarifying offline.
Ok to resolve 😋

zFernand0 (approved):

PR #721, opened 2022-01-11
Introduce `--show-inputs-only` Flag

dkelosky:
Draft PR to address: https://github.com/zowe/zowe-cli/issues/870.  

The draft name of the flag is `--show-resolved-args` and is a hidden flag.  I wanted to use `--dry-run` but that appears to be taken in `zowe config init` 😢 .  

If I use old profiles which contain connection info, set environmental variables like `set ZOWE_OPT_PASS=fakepass`, and supply command line arguments, `--show-resolved-args` describes what input would be used for the given command without actually executing the command:
image

Or:
```
C:\>zowe endevor list elements --user mrendvr --host endevor.com --show-resolved-args --subsystem notreal
commandValues:
  environment:        dev
  system:             GACODE
  subsystem:          notreal
  stage-number:       1
  maxrc:              12
  data:               all
  host:               endevor.com
  port:               50002
  instance:           WEBSMFNE
  protocol:           http
  user:               mrendvr
  password:           fakepass
  show-resolved-args: true
optionalProfiles:
  - endevor
  - endevor-location
  - base
```

This is a draft only PR.  We can change or close this in it's current form, but I'm hopeful for some feedback.  

Signed-off-by: Dan Kelosky <dkelosky@gmail.com>


MikeBauerCA:
This would be an excellent debugging tool as you can easily see what values are being used. Oftentimes it is tedious to figure this out as you may need to look at multiple configs and env vars.

If a value is stored in the secure credential store, I would want a "securely stored"-like message to be displayed rather than the value itself. If these values must be printed to output, I think it should made very clear to the end user that securely stored properties are about to be displayed - perhaps with an a prompt to continue. I'd like to get some feedback from the community before merging the PR if the secure values are to be printed to output.

I'm also open to this being a visible flag, allowing users to debug on their own. This decision is also influenced by whether secure values are printed.

dkelosky:
thanks for looking, we should be able to accommodate either the prompt or a flag... I'll experiment and get back.

it turns out, in my case, I wanted to see both user and password because I was getting a http 403.  it turned out that both user and password were unexpected when I used this flag, and so I was able to solve my problem

MikeBauerCA:
@dkelosky another option would be to add a `--prompt` flag with a default value of `true`. May be easier for users to append `--prompt false` than set an env var.

dkelosky:
@MikeBauerCA good thought.  a reason I was avoiding `--no-prompt` would be a global option on every command, but i'd really want a `--no-prompt` flag which only applies to a `--show-resolved-args` flag.

i agree `--no-prompt` is easier for for most interactive users - except for me, i'd put `ZOWE_SHOW_RESOLVED_ARGS=true` as a system environmental variable 😄

t1m0thyj:
This looks really useful for debugging 🙂 A few thoughts:
* Would it be useful to show where each command value came from? Like `env var`, `cli arg`, `profile`, etc. Depending on how advanced this logic is, it could show more specific info like "base profile" vs "service profile", or "base profile from global config" vs "zosmf.lpar1 profile from project config".
* IMO it would be nice if we could use `--dry-run` for the argument name. I wonder if we should consider making a v2-v2 breaking change on `zowe config init` and `zowe config auto-init` that would rename `--dry-run` to something else like `--preview-changes`. Or maybe we could add an optional dry run handler to commands, allowing them to run their own custom code after we execute your default `--dry-run` logic in CommandProcessor.

dkelosky:
I'm with you there.  When we talked about this in the original issue, we planned on `--dry-run`.  I hadn't realized another `--dry-run` had come along in the meantime; however, I do think `--dry-run` would be far more intuitive for this purpose than to only apply to `zowe config`.  @MikeBauerCA - what do you think?


I'll keep your first point in mind when I revisit this.  I forget how much detail I was able to derive at the time this runs to determine things like where arguments came from.  Thanks for the suggestion.  I'll try to get back to this soon!

MikeBauerCA:
@dkelosky @t1m0thyj 
What do you think of `show-input-only`, `show-input`, `input`, or something similar? 

In addition to being a v2-v2 breaking change, I worry that a command like `zowe jobs submit ds "hlq.ex(ample)" --dry-run` would give the impression that perhaps it is checking if the job would run successfully without actually running it. There may be other similar examples. Perhaps it isn't significant but we may be able to use something other than `show-resolved-args` or `dry-run`.

dkelosky:
Here's a stab at `--show-inputs-only`.  By default, hides user name and password the same way prompting for certain command parameters or log messages are masked (as a side note, there does not appear to be a single place where a "secure" items list is retained):
image

When you set the env, the real values are displayed without the red text.

Please let me know what you think, and I'll start finalizing this.

dkelosky:
right now `--show-secure-args` is not an option since it only made sense if `--show-inputs-only` is also present.  loosely, it also seems similar to log level adjustments which are only via env vars.  is that fair?

t1m0thyj:
This looks good to me and I like the name `--show-inputs-only` 🙂

To determine which values should be treated as secure, I believe the default list of secure values is defined in `LoggerUtils.CENSORED_OPTIONS`. I'll defer to @zFernand0 on how to get the list of user-defined secure values from team config.

dkelosky:
I copied this list: https://github.com/zowe/imperative/blob/master/packages/rest/src/session/ConnectionPropsForSessCfg.ts#L294

@zFernand0 WHAT DO

zFernand0:
> how to get the list of user-defined secure values from team config

For the user-defined  secure properties, isn't `config.api.secure.secureFields()` enough? 
It will get you the full path, and if you know the profile name, then doing `config.api.secure.securePropsForProfile("myProfile")` should work.

Maybe I'm missing the point 😋 

> WHAT DO

`¯\_(ツ)_/¯`

zFernand0:
@dkelosky How about this?
https://github.com/zowe/imperative/compare/dry-run-show-args...test-dry-run-branch?expand=1

Result
image

dkelosky:
I can use that util - my question was because I saw this: `private static readonly secureSessCfgProps: string[] = ["user", "password", "tokenValue", "passphrase"];` [here](https://github.com/zowe/imperative/blob/master/packages/rest/src/session/ConnectionPropsForSessCfg.ts#L294).

dkelosky:
I had unpushed changes pending some feedback so I can't use the diff as-is but I'll pull the code in.  I think I'll try to fall back to the list from `ConnectionPropsForSessCfg` in case old profiles are in effect.

zFernand0:
> I can use that util - my question was because I saw this: `private static readonly secureSessCfgProps: string[] = ["user", "password", "tokenValue", "passphrase"];` [here](https://github.com/zowe/imperative/blob/master/packages/rest/src/session/ConnectionPropsForSessCfg.ts#L294).

I knew I was missing something 😋 
What about making that a public property of the class (or introduce a getter `/shrug`) ? 😋

dkelosky:
✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️

zFernand0:
> ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️

@dkelosky , Where are the 2 skipped ones ?
🚫 🚫 ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️

dkelosky:
I made a mistake - THANKS FOR NOTICING 😄

gejohnston:
Sorry that I came so late to the table. I agree with the value of this flag, and in the direction that your UX conversation has taken things.

I just have one question: The CHANGELOG refers to this as a "hidden flag". Does this imply that we do not want users to see this flag in our online help?

dkelosky:
I marked the flag as hidden to avoid [what I thought might be] cluttering up the help menu, not to hide it from the users.  I have no strong opinion though - if a visible flag is preferred, I can make that change

packages/cmd/src/CommandProcessor.ts
```diff
@@ -721,6 +733,122 @@ export class CommandProcessor {
...
+        // only attempt to show the input if it is in the command definition
+        for (let i = 0; i < commandParameters.definition.options?.length; i++) {
+            const name = commandParameters.definition.options[i].name;
+
+            if (commandParameters.arguments[name] != null) {
+
+                if (showSecure || !secureInputs.has(name)) {
+                    showInputsOnly.commandValues[name] = commandParameters.arguments[name];
+                } else {
+                    showInputsOnly.commandValues[name] = LoggerUtils.CENSOR_RESPONSE;
```
t1m0thyj, line 807: Another option for a "censor response" would be `(secure value)` which is what the "config list" command uses.

Not sure which value is preferred here, just mentioning this for consideration 🙂 Perhaps we set a precedent of using `****` in log files and `(secure value)` in CLI output?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Introduce hidden flag `--show-inputs-only` to show the inputs of the command
+that would be used if a command were executed.
```
t1m0thyj, line 8: Based on discussion at standup, I think we agreed that making the option unhidden would be preferred, as long as there's an easy way to define it at the top level so that it applies to all commands but not command groups.

@gejohnston @awharn @zFernand0 Do you recall the same outcome of our discussion?

gejohnston: Yes Timothy, the visibility of the flag as you described it was what I recall that Mike B advocated. In fact he said he would add that comment to this PR, but I suspect that he got sidetracked.

dkelosky:
Thanks all - I'll try to get these changes "soon"

packages/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -13,7 +13,7 @@ import { CliUtils, ImperativeConfig } from "../../../utilities";
 import { ICommandArguments, IHandlerParameters } from "../../../cmd";
 import { ImperativeError } from "../../../error";
 import { IOptionsForAddConnProps } from "./doc/IOptionsForAddConnProps";
-import { Logger } from "../../../logger";
+import { Logger, LoggerUtils } from "../../../logger";
```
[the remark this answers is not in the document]
dkelosky: i blame your CI 😎 - fixed

packages/cmd/src/CommandProcessor.ts
```diff
@@ -721,6 +733,122 @@ export class CommandProcessor {
...
+
+        /**
+         * Determine if Zowe V2 Config is in effect.  If it is, then we will construct
+         * a Set of secure fields from its API.  If it is not, then we will construct
+         * a Set of secure fields from the `ConnectionPropsForSessCfg` defaults.
+         */
+        const secureInputs: Set<string> =
+            useConfig ?
+                new Set(this.mConfig.api.secure.secureFields().map(v => v.split('.').slice(-1)[0])) :
+                new Set([...LoggerUtils.CENSORED_OPTIONS, ...LoggerUtils.SECURE_PROMPT_OPTIONS]);
```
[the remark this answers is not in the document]
dkelosky: This one is tricky.  I spent some time with this to get it to work, and ideally, I'd like this censor logic to be in one place.  

`LoggerUtils.isSpecialValue` has a copied list of fields from [here](https://github.com/zowe/imperative/blob/master/packages/rest/src/session/ConnectionPropsForSessCfg.ts#L294).  This copied list is what we fall back to for v1 so we "support"  `--show-inputs-only` for both versions.  

I'm not clear if `LoggerUtils.isSpecialValue` is a generic API.  It adds a `.` to the special property name:
```typescript
        for (const v of specialValues) {
            if (prop.endsWith(`.${v}`)) return true;
        }
```
I'm also unclear where but with debug printing messages, I get `protocol` in the output when I print `specialValues`:
```
[
  'user',
  'password',
  'tokenValue',
  'keyPassphrase',
  'token-value',
  'base-path',
  'protocol',
  'token'
]
```

Unless there is something obvious I'm missing, I'd like to leave reconciling the secure arrays as a future tech debt item to resolve apart from this enhancement. 

Side note, idk if I'm reading this correctly, but is this logic correct?:
```typescript
            if (sec && typeof sec !== "object" && !LoggerUtils.isSpecialValue(prop))
                newData = newData.replace(new RegExp(sec, "gi"), LoggerUtils.CENSOR_RESPONSE);
```

The last condition reads like: "if not is special value then replace data"?  It seems like it would be, "if is special value then censor".

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Introduce flag `--show-inputs-only` to show the inputs of the command
```
t1m0thyj, line 7: `Introduce` -> `Introduced` - please use past tense for consistency with other changelog messages 🙂

__tests__/__integration__/cmd/__tests__/integration/cli/gen-help/__snapshots__/Cmd.cli.gen-help.allowable-options.integration.test.ts.snap
```diff
@@ -67,6 +67,10 @@ exports[`cmd-cli gen-help allowable-options should display the help with allowab
  GLOBAL OPTIONS
  --------------
 
+   --show-inputs-only (boolean)
+
+      Show command inputs and do not run the command
+
    --response-format-json | --rfj (boolean)
 
       Produce JSON formatted data from a command
```
t1m0thyj, line 76: Would it be ideal for the "Global Options" to be listed in alphabetical order? I know they already were out of order with `--rfj` coming before `-h`, but now that we're adding more options to the list it strikes me as odd that they have no clear order.

CHANGELOG.md
dkelosky: fixe**D**

__tests__/__integration__/cmd/__tests__/integration/cli/gen-help/__snapshots__/Cmd.cli.gen-help.allowable-options.integration.test.ts.snap
dkelosky: I think alphabetical menu is a good idea if universally applied.  `ZOSMF CONNECTION OPTIONS`, `RESPONSE FORMAT OPTIONS`, and `OPTIONS` are also non-alphabetical.  

`CommandPreparer.appendAutoOptions` occurs in logic after the new `CommandPreparer.appendPassOnOptions`.  Perhaps this could be adjusted, but I'm unsure of the implications.  `passOn` has a `merge` behavior and changing the ordering might be non-trivial.  Is it acceptable to address menu organization as a separated issue?

t1m0thyj: Good points, changing the order could be non-trivial and seems outside the scope of this PR. Created #761 🙂

packages/cmd/src/CommandProcessor.ts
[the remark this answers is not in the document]
zFernand0: > It adds a . to the special property name

that's v2 only, for v1 I guess we could replace the . with a _ 🙂
> I'd like to leave reconciling the secure arrays as a future tech debt item to resolve apart from this enhancement

That's fine by me (though we may never get to it) 😛
> The last condition reads like: "if not is special value then replace data"? It seems like it would be, "if is special value then censor".

Definitely feels like a bug unless the isSpecialValue give backward results 😛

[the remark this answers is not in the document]
dkelosky: Thanks for reviewing @zFernand0 !

dkelosky:
Yo - can you guys let me know when would be a good time would be a good time to get this reviewed?  I'll delay `CHANGELOG.md`  conflict resolution until then.  

It's kind of a pain to keep resolving the same conflict only to need to do it again a few days later.  Is there some process improvement here that would make contributions more seamless?  I don't have any good ideas, but I think this current approach might discourage contributions.

packages/cmd/src/CommandProcessor.ts
```diff
@@ -721,6 +777,115 @@ export class CommandProcessor {
...
+        });
+
+        /**
+         * Determine if Zowe V2 Config is in effect.  If it is, then we will construct
+         * a Set of secure fields from its API.  If it is not, then we will construct
+         * a Set of secure fields from the `ConnectionPropsForSessCfg` defaults.
+         */
+        const secureInputs: Set<string> =
+            useConfig ?
+                new Set(this.mConfig.api.secure.secureFields().map(v => v.split('.').slice(-1)[0])) :
```
t1m0thyj, line 820: I think `this.mConfig.api.secure.securePropsForProfile` should be used here instead. Getting the list of all secure fields in the file may have undesired behavior. For example with the following config:
```
{
    "profiles": {
        "lpar1": {
            "type": "zosmf",
            "properties": {
                "host": "lpar1.broadcom.net",
                "port": 443
            },
            "secure": []
        },
        "lpar2": {
            "type": "zosmf",
            "properties": {
                "port": 1443
            },
            "secure": [
                "host"
            ]
        },
        "base": {
            "type": "base",
            "properties": {
                "rejectUnauthorized": true
            },
            "secure": [
                "user",
                "password"
            ]
        }
    },
    "defaults": {
        "zosmf": "lpar1",
        "base": "base"
    },
    "autoStore": true
}
```

Here the host property is only secure in the lpar2 profile, so running `zowe jobs ls jobs --show-inputs-only` should show the host property in plain text that is present in the default lpar1 profile. Currently I see `host: (secure value)` instead.

gejohnston (approved):

packages/cmd/src/CommandProcessor.ts
dkelosky: your point makes sense - i suppose the same will happen if a base profile marks `host` secure while zosmf does not; i think `host` will be in the secure list.  the undesired behavior is that we may mistakenly show something as `(secure value)` when it is not.

one struggle i have - i'm not clear how to determine the name of the profile used from `CommandProcessor`.  i think we had this problem before in command handlers: we have access to the profile contents / values and type but not the profile name to use `securePropsForProfile`.  do you see some way to determine this?

CHANGELOG.md
dkelosky: side note - and I'm being a little difficult here - if this is a standard it should be documented in `contributing.md`.  if it is, it's fine to link it.  otherwise, these things feel arbitrary and unknowable to casual contributors.

packages/cmd/src/CommandProcessor.ts
t1m0thyj: Sorry I missed this comment until now, I believe `ConfigUtils.getActiveProfileName` might help here: https://github.com/zowe/imperative/blob/next/packages/config/src/ConfigUtils.ts#L49

dkelosky: side note, i noticed, `ConfigUtils.getActiveProfileName` returns `host` as a secure prop for `zosmf2` even though it is not:
```
        "lpar1": {
            "properties": {
                "host": "usilca11.lvn.broadcom.net"
            },
            "profiles": {
                "zosmf": {
                    "type": "zosmf",
                    "properties": {
                        "port": 1443
                    },
                    "secure": [
                        "host"
                    ]
                },
                "zosmf2": {
                    "type": "zosmf",
                    "properties": {
                        "port": 1443
                    },
                    "secure": []
                },
```

in any case, this is only a display command option to help users.  the only impact in worst case scenario is that we show a field as `(secure value)` when it is not.  it also seems unlikely in both of our examples that a user would want a field like `host` secure for one `zosmf` profile and not another.  

no action needed - i'm just recording considerations for this discussion

gejohnston: I thought that a PR was recently merged which resolved an issue about returning host as secure when it is only secure for some profiles. Perhaps after you merge next into this branch, the problem will go away?

t1m0thyj (approved):

packages/cmd/src/CommandProcessor.ts
t1m0thyj: I believe the latest changes from next have been merged in, if this behavior persists I think we should file it as an issue. 🙂

dkelosky: correction, it's `this.mConfig.api.secure.securePropsForProfile` instead of `ConfigUtils.getActiveProfileName` that causes an issue.  

it still happens to me after syncing with `next` this morning.  i dont get an issue using @t1m0thyj's sample profile with `lpar1.zosmf` and `lpar2.zosmf`, but i do get an issue with `lpar1.zosmf` and `lpar1.zosmf2` as in my example.  

two instances of zosmf on the sample lpar2 don't make sense, but i the `lpar` higher level is an arbitrary construction.  

Here is the code with trace messages added:
```typescript
    public securePropsForProfile(profileName: string) {
        console.log(`>>>> entry <<<<`)
        const profilePath = this.mConfig.api.profiles.expandPath(profileName);
        console.log(`profilePath: ${profilePath}`)
        const secureProps = [];
        for (const propPath of this.secureFields()) {
            console.log(`propPath: ${propPath}`)
            const pathSegments = propPath.split(".");  // profiles.XXX.properties.YYY
            // eslint-disable-next-line @typescript-eslint/no-magic-numbers
            if (profilePath.startsWith(pathSegments.slice(0, -2).join("."))) {
                secureProps.push(pathSegments.pop());
            }
        }
        console.log(`secureProps ${secureProps}`)
        console.log(`>>>> exit <<<<`)
        return secureProps;
    }
```

here's the command i run and output I see:
```
C:\dev\systems\ca11>zowe jobs list jobs --show-inputs-only --zosmf-p lpar1.zosmf2
>>>> entry <<<<
profilePath: profiles.lpar1.profiles.zosmf2
propPath: profiles.lpar1.profiles.zosmf.properties.host
propPath: profiles.my_base.properties.user
propPath: profiles.my_base.properties.password
secureProps host
>>>> exit <<<<
>>>> entry <<<<
profilePath: profiles.my_base
propPath: profiles.lpar1.profiles.zosmf.properties.host
propPath: profiles.my_base.properties.user
propPath: profiles.my_base.properties.password
secureProps user,password
>>>> exit <<<<
```

it looks like `zosmf2` becomes `zosmf` in `ConfigProfiles.expandPath`

PR #727, opened 2022-01-27
Improve Imperative logging for the Credential Manager

t1m0thyj:
* Fixed `causeErrors` missing from the Imperative log file. The default behavior of `util.inspect` is to print objects up to 2 levels deep. This works fine for a single ImperativeError object where `causeErrors` is 2 levels deep, but when there is an array of ImperativeError objects then `causeErrors` is 3 levels deep.
* Added more details to the error message displayed to the user when Keytar fails to load. Thanks @gejohnston for this suggestion 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed useful debugging information missing from error message when Keytar module fails to load.
+
 ## `5.0.0-next.202201102100`
 
 - BugFix: Fixed ZOWE_CLI_HOME environment variable not respected by team config in daemon mode. [zowe/zowe-cli#1240](https://github.com/zowe/zowe-cli/issues/1240)
--- a/packages/cmd/src/yargs/CommandYargs.ts
+++ b/packages/cmd/src/yargs/CommandYargs.ts
@@ -147,7 +147,7 @@ export class CommandYargs extends AbstractCommandYargs {
                                     `Error in command ${this.definition.name}`,
                                     "command handler invoked", errorResponses);
                                 this.log.error(`Error in command ${this.definition.name}`);
-                                this.log.error(inspect(errorResponses));
+                                this.log.error(inspect(errorResponses, { depth: null }));
                                 commandExecuted(argsForHandler, response);
                             });
                     } else {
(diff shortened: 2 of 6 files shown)


gejohnston:
Given that we have had to debug SCS problems at a few different sites, have you considered adding some SCS-related log messages (maybe debug-level messages) at various key steps along the way, in addition to the improved details when an error actually occurs?

I am not claiming that I know that there are really useful pieces of information to display. I just want to make sure that we have considered the value of adding such debug messages before we complete this update.

zFernand0 (approved):

t1m0thyj:
> Given that we have had to debug SCS problems at a few different sites, have you considered adding some SCS-related log messages (maybe debug-level messages) at various key steps along the way, in addition to the improved details when an error actually occurs?
> 
> I am not claiming that I know that there are really useful pieces of information to display. I just want to make sure that we have considered the value of adding such debug messages before we complete this update.

@gejohnston I added a debug-level message right before importing Keytar that logs the path where we found it, and another debug-level message that gets logged immediately when we fail to import Keytar, rather than waiting until the time when we try to load/store credentials.

I'm not sure if there are more debug-level messages that would make sense to add, but am open to ideas 🙂

gejohnston:
Re: I'm not sure if there are more debug-level messages that would make sense to add, but am open to ideas

> > Given that we have had to debug SCS problems at a few different sites, have you considered adding some SCS-related log messages (maybe debug-level messages) at various key steps along the way, in addition to the improved details when an error actually occurs?
> > I am not claiming that I know that there are really useful pieces of information to display. I just want to make sure that we have considered the value of adding such debug messages before we complete this update.
> 
> @gejohnston I added a debug-level message right before importing Keytar that logs the path where we found it, and another debug-level message that gets logged immediately when we fail to import Keytar, rather than waiting until the time when we try to load/store credentials.
> 
> I'm not sure if there are more debug-level messages that would make sense to add, but am open to ideas 🙂

Are there any steps that could be successful before we hit a true error message? Debug messages that some early steps were successful might be helpful, just to see what occurred. I do not see any such steps that we could log. If you are not aware of any such steps either, then I am good with what you have added.

gejohnston (approved):

awharn (approved):

t1m0thyj:
> Are there any steps that could be successful before we hit a true error message? Debug messages that some early steps were successful might be helpful, just to see what occurred. I do not see any such steps that we could log. If you are not aware of any such steps either, then I am good with what you have added.

Ideally we could inject logging statements into the Keytar package since they handle loading the native module, but I think the best we can do for now is work with the error messages they give us. These error messages may arise either when we require Keytar, or when we call its `getPassword`/`setPassword` methods. As long as we are satisfied with the logging around those steps, then I also don't see any more places to add log statements in Imperative.

PR #731, opened 2022-01-28
Fix angle brackets escaped in web help code blocks

t1m0thyj:
Both the [`markdown-it`](https://github.com/markdown-it/markdown-it/issues/167) and [`sanitize-html`](https://github.com/apostrophecms/sanitize-html/issues/249) packages used to generate web help should handle angle brackets correctly.

The problem seems to have been we were calling them in the wrong order - sanitizing an MD/HTML mix before passing it to `markdown-it` rather than sanitizing the output HTML resulted in some unexpected behavior.

While investigating this, I realized the amount of HTML tags being injected into the MD we feed to `markdown-it` was very small (only in the method `buildChildrenSummaryTables`). So this PR replaces those HTML tags with Markdown, removes the `html: true` option from the `markdown-it` parser, and no longer requires the `sanitize-html` package 🙂

Here is a diff showing the changes introduced by this branch to the `all.html` file in web help: [web-help-730.patch](https://github.com/zowe/imperative/files/7959707/web-help-730.patch.txt)

### DIFF
--- a/packages/cmd/src/help/WebHelpGenerator.ts
+++ b/packages/cmd/src/help/WebHelpGenerator.ts
@@ -159,7 +159,7 @@ export class WebHelpGenerator {
         const helpGen = new DefaultHelpGenerator({ produceMarkdown: true, rootCommandName } as any,
             { commandDefinition: uniqueDefinitions, fullCommandTree: uniqueDefinitions, skipTextWrap: true });
         rootHelpContent += this.renderMarkdown(this.buildChildrenSummaryTables(helpGen, rootCommandName) + "\n\n" +
-            helpGen.buildGlobalOptionsSection().replace(/Global options/, "Global Options"));
+            helpGen.buildGlobalOptionsSection().replace("# Global options", "# Global Options"));
         this.singlePageHtml = rootHelpContent.replace(/<h4>Groups.+?<\/ul>/s, "");
         rootHelpContent += this.genDocsFooter();
         fs.writeFileSync(rootHelpHtmlPath, rootHelpContent);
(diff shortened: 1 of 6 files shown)


packages/cmd/src/help/WebHelpGenerator.ts
```diff
@@ -264,18 +263,14 @@ export class WebHelpGenerator {
...
+            .map((line: string) => {
                 // Wrap group/command names inside links
                 const match = line.match(/^\s{0,4}([a-z0-9-]+(?:\s\|\s[a-z0-9-]+)*)\s+\S/i);
                 if (match) {
                     const href = `${hrefPrefix}${match[1].split(" ")[0]}.html`;
-                    return `\n* <a href="${href}">${match[1]}</a> -` + line.slice(match[0].length - 2);
-                } else if (/^\s*#{4}\s*COMMANDS\s*$/i.test(line)) {
-                    return `${(index > 0) ? "\n" : ""}<h4>Commands</h4>\n`;
-                } else if (/^\s*#{4}\s*GROUPS\s*$/i.test(line)) {
-                    return `${(index > 0) ? "\n" : ""}<h4>Groups</h4>\n`;
```
t1m0thyj, line 276: This "hack" is no longer needed, because I fixed the casing of "Commands" and "Groups" in DefaultHelpGenerator.

packages/cmd/src/help/WebHelpGenerator.ts
```diff
@@ -320,17 +315,13 @@ export class WebHelpGenerator {
             { commandDefinition: definition, fullCommandTree: this.mFullCommandTree, skipTextWrap: true });
 
         let markdownContent = helpGen.buildHelp() + "\n";
-        markdownContent = markdownContent.replace(/</g, "&lt;").replace(/>/g, "&gt;");
```
t1m0thyj, line 323: This "hack" is no longer needed, since `markdown-it` handles angle brackets correctly. It seems to have been left behind as a remnant from when we were using `marked`.

packages/cmd/src/help/WebHelpGenerator.ts
```diff
@@ -320,17 +315,13 @@ export class WebHelpGenerator {
...
         if (definition.type === "group") {
             // this is disabled for the CLIReadme.md but we want to show children here
             // so we'll call the help generator's children summary function even though
             // it's usually skipped when producing markdown
             markdownContent += this.buildChildrenSummaryTables(helpGen, rootCommandName + "_" + fullCommandName);
         }
 
-        // Prevent line breaks from being lost during Markdown to HTML conversion
-        markdownContent = markdownContent.replace(/^(\s+Default value:.+$)(\s+Allowed values:.+$)/gm, "$1\n$2");
```
t1m0thyj, line 332: This "hack" is no longer needed, thanks to https://github.com/zowe/zowe-cli/pull/1275

gejohnston (approved):

awharn (approved):

PR #744, opened 2022-03-03
Enhancements to plugins list command

### LINKED ISSUE #489: zowe plugins list - alphabetical order
Currently, it appears that `zowe plugins list` returns the list of installed plugins in the order in which they were installed. Over time this list may become large and difficult to read through. I suggest we order the list alphabetically by pluginName.

t1m0thyj:
Resolves #489 and #743

### DIFF
--- a/packages/imperative/src/plugins/cmd/list/list.definition.ts
+++ b/packages/imperative/src/plugins/cmd/list/list.definition.ts
@@ -25,7 +25,7 @@ export const listDefinition: ICommandDefinition = {
     options: [
         {
             name: "short",
-            aliases: ["-s"],
+            aliases: ["s"],
             description: "Show output in abbreviated format",
             type: "boolean"
         }
--- a/packages/imperative/src/plugins/cmd/list/list.definition.ts
+++ b/packages/imperative/src/plugins/cmd/list/list.definition.ts
@@ -21,5 +21,13 @@ export const listDefinition: ICommandDefinition = {
     type: "command",
     summary: "List installed plug-ins",
     description: "List all plug-ins installed.",
-    handler: join(__dirname, "list.handler")
+    handler: join(__dirname, "list.handler"),
+    options: [
+        {
+            name: "short",
+            aliases: ["-s"],
+            description: "Show output in abbreviated format",
+            type: "boolean"
+        }
+    ]
 };
--- a/packages/imperative/src/plugins/cmd/list/list.handler.ts
+++ b/packages/imperative/src/plugins/cmd/list/list.handler.ts
@@ -48,21 +48,25 @@ export default class ListHandler implements ICommandHandler {
         let listOutput: string = "";
         let firstTime = true;
 
-        for (const pluginName in installedPlugins) {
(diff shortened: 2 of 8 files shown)


packages/imperative/src/plugins/cmd/list/list.handler.ts
```diff
@@ -48,21 +48,25 @@ export default class ListHandler implements ICommandHandler {
...
+                    listOutput = listOutput + `${chalk.yellow.bold(" -- pluginName: ")}` +
+                        `${chalk.red.bold(pluginName)} \n`;
+                    listOutput = listOutput + `${chalk.yellow.bold(" -- package: ")}` +
+                        `${chalk.red.bold(installedPlugins[pluginName].package)} \n`;
+                    listOutput = listOutput + `${chalk.yellow.bold(" -- version: ")}` +
+                        `${chalk.red.bold(installedPlugins[pluginName].version)} \n`;
+                    listOutput = listOutput + `${chalk.yellow.bold(" -- registry: ")}` +
+                        installedPlugins[pluginName].registry + "\n\n";
+                } else {
+                    listOutput += `${chalk.yellow(pluginName)}@${installedPlugins[pluginName].version}\n`;
```
t1m0thyj, line 68: Not sure whether to show `pluginName@version` or `package@version`? Can't remember seeing a case where `pluginName` and `package` are different.

gejohnston: Perhaps they are always the same. If they are ever different, a user who does a 'plugin list' is to get the correct name to be able to run 'plugin uninstall' would probably need the package name. Can you think of other use cases in which the pluginName would be the needed output?

gejohnston (approved):

packages/imperative/src/plugins/cmd/list/list.definition.ts
```diff
@@ -21,5 +21,13 @@ export const listDefinition: ICommandDefinition = {
     type: "command",
     summary: "List installed plug-ins",
     description: "List all plug-ins installed.",
-    handler: join(__dirname, "list.handler")
+    handler: join(__dirname, "list.handler"),
+    options: [
+        {
+            name: "short",
+            aliases: ["-s"],
```
awharn, line 28: Is there a reason this is -s instead of s?

awharn (commented):
One question

packages/imperative/src/plugins/cmd/list/list.definition.ts
t1m0thyj: Good catch, fixed

packages/imperative/src/plugins/cmd/list/list.handler.ts
t1m0thyj: It looks like the `plugins uninstall` and `plugins update` commands assume their positional argument is present as a key in the plugins JSON file. So using `pluginName` in this case which is also a key in the plugins JSON file makes sense.

awharn (approved):

PR #748, opened 2022-03-07
Update creds

### LINKED ISSUE #738: [next] The `ProfileInfo.mergeArgsForProfile` should not return user-defined secure fields
Seems like `mergeArgsForProfile` is giving preference to the schema files instead of the user-defined secure properties.

**Current behavior:**
Calling `ProfileInfo.mergeArgsForProfile( someProfile, {getSecureVals: false} )` will return the `host` property given the config below.
```json
{
    "profiles": {
        "lpar1": {
            "properties": { "rejectUnauthorized": false },
            "profiles": {
                "zosmf": { "type": "zosmf", "properties": { "port": 449 } }
            },
            "secure": [ "host", "user", "password" ]
        }
    },
    "defaults": { "zosmf": "lpar1.zosmf" },
    "autoStore": true
}
```

**Expected behavior:**
Check for `config.api.secure.secureFields()` before assigning the `argValue`  property.

### LINKED ISSUE #739: [next] The `IProfArgAttrs.secure` property does not consider the user-defined secure properties
If the user added a property in the secure array, that property should be treated as secure by any APIs referencing the `IProfArgAttrs.secure` proeprty

### LINKED ISSUE #755: [next] The `ProfileInfo.mergeArgsForProfile` should return secure fields defined in the schema if the user removed it from the secure array
This is the reverse situation of:
- #738

**Current behavior:**
Calling `ProfileInfo.mergeArgsForProfile( someProfile, {getSecureVals: false} )` will **not** return the `user` property given the config below.
```json
{
    "profiles": {
        "lpar1": {
            "properties": { "rejectUnauthorized": false },
            "profiles": {
                "zosmf": { "type": "zosmf", "properties": { "user": "zoweUser", "port": 449 } }
            },
            "secure": [ "password" ]
        }
    },
    "defaults": { "zosmf": "lpar1.zosmf" },
    "autoStore": true
}
```

**Expected behavior:**
Return all user-defined plain-text fields when `getSecureVals` is false

### LINKED ISSUE #758: [next] The `osLoc` array causes the loadSchema function to return `null` when there is a user config present
Upon further investigation, I found what seems like a bug in `loadSchema` in these lines:

Since the `profile.profLoc.osLoc` contains an array of 2 items, the `schemaMapKey` looks like:
image

and because `mProfileSchemaCache` looks like:

it is not able to find the correct schema and returns null. This causes the error message.

_Originally posted by @crawr in https://github.com/zowe/vscode-extension-for-zowe/issues/1620#issuecomment-1069007383_

zFernand0:
This PR:
- Fix #738 by removing user-defined secure properties unless requested by `getSecureValues: true`
- Fix #739 by searching the user-defined secure properties in addition to the schema definition
- Fix #755 by specifying `secure: false` for plain-text properties in the config.json file.
- Fix #758 by selecting the first `osLoc` in the array. 
- Introduce `updateKnownProperty` which will update a given property in most cases and `resolve(false)` if it wasn't able to do so. _This applies to v1 profiles as well._
- Introduce `updateProperty` which takes care of special cases where the property is nowhere to be found in any of the config layer. _This applies to v1 profiles as well._
- Add support for removing properties is they are updated to `undefined` . _This applies to v1 profiles as well._
- Removes the strict requirement of `IHandlerParameter` from the `ConfigAutoStore` class by implementing helper method, which led to either deprecating(?) a few of them
- This issue may also be fixed: https://github.com/zowe/zowe-cli/issues/1292

To Do:
- [x] Outline and test the many scenarios that can occur 
- [x] Add proper JSDoc to helper functions introduced
- [x] Figure out how to remove the dependency on the schema definition
- Decide on the following:
  - [x] Should we allow new/unkown properties to be added ?
    - Yes
  - [x] Should we allow existing properties to be deleted ?
    - Yes
  - [x] Should we have a `setSecure` flag that allows to change a property from secure to unsecure (and vice versa) ? 
    - Yes
- [x] (v1 only) Update  the state of `mergedArgs` (mLoadedConfig) without calling `readProfilesFromDisk`
- [x] (v2 only) Implement `setSecure` parameter (following the `config.set` behavior)
- [x] Clean up and mark as ready for review


packages/config/src/ProfileInfo.ts
```diff
@@ -524,19 +641,54 @@ export class ProfileInfo {
...
+                            argName: propName,
+                            dataType: this.argDataType(propInfoInSchema.type),
+                            argValue: (propInfoInSchema as ICommandProfileProperty).optionDefinition?.defaultValue,
+                            argLoc: { locType: ProfLocType.DEFAULT },
+                            secure: propInfoInSchema.secure
+                        });
+                    }
                 } else {
-                    knownArg.secure = propObj.secure;
+                    knownArg.secure = knownArg.secure ?? propInfoInSchema.secure;
```
zFernand0, line 691: changing this line to `knownArg.secure = knownArg.secure || propInfoInSchema.secure;` will revert:
- #755

zFernand0: After some discussion with both CLI and ZE squads...
Since the user went out of their way to make the property insecure, the API should respect that and return schema-defined secure properties (stored in plain-text by the user) when requested with `getSecureVals: false`.

gejohnston (approved):

packages/config/src/ProfileInfo.ts
```diff
@@ -179,6 +181,116 @@ export class ProfileInfo {
...
+    }
+
+    /**
+     * Update a given property with the value provided.
+     * This function only works for properties that can be found in the config files (including secure arrays).
+     * If the property cannot be found, this function will resolve to false
+     * This function supports v1 profiles
+     * @param options Set of options required to update a known property
+     */
+    public async updateKnownProperty(options: IProfInfoUpdateKnownPropOpts): Promise<boolean> {
```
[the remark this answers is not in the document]
zFernand0: The intention was to provide the means for developers to use the `mergedArgs` they already had to update properties, and if it resolves to `false`, then they know the property was not updated, and they may need to call `updateProperty` with more details.

But no request was made to be able to do it using the `mergedArgs` as a starting point.

packages/config/src/doc/IProfArgAttrs.ts
```diff
@@ -11,6 +11,12 @@
 
 import { IProfLoc } from "./IProfLoc";
 
+/** The type of data for this property */
+export type IProfDataType = "string" | "number" | "boolean" | "array" | "object";
+
+/** The value for the argument */
+export type IProfArgValue = string | number | boolean | string[] | object;
```
t1m0thyj, line 18: Thanks for making this a separate type 🙂 Curious why we chose `string[]` rather than `any[]` for the array type here. Is it because the CLI always parses each item in an array arguments as a string?

packages/config/src/ProfileInfo.ts
```diff
@@ -179,6 +181,116 @@ export class ProfileInfo {
...
+            return false;
+        }
+
+        switch (toUpdate.argLoc.locType) {
+            case ProfLocType.OLD_PROFILE: {
+                const filePath = toUpdate.argLoc.osLoc;
+                const profileName = ProfileIO.fileToProfileName(filePath[0], "." + filePath[0].split(".").slice(-1)[0]);
+                const profileType = filePath[0].substring(this.mOldSchoolProfileRootDir.length + 1).split("/")[0];
+                const profMgr = new CliProfileManager({ profileRootDirectory: this.mOldSchoolProfileRootDir, type: profileType });
+                if (options.value != null) {
```
t1m0thyj, line 260: Wondering if we ought to allow folks to set properties to `null` in a profile. I believe I have seen this done before in v1 profiles. Should we check if `options.value !== undefined` here instead?

packages/config/src/ConfigUtils.ts
```diff
@@ -38,12 +38,12 @@ export function coercePropValue(value: any) {
...
  * @returns The profile name
  */
-export function getActiveProfileName(profileType: string, cmdArguments: ICommandArguments, defaultProfileName?: string): string {
+export function getActiveProfileName(profileType: string, cmdArguments?: ICommandArguments, defaultProfileName?: string): string {
     // Look for profile name first in command line arguments, second in
     // default profiles defined in config, and finally fall back to using
     // the profile type as the profile name.
-    return cmdArguments[`${profileType}-profile`] ||
-        ImperativeConfig.instance.config.properties.defaults[profileType] ||
+    return (cmdArguments && cmdArguments[`${profileType}-profile`]) ||
```
t1m0thyj, line 45: No change required here. Although I believe ``cmdArguments?.[`${profileType}-profile`]`` is possible if you like to use short-hand form 😛

packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
```diff
@@ -1096,19 +889,21 @@ describe("ProfileInfo tests", () => {
             const mergedArgs = profInfo.mergeArgsForProfile(profAttrs);
 
             const userArg = mergedArgs.knownArgs.find((arg) => arg.argName === "user");
-            expect(userArg.argValue).toBeUndefined();
+            // expect(userArg.argValue).toBeUndefined(); // See zowe/imperative#755
+            expect(userArg.argValue).toBe("userNameBase"); // See zowe/imperative#755
```
t1m0thyj, line 893: Since this PR fixes #755, should this change be reverted? The same comment applies for `passwordArg` below.

packages/config/__tests__/ProfileInfo.OldProfiles.test.ts
```diff
@@ -0,0 +1,481 @@
...
+    afterAll(() => {
+        // ensure that jest reports go to the right place
+        process.chdir(origDir);
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+    });
+
+    describe("Old-school Profile Tests", () => {
```
t1m0thyj, line 60: Thanks for splitting the old-school and team config tests into 2 separate files 👍 Is this `describe` block still necessary, or is it an extra layer that can be removed if we rename the top-level `describe` to "Old-school Profile Tests"?

packages/config/__tests__/ProfileInfo.OldProfiles.test.ts
```diff
@@ -0,0 +1,481 @@
...
+
+        describe("getDefaultProfile", () => {
+
+            afterEach(() => {
+                jest.clearAllMocks();
+            });
+
+            it("should return null if no default for that type exists 1: oldSchool", async () => {
```
t1m0thyj, line 68: The "oldSchool" at the end of each test name seems redundant now that all tests in this file are for old-school profiles. No change required here, just an observation and sorry for being picky 😢

packages/config/src/doc/IProfArgAttrs.ts
zFernand0: I believe so, but will change it to `any[]` to be safe here

packages/config/__tests__/ProfileInfo.OldProfiles.test.ts
zFernand0: will do the same for teamConfig on the other file 😋

packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
zFernand0: I think the userNameBase (as well as the password) should be expected to be returned with `getSecureVals: false|undefined` since they are in plain text in the `__tests__/__resources__/ProfInfoApp_team_config_proj/ProfInfoApp.config.json`.

I could try to add one additional tests with a new project config that specifies them in the secure array.

packages/config/src/doc/IProfArgAttrs.ts
t1m0thyj: Is changing this potentially a v2-v2 breaking change? For anyone who was depending on the type being `string[]`

packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
t1m0thyj: No need to add additional test, maybe its ok to just remove the comments from this file? I assumed they were TODO-related, but I also don't mind if you want to keep the reference to the issue 🙂

packages/config/src/doc/IProfArgAttrs.ts
zFernand0: After looking into this a bit more...
Its main use was for getting arrays from ENV variables. so they will always be strings.
So, maybe it's fine to keep it as `string[]`

packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
zFernand0: I don't feel strongly about keeping the issue number there. 😋

packages/config/__tests__/__resources__/ProfInfoApp_user_and_team_config_proj/ProfInfoApp.schema.json
```diff
@@ -0,0 +1,186 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$version": 2,
```
t1m0thyj, line 3: This schema will need to be updated in #751. Nothing to change in this PR, just a note for my own future reference 🙂

t1m0thyj (approved):
LGTM, thanks for making so many enhancements and fixes to the ProfileInfo API 🙂

PR #759, opened 2022-03-21
Fix compressed error messages returned by AbstractRestClient

t1m0thyj:
Backports #756 from `next` to `zowe-v1-lts`


zFernand0 (approved):
thanks for porting #756

awharn (approved):

PR #760, opened 2022-03-21
Fix Keytar not found outside top-level node_modules folder

t1m0thyj:
This is part of the fix for https://github.com/zowe/zowe-cli/issues/1314.

The remainder of the fix is https://github.com/zowe/zowe-cli/pull/1347 (will need to be ported to `next` branch).

Once both these fixes are published, it should be possible to `import { imperative } from "@zowe/cli"` and use the ProfileInfo API from the shrinkwrapped CLI package without any duplicated Imperative or Keytar dependencies.

### DIFF
--- a/packages/security/src/DefaultCredentialManager.ts
+++ b/packages/security/src/DefaultCredentialManager.ts
@@ -121,7 +121,7 @@ export class DefaultCredentialManager extends AbstractCredentialManager {
             // within our caller's path.
             const requireOpts: any = {};
             if (process.mainModule?.filename != null) {
-                requireOpts.paths = [process.mainModule.filename];
+                requireOpts.paths = [process.mainModule.filename, ...require.resolve.paths("keytar")];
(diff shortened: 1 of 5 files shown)


packages/security/src/DefaultCredentialManager.ts
```diff
@@ -121,7 +121,7 @@ export class DefaultCredentialManager extends AbstractCredentialManager {
             // within our caller's path.
             const requireOpts: any = {};
             if (process.mainModule?.filename != null) {
-                requireOpts.paths = [process.mainModule.filename];
+                requireOpts.paths = [process.mainModule.filename, ...require.resolve.paths("keytar")];
```
zFernand0, line 124: thanks for providing a meaningful string to the `.paths()` method 😋

zFernand0 (approved):

awharn (approved):

PR #765, opened 2022-03-29
Check teamConfig file to determine if isSecured or not

zFernand0:


phaumer:
Thanks for adding that test @zFernand0. Another part of the problem that will make this new test fail is that the `zowe config init` will automatically add secured fields. Do we need another parameter for that command to prevent that?

gejohnston (approved):
This provides useful information to the caller.

The challenge for the caller will be what to do when it receives an answer of 'secure', when it wants to run 'insecure'.

packages/config/src/ProfileCredentials.ts
```diff
@@ -86,6 +85,16 @@ export class ProfileCredentials {
         }
     }
 
+    /**
+     * Check whether a teamConfig is secure or not
+     * @returns False if not using teamConfig
```
t1m0thyj, line 90: Should the jsdoc be updated?

t1m0thyj (approved):

PR #768, opened 2022-03-30
Check aliases for values

zFernand0:
Enhanced backward compatibility with v1 profiles by checking the aliases for profile values and not just the option name.

Limitation:
- V1 environments variables (e.g. `ZOWE_OPT_USERNAME`) are not recognized and thus won't take precedence.

**Question:**
- Should we support V1 environmental variables based on aliases ?
  - **Answer:** We will not be checking for `ZOWE_OPT_<alias>` for values
    - Imagine someone defined `ZOWE_OPT_U=<unicode-string-with-alias-U-required-by-a-plugin>`. We could accidently be using that as the username since U might be a conflicting alias

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
+- Enhancement: Added support for profile name aliases in team config so that `--zosmf-profile lpar1` falls back to profile "zosmf_lpar1" if "lpar1" does not exist.
+- BugFix: Reworded potentially misleading output of `config convert-profiles` command mentioning obsolete plug-ins.
+- BugFix: Made `--dry-run` and `--prompt` options mutually exclusive on `config init` command.
+- **Next Breaking**: The team config API method `config.api.profiles.get` now returns `null` if a profile doesn't exist unless `mustExist` is false. [#518](https://github.com/zowe/imperative/issues/518)
 - BugFix: Added the ability to read option values from aliases. Enhanced backward compatibility with V1 profiles. [#770](https://github.com/zowe/imperative/issues/770)
 
 ## `5.0.0-next.202203311701`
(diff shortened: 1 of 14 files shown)


packages/utilities/src/CliUtils.ts
```diff
@@ -143,9 +143,20 @@ export class CliUtils {
...
                 // For each option - extract the value if that exact property exists
                 options.forEach((opt) => {
-
                     // Get the camel an kebab case
-                    const cases = CliUtils.getOptionFormat(opt.name);
+                    let cases;
+                    if (profile[opt.name] == null) {
+                        // Use aliases for backwards compatibility
+                        // Search for first alias available in the profile
+                        if ("aliases" in opt) {
```
t1m0thyj, line 151: Should these nested `if` statements be combined into a single one?

packages/utilities/src/CliUtils.ts
```diff
@@ -210,8 +221,17 @@ export class CliUtils {
...
-            if (envValue != null) {
 
+            if (envValue == null) {
+                // Use aliases for backwards compatibility
+                // Search for first alias available in the process.env
+                if ("aliases" in opt) {
+                    const oldOption = opt.aliases.find(o => CliUtils.getEnvValForOption(envPrefix, o) != null);
+                    if (oldOption) envValue = CliUtils.getEnvValForOption(envPrefix, oldOption);
+                }
+            }
```
t1m0thyj, line 232: I would vote not to check aliases for environment variables, since it could lead to conflicts as you mentioned, and I believe the enhancement request we received was specifically for v1 profiles. Curious what others' thoughts are? 🙂

packages/utilities/src/CliUtils.ts
zFernand0: combined both for now 😋

packages/utilities/src/CliUtils.ts
zFernand0: Will proceed with not lookin at ENV vars 😋

t1m0thyj (approved):

gejohnston (approved):

zFernand0:
It seems that we got confirmation from the requester that this PR fixes the issue. 😋

PR #773, opened 2022-04-04
[next] Fix `config set` not respecting property types

### LINKED ISSUE #772: zowe config set "profiles.profilename.properties.propertyname" "12345" always defines numeric value regardless of what is defined in the zowe.schema.json

t1m0thyj:
Resolves #772

I've attempted to make the new method `ConfigSchema.findPropertyType` flexible and easy to use not only in the CLI, but also in apps that use the ProfileInfo API like Zowe Explorer.

### DIFF
--- a/packages/config/src/ConfigSchema.ts
+++ b/packages/config/src/ConfigSchema.ts
@@ -436,7 +436,7 @@ export class ConfigSchema {
 
         const pathSegments = path.split(".");
         const propertyName = pathSegments.pop();
-        const profilePath = pathSegments.slice(0, -1).join(".");  // eslint-disable-line @typescript-eslint/no-magic-numbers
+        const profilePath = pathSegments.slice(0, -1).join(".");
         const profileType: string = lodash.get(config, `${profilePath}.type`);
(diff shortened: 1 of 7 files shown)


packages/config/src/ConfigSchema.ts
```diff
@@ -420,4 +422,34 @@ export class ConfigSchema {
...
+
+        const profileSchemas = schema ? this.loadSchema(schema) : ImperativeConfig.instance.loadedConfig.profiles;
+        const profileSchema = profileSchemas.find(p => p.type === profileType)?.schema;
+        if (profileSchema?.properties[propertyName] == null) {
+            return;
+        }
+
+        // TODO How to handle profile property with multiple types
+        const propertyType = profileSchema.properties[propertyName].type;
+        return Array.isArray(propertyType) ? propertyType[0] : propertyType;
```
zFernand0, line 453: good question 😋

zFernand0 (approved):
LGTM

awharn (approved):

PR #778, opened 2022-04-11
Update schema version for Zowe v2

t1m0thyj:
I believe we wanted to make this change before the Zowe v2 code freeze, not sure what to put in the changelog though.

### DIFF
--- a/packages/config/src/ConfigSchema.ts
+++ b/packages/config/src/ConfigSchema.ts
@@ -34,7 +34,7 @@ export class ConfigSchema {
      * @readonly
      * @memberof ConfigSchema
      */
-    private static readonly SCHEMA_VERSION = 0.4;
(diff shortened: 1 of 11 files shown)


zFernand0 (approved):
Thanks for update it 😋

packages/config/src/ConfigSchema.ts
```diff
@@ -34,7 +34,7 @@ export class ConfigSchema {
      * @readonly
      * @memberof ConfigSchema
      */
-    private static readonly SCHEMA_VERSION = 0.4;
+    private static readonly SCHEMA_VERSION = 1.0;
```
gejohnston, line 37: The team decided that we should make the version a string of "1.0".

gejohnston (changes requested):

gejohnston (approved):

awharn (approved):

PR #779, opened 2022-04-12
Fix merging of autoStore property between layers

t1m0thyj:
Fixes incorrect behavior where `autoStore=true` in global config overrides `autoStore=false` in project config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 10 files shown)


KevinLoesch1 (approved):

awharn (approved):

zFernand0 (approved):
I believe this fixes some unexpected autoStore behavior from a ZE perspective 😋 
A follow up update may be needed on the ProfileInfo APIs if autoStore for V1 profiles is not desired in ZE though 😅  (see https://github.com/zowe/vscode-extension-for-zowe/issues/1738)

PR #783, opened 2022-04-12
Remove moment dependency

t1m0thyj:
The developers of Moment.js recommend not to use it anymore: https://momentjs.com/docs/#/-project-status/

Since the package had a recent vulnerability, we only used it in one place, and it was easy to replace with native JS, I've done so in this PR.

Also added a test to verify the timestamp is correct, since the original format was wrong. It was printing the month in place of minutes, and 100ths of seconds instead of 60ths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Removed `moment` dependency.
+
 ## `4.18.2`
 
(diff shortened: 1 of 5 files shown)


zFernand0 (approved):
It'd be nice if we could get rid of moment!
Wondering if it makes sense to expose the buildPrefix function 😋

awharn (approved):

KevinLoesch1 (approved):

gejohnston (approved):

PR #785, opened 2022-04-12
Add osLoc for global profiles in `ProfileInfo` APIs

zFernand0:
In this PR:
- Fix #771

PR #786, opened 2022-04-12
Implement `ExcludeHomeDir` for `ProfileInfo.getAllProfiles`

### LINKED ISSUE #787: `ProfileInfo.getAllProfiles` should allow developers to exclude the global layer (homeDir)
see title

zFernand0:
⚠️ This PR build on top of #785 ⚠️ 

In this PR:
- Fix #787


awharn (approved):

gejohnston (approved):

packages/config/src/Config.ts
```diff
@@ -405,6 +405,15 @@ export class Config {
...
     }
 
+    /**
+     * Helper function to get masked properties and exclude the global layer
+     * @param excludeGlobal Indicates whether or not to exclude the global layer
+     * @returns Same as this.maskedProperties
+     */
+    public getMaskedProperties(excludeGlobal?: boolean): IConfig {
+        return this.layerMerge(true, excludeGlobal);
+    }
```
t1m0thyj: Instead of adding this method, we may be able to make `layerMerge` a public internal method as we discussed 🙂

packages/config/src/doc/IProfiInfoProps.ts
```diff
@@ -0,0 +1,29 @@
+/*
```
t1m0thyj, line 1: Should the file name be `IProfileInfoProps.ts`?

packages/config/src/doc/IProfiInfoProps.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Options that will affect the behavior of the ProfileInfo class.
+ * They are supplied on the ProfileInfo getAllProfiles method.
+ */
+ export interface IGetAllProfilesOptions {
+
+  /**
+   * Indicates if the home directory should be excluded
+   * The default is false.
+   */
+   excludeHomeDir?: boolean;
```
t1m0thyj, line 22: Passing `excludeHomeDir` on the getAllProfiles method makes sense if it meets the use case 👍 Another option we could consider is an enum option on the readProfilesFromDisk method to `"ignore" | "fallback" | "merge"` the home dir.

t1m0thyj: Discussed this idea and decided against it, as it seems unnecessary and could be confusing. As long as we maintain the desired behavior of the ProfileInfo API in Zowe Explorer, I'm good with the current implementation 🙂

packages/config/src/Config.ts
zFernand0: done! 😋

packages/config/src/doc/IProfiInfoProps.ts
zFernand0: I will consider these sort of comments whenever we need to implement some sort of binary implementation incase there is more than just two routes we can take 👍 
Enums are cool too 😋

packages/config/src/doc/IProfiInfoProps.ts
zFernand0: renamed to `IProfInfoProps.ts` since all the other files are abbreviated like that 😋

zFernand0:
Merging this PR into `fix-771` since all the pipelines will also run over there.

That way, I can address the 1 TODO comment on #785 about adding additional unit tests for the new API `getOSLocInfo`.

PR #788, opened 2022-04-13
Design Discussion - `osLoc` with more information

zFernand0:
⚠️ Do not merge ⚠️ 
Also, no need to approve this PR

**What ?**
- Enhance (break) the osLoc property to include more information about which layer each of the config paths belongs to.

**Why ?**
- To allow application developers to exclude global team profiles.

**Alternatives ?**
- Non-breaking is to implement an utility function in ProfileInfo that returns layer information based profile attributes

**Pros and Cons ?**
- Pro: More information available for future (currently unknown) use cases
- Con: Significant braking change to the API

### DIFF
--- a/packages/config/__tests__/ProfileInfo.OldProfiles.test.ts
+++ b/packages/config/__tests__/ProfileInfo.OldProfiles.test.ts
@@ -107,7 +107,7 @@ describe("Old-school ProfileInfo tests", () => {
             expect(profAttrs.profType).toBe(desiredProfType);
             expect(profAttrs.profLoc.locType).not.toBeNull();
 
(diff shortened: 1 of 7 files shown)


zFernand0:
I tend to agree with Gene's comment above.
Here is the utility function as an alternative (already in the fix-771 branch)
https://github.com/zowe/imperative/blob/f613e291008c819907dc0cda32bb9b86a77f1ecf/packages/config/src/ProfileInfo.ts#L901-L917

zFernand0:
Let's move on with adding a new API.
Thanks for the quick feedback (in the PR and in person) 🥳 

We can consider doing something like this for Zowe v3 😋 
Closing for now.

PR #796, opened 2022-04-20
Use base from operating layer

zFernand0:
Since there is already a feature merged into `next`, I believe I can switch the base to `master` and have this PR go out as a 5.0.1 (if needed)
I don't think the feature in next will cause any harm to the ProfileInfo APIs, so they could both be added in a 5.1.0.
Please keep in mind that ZE needs this for Zowe RC2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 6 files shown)


t1m0thyj (approved):

t1m0thyj:
Since this a bug fix needed by ZE, I vote that we update this PR to target "master" and publish it as 5.0.1 😛

PR #798, opened 2022-04-20
Fix defect found when attempting to create a V1 profile with no secure properties

awharn:


t1m0thyj (approved):

packages/cmd/src/profiles/CliProfileManager.ts
```diff
@@ -405,18 +405,23 @@ export class CliProfileManager extends BasicProfileManager<ICommandProfileTypeCo
              * @return {Promise<string>}
              */
             securelyStoreValue = async (propertyNamePath: string, propertyValue: string): Promise<string> => {
-                try {
-                    if (isNullOrUndefined(propertyValue)) {
-                        // don't store null values but still remove value that may have been stored previously
-                        this.log.debug(`Deleting secured field with key ${propertyNamePath}` +
-                            ` for profile (of type "${this.profileType}").`);
+                if (propertyValue == null) {
+                    // don't store null values but still remove value that may have been stored previously
```
zFernand0, line 409: isn't this a breaking change ?
I believe v1 profiles (and in team config too) the user was able to store null in the profiles.
Maybe I'm remembering wrong though.

If this is unrelated, please feel free to ignore 😋

awharn: That comment and logic was in place before. The only change that was made in this PR is that the credential manager will not throw an error now if it cannot delete the property from the keyring. That shouldn't be a breaking change.

gejohnston (approved):

zFernand0 (approved):

PR #799, opened 2022-04-22
Merge `--show-inputs-only` flag into master branch

t1m0thyj:
⚠️ Do not merge until Zowe v2 is published on zowe.org.

#721 was merged into next branch after the Zowe v2 code freeze. This PR ports it to the master branch to be published.


gejohnston:
I agree with Fernando. Everything looks good.

PR #800, opened 2022-04-22
Make private function `argTeamConfigLoc` aware of multi-layer operations

zFernand0:
This PR intends to change the private function signature of `argTeamConfigLoc` and make it aware of multi-layer operations.

_What does `multi-layer` operation mean?_
Operating against a project layer and a global layer simultaneously. 😋 

Example operation: 
- Getting argument location information from the `mergeArgsForProfile` method


packages/config/src/ProfileInfo.ts
```diff
@@ -1268,30 +1271,30 @@ export class ProfileInfo {
      * @param propName Name of a team config property (e.g., host)
```
gejohnston, line 1273: Does this parameter documentation need to be updated?

t1m0thyj (commented):
LGTM once lint errors are resolved 🙂

gejohnston (approved):

t1m0thyj (approved):

awharn (approved):

PR #801, opened 2022-04-26
Port #768 back to v1-lts

zFernand0:
In this PR:
- Port #768 

I recommend reviewing with whitespaces hidden 😋 
https://github.com/zowe/imperative/pull/801/files?w=1

### DIFF
--- a/.github/workflows/imperative.yml
+++ b/.github/workflows/imperative.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Unit Tests
       id: unit
       if: ${{ always() && steps.build.outcome == 'success' }}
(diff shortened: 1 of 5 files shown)


gejohnston (approved):

KevinLoesch1 (approved):

t1m0thyj:
Thanks for backporting this! Would you mind also backporting the changes from 29cf7638a4254cc34d574b6137de8b935454a60f to get tests passing on Node.js 12? Alternatively, I don't know if we should stop caring about Node 12 since it reaches EoL in less than a week. But we did make the fix in the master branch 🙂

t1m0thyj (approved):

PR #815, opened 2022-05-04
Fix `config secure` not respecting rejectUnauthorized

t1m0thyj:
When `config secure` prompts for user and password to obtain a token, it was not respecting the `rejectUnauthorized` property on the associated profile.

To fix this we need access to the `BaseAuthHandler.createSessCfgFromArgs` method. I've deprecated `getPromptParams` and replaced it with `getAuthHandlerApi` to expose this method. This works but not sure if it is the best implementation, so opening this PR in draft mode to get feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed `config secure` not respecting the `rejectUnauthorized` property in team config. [#813](https://github.com/zowe/imperative/issues/813)
+
 ## `5.1.0`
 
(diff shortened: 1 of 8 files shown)


gejohnston:
These changes look good. It seems like a lot of work to get access to the rejectUnauthorized value. This is probably the best available approach. The original functions were created with a rather awkward implementation, which was required by the need to add automatic prompting in V1 without making any breaking changes.

zFernand0 (approved):
I know this is still in draft mode 😋 
However, I do believe I would've done the exact same thing here.

gejohnston (approved):

awharn (approved):

Issue #764, opened 2022-03-29
Poor performance formatting output


dronky:
Hello,
We use [format()](https://github.com/zowe/imperative/blob/4926f56baf27fb1e36c797f26a7dcad79009bc83/packages/cmd/src/response/CommandResponse.ts#L218) function in our Endevor plugin to format output from `endevor list elements` command. It works perfect for small amount of data, but when we try to list about 100000 elements (2000000 rows), format() function takes more than 1 hour.

Do we have a lightweight alternative or could we optimize it somehow?

t1m0thyj:
I've attempted to reproduce this issue with the following code in a command handler:
```
    const outputData = [];
    for (let i = 0; i < 1e6; i++) {
        outputData.push({ a: 0, b: 1, c: 2, d: 3, e: 4, f: 5 });
    }
    params.response.format.output({ output: outputData, format: "table", header: true });
```

The command takes a while to run and after a few minutes I get the error `Maximum call stack size exceeded`. This seems like different behavior from what you observed, but confirms that `format.output` does not handle large data well.

I believe we're limited here by the `cli-table3` package which Imperative uses to output data in a tabular format. The devs of that package seem to be aware of performance issues and are working to significantly decrease rendering time: https://github.com/cli-table/cli-table3/issues/68#issuecomment-1080964304

dronky:
> I've attempted to reproduce this issue with the following code in a command handler:
> 
> ```
>     const outputData = [];
>     for (let i = 0; i < 1e6; i++) {
>         outputData.push({ a: 0, b: 1, c: 2, d: 3, e: 4, f: 5 });
>     }
>     params.response.format.output({ output: outputData, format: "table", header: true });
> ```
> 
> The command takes a while to run and after a few minutes I get the error `Maximum call stack size exceeded`. This seems like different behavior from what you observed, but confirms that `format.output` does not handle large data well.
> 
> I believe we're limited here by the `cli-table3` package which Imperative uses to output data in a tabular format. The devs of that package seem to be aware of performance issues and are working to significantly decrease rendering time: [cli-table/cli-table3#68 (comment)](https://github.com/cli-table/cli-table3/issues/68#issuecomment-1080964304)

Thank you for the info Timothy, hope it will be fixed soon.

dronky:
There is a 0.6.2 version available. 
Just wondering is it possible to update it in both (v1-lts and v2-lts) versions?

dronky:
Thank you for updating it Timothy. 
[...]

PR #819, opened 2022-05-10
Update cli-table3 and add Node 18 to CI

t1m0thyj:
* Update `cli-table3` dependency as requested in https://github.com/zowe/imperative/issues/764#issuecomment-1122121660
* Drop Node 12 (EoL) from our CI matrix and replace it with Node 18

### DIFF
--- a/.github/workflows/imperative.yml
+++ b/.github/workflows/imperative.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
         os: [windows-latest, ubuntu-latest, macos-latest]
 
     env:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Updated `cli-table3` dependency for performance improvements.
+
 ## `5.2.2`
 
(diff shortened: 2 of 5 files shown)


gejohnston (approved):
Just curious - did you see a performance improvement after the upgrade to cli-table3 ?

zFernand0 (approved):

t1m0thyj:
Node 18 builds are passing on Windows and macOS, but currently all Ubuntu builds are failing because the Azure Ubuntu package repository seems broken: http://azure.archive.ubuntu.com/ubuntu/

I've also noticed that Jest seems to hang for a while after running unit tests. The unit tests take <3 minutes to run, but the build step takes 17 minutes to run. Investigating why Jenkins doesn't have this issue but GHA does.

t1m0thyj:
> Node 18 builds are passing on Windows and macOS, but currently all Ubuntu builds are failing because the Azure Ubuntu package repository seems broken: http://azure.archive.ubuntu.com/ubuntu/
> 
> I've also noticed that Jest seems to hang for a while after running unit tests. The unit tests take <3 minutes to run, but the build step takes 17 minutes to run. Investigating why Jenkins doesn't have this issue but GHA does.

The Ubuntu package repository wasn't broken, there was a bug in the unlock-keyring action now fixed in https://github.com/t1m0thyj/unlock-keyring/pull/2. The unit test suite that was causing Jest to hang should also be fixed in https://github.com/zowe/imperative/pull/819/commits/ec07bfe047e60a016bbba943b2206216352f9b5d.

awharn (approved):

PR #822, opened 2022-05-11
Fix `config init` discarding values that it prompts for

### LINKED ISSUE #821: `config init` discards values that it prompts for
**Steps**
* Run `zowe config init --no-prompt` so that `profiles.base.properties.host` is an empty string
* Run `zowe config init` and enter a value for host when prompted

**Actual Behavior**
The `host` property in zowe.config.json is still an empty string

**Expected Behavior**
The `host` property in zowe.config.json is the value I entered when prompted

t1m0thyj:
Fixes #821

### DIFF
--- a/packages/config/src/ConfigBuilder.ts
+++ b/packages/config/src/ConfigBuilder.ts
@@ -64,8 +64,7 @@ export class ConfigBuilder {
         // Prompt for properties missing from base profile
         if (impConfig.baseProfile != null && opts.getValueBack != null) {
             for (const [k, v] of Object.entries(impConfig.baseProfile.schema.properties)) {
-                const oldPropValue = opts.mergeConfig?.profiles[impConfig.baseProfile.type]?.properties[k];
-                if (v.includeInTemplate && v.optionDefinition?.defaultValue == null && (oldPropValue == null || oldPropValue === "")) {
+                if (v.includeInTemplate && v.optionDefinition?.defaultValue == null) {
                     const propValue = await opts.getValueBack(k, v);
                     if (propValue != null) {
                         lodash.set(config, `profiles.${impConfig.baseProfile.type}.properties.${k}`, propValue);
--- a/packages/config/src/ConfigBuilder.ts
+++ b/packages/config/src/ConfigBuilder.ts
@@ -64,7 +64,8 @@ export class ConfigBuilder {
         // Prompt for properties missing from base profile
         if (impConfig.baseProfile != null && opts.getValueBack != null) {
             for (const [k, v] of Object.entries(impConfig.baseProfile.schema.properties)) {
-                if (v.includeInTemplate && v.optionDefinition?.defaultValue == null) {
+                const oldPropValue = opts.mergeConfig?.profiles[impConfig.baseProfile.type]?.properties[k];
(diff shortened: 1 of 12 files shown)


gejohnston (approved):

t1m0thyj:
Question for reviewers of this PR @awharn @zFernand0 @gejohnston @KevinLoesch1:

In the case that `profiles.base.properties.host` is already defined in zowe.config.json and you run `zowe config init`, what should happen?
    :rocket: - it should skip prompting for host since it is already defined
    :eyes: - it should prompt for host and overwrite the existing value if a new one is entered

Note: I'm not proposing a change to the config merge API with either approach. The question only pertains to how `zowe config init` should behave with prompting for values.

KevinLoesch1 (approved):

awharn (approved):

packages/config/src/ConfigBuilder.ts
```diff
@@ -64,7 +64,8 @@ export class ConfigBuilder {
         // Prompt for properties missing from base profile
         if (impConfig.baseProfile != null && opts.getValueBack != null) {
             for (const [k, v] of Object.entries(impConfig.baseProfile.schema.properties)) {
-                if (v.includeInTemplate && v.optionDefinition?.defaultValue == null) {
+                const oldPropValue = opts.mergeConfig?.profiles[impConfig.baseProfile.type]?.properties[k];
+                if (v.includeInTemplate && v.optionDefinition?.defaultValue == null && (oldPropValue == null || oldPropValue === "")) {
```
zFernand0, line 68: should we `.trim()` the `oldPropValue` when comparing `=== ""`  to account for `"    "` ? 😋

packages/imperative/src/config/cmd/init/init.handler.ts
```diff
@@ -121,12 +121,23 @@ export default class InitHandler implements ICommandHandler {
...
 
         // Build new config and merge with existing layer or overwrite it if overwrite & forSure options are present
         const newConfig: IConfig = await ConfigBuilder.build(ImperativeConfig.instance.loadedConfig, opts);
         if (overwrite) {
             config.api.layers.set(newConfig);
         } else {
+            if (opts.mergeConfig.profiles.base?.properties != null) {
+                // Remove null or empty values that should be overwritten from old base profile
+                for (const [k, v] of Object.entries(opts.mergeConfig.profiles.base.properties)) {
+                    if ((v === null || v === "") && newConfig.profiles.base.properties[k] != null) {
```
zFernand0, line 136: should we `v.trim()` when comparing `=== ""`  to account for `"    "` ? 😋

zFernand0 (approved):
The changes LGTM.
Left a minor comment in case it makes sense to account for whitespaces.

packages/config/src/ConfigBuilder.ts
t1m0thyj: I believe based on the team vote above, we don't need to check the old property value to see if it's null or empty, but will always overwrite with the prompted for value 🙂

t1m0thyj:
I think the team decision was a good one 🙂 It should make our prompting behavior more consistent, and I was able to remove ~50 lines of code since the ConfigBuilder API no longer needs any changes.

awharn (approved):

PR #823, opened 2022-05-12
Default `homeDir` to `cliHome` to support environmental variables

zFernand0:
By adding support for ENV variables, a few other issues were uncovered when `ZOWE_CLI_HOME` points to the same location as the project directory (`mHomeDir === mProjectDir`).
  - OsLoc information was getting duplicated in `.knownArgs`
  - `getOsLocInfo` duplicated layer information making it look like there were two of each layer available (`min: 4`, `max: 8`)
  - Tests that deleted `.properties.defaults.base` from the active layer, also needed to delete the global counterpart
    - There is a comment for each test that needed to be modified for this
  - Tests that relied on `getOsLocInfo` needed to be updated to account for the global copy of the layer

All issues mentioned above should be resolved in this PR 😋 

---

This PR is related to https://github.com/zowe/vscode-extension-for-zowe/issues/1777

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added environmental variable support to the ProfileInfo APIs by defaulting `homeDir` to `cliHome`. [zowe/vscode-extension-for-zowe#1777](https://github.com/zowe/vscode-extension-for-zowe/issues/1777)
+
 ## `5.2.2`
 
 - BugFix: Fixed `config secure` not respecting the `rejectUnauthorized` property in team config. [#813](https://github.com/zowe/imperative/issues/813)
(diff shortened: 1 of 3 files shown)


gejohnston (approved):

KevinLoesch1 (approved):

robertblum-psi:
The title states "Default `homeDir` to `cliHome` to support environmental variables".  Is there a default setting for this software already in end-user hands that use this setting.  If so, changing a default could be a breaking change.  Please confirm this is not a breaking change.

zFernand0:
> The title states "Default `homeDir` to `cliHome` to support environmental variables". Is there a default setting for this software already in end-user hands that use this setting. If so, changing a default could be a breaking change. Please confirm this is not a breaking change.

Hey @robertblum-psi,
Thank you for such a quick reaction and response to the PR 😋 
You make a good point about double-checking for breaking changes.

I want to make a few things clear:
- These change only affects the `ProfileInfo` APIs, which the `@zowe/cli` package does not use.
- The default will still be `~/.zowe/`.
- This behavior allows Zowe Explorer to behave more closely like the CLI by honoring the `ZOWE_CLI_HOME` (if set)

I will love to discuss more (preferably on a call) if we believe that the above may be considered a breaking change. 😋

robertblum-psi:
@zFernand0 I'm okay with your explanation.  I don't want the behavior of anyone's cli changing on them across imperative updates.  If you think that won't happen, I'm good with it.  Thanks for your quick reply.

t1m0thyj (approved):

PR #825, opened 2022-05-18
Fix `config init` discarding values that it prompts for (part 2)

t1m0thyj:
Resolves issues that were missed in #822 and observed during demo:
* `config init --no-prompt` resets credentials to empty strings
* `config init` resets hostname to empty string when prompt is skipped

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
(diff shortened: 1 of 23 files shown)


gejohnston (approved):

KevinLoesch1 (approved):

zFernand0 (approved):
Sorry for the delayed review
LGTM

PR #832, opened 2022-05-31
Fix bugs in ProfileInfo API

t1m0thyj:
Fixes a few bugs identified in the ProfileInfo API when using it in a VS Code extension.

It should be possible to call `ProfileInfo.readProfilesFromDisk` repeatedly in a VSCE with the following results:
* In an environment where no team config files exist and the `~/.zowe/profiles` directory does not exist, it should not fail.
* If team config files did not exist before, but another application created them, the config layers should be repopulated.
* If secure properties did not exist in team config before, but another application created them, they should be loaded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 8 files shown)


zFernand0 (approved):
LGTM!

gejohnston (approved):
Looks good.

PR #838, opened 2022-06-08
Quick fix for an NPM regression/issue

t1m0thyj:
Attempt 2 🙂

### DIFF
--- a/.github/workflows/imperative.yml
+++ b/.github/workflows/imperative.yml
@@ -13,7 +13,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


zFernand0 (approved):

KevinLoesch1 (approved):

t1m0thyj:
The issue was introduced in npm@8.11.0 and fixed in npm@8.12.1.

On Windows, updating or downgrading the npm version doesn't fix the issue (see https://github.com/npm/cli/issues/4989), so a better workaround for now seems to be locking down the Node.js versions.

PR #841, opened 2022-06-29
Fix line endings not always being normalized

t1m0thyj:
Fixes a bug in Imperative that caused zowe/zowe-cli#1458

This PR changes the default behavior of `IO.processNewlines` (when `lastByte` is null) to replace `\n` at the beginning of a string with `\r\n` on Windows. IMO this is a bug fix, not a breaking change.

### DIFF
--- a/packages/io/src/IO.ts
+++ b/packages/io/src/IO.ts
@@ -259,16 +259,21 @@ export class IO {
      * (basically, if the user is on Windows, change  \n to \r\n)
      * @static
      * @param {string} original - original input
+     * @param {number} lastByte - last byte of previous input, if it is being processed in chunks
      * @returns {string} - input with removed newlines
      * @memberof IO
      */
-    public static processNewlines(original: string): string {
+    public static processNewlines(original: string, lastByte?: number): string {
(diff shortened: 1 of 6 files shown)


packages/io/src/IO.ts
```diff
@@ -259,16 +259,21 @@ export class IO {
...
         if (os.platform() !== IO.OS_WIN32) {
             return original;
         }
         // otherwise, we're on windows
-        return original.replace(/([^\r])\n/g, "$1\r\n");
+        const processed = original.replace(/([^\r])\n/g, "$1\r\n");
+        if ((lastByte == null || lastByte !== "\r".charCodeAt(0)) && original.startsWith("\n")) {
+            return "\r" + processed;
+        }
+        return processed;
```
t1m0thyj, line 276: I'd welcome any suggestions for how to make this code more efficient, since it runs for every chunk of data that is downloaded by streamed requests. 😛

gejohnston (approved):
Looks good.

KevinLoesch1 (approved):

PR #844, opened 2022-06-29
Fix `--response-format-json` issues in daemon mode

### LINKED ISSUE #842: `zowe --version --rfj` returns an extra `\n` in the version field
```
$ zowe --version --rfj

{
  "success": true,
  "exitCode": 0,
  "message": "Version displayed",
  "stdout": "6.37.2\n",
  "stderr": "",
  "data": {
    "version": "6.37.2\n"
  }
}
```

The `version` field is not just version but it has an extra `\n` like in `6.37.2\n` in the example. It causes problems in automation that tries to ensure expected version but the workaround is easy.

t1m0thyj:
Fixes #842 and #843

### DIFF
--- a/packages/cmd/src/yargs/YargsConfigurer.ts
+++ b/packages/cmd/src/yargs/YargsConfigurer.ts
@@ -21,9 +21,9 @@ import { CommandUtils } from "../utils/CommandUtils";
 import { IProfileManagerFactory } from "../../../profiles";
 import { ICommandProfileTypeConfiguration } from "../doc/profiles/definition/ICommandProfileTypeConfiguration";
 import { IHelpGeneratorFactory } from "../help/doc/IHelpGeneratorFactory";
-import { CliUtils } from "../../../utilities/src/CliUtils";
 import { ImperativeConfig } from "../../../utilities";
 import { closest } from "fastest-levenshtein";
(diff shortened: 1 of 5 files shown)


awharn (approved):

packages/cmd/src/yargs/YargsConfigurer.ts
```diff
@@ -48,17 +48,8 @@ export class YargsConfigurer {
         /**
          * Add the command definitions to yargs
          */
-        const jsonResponseFormat =
```
gejohnston, line 51: What was it about this code that was failing, causing you to use a different implementation in getResponseFormat( )?

gejohnston (approved):
I am sure that you made it work. I just had a question for my own understanding.

packages/cmd/src/yargs/YargsConfigurer.ts
t1m0thyj: Previously we were checking `process.argv` when Imperative initialized to look for the `--response-format-json` argument. This worked in Zowe v1, but no longer works in Zowe v2 with daemon mode because:
* `process.argv` contains the arguments that the daemon server was launched with (`--daemon`), not the arguments passed to the current command.
* The CLI arguments need to be reloaded every time the daemon client issues a command, rather than being cached at the time Imperative initializes.

zFernand0 (approved):

PR #851, opened 2022-07-12
Diff utilities added to get the diff strings and to open the diffs in the browser view.

sarthakjdev:
This PR is related to the [PR#1460](https://github.com/zowe/zowe-cli/pull/1460) . 

Here, a utility for the compare command is added.  **Added web diff generator, web diff manager, and Diffutil functions to get the diff strings and to open up the diffs in the browser**. 
This has been done to remove the code redundancy from the CLI tool to compare the files, get the diff strings and open the same in the browser view. 
Refer [here](https://github.com/zowe/zowe-cli/pull/1460#issuecomment-1178977425)

These additions will help in almost all the issues mentioned in [here](https://github.com/zowe/zowe-cli/milestone/65)


KevinLoesch1 (approved):
Looks good to me, but a change.log entry is still needed.

packages/utilities/src/diff/doc/IWebDiffGenerator.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @interface IWebDiffGenerator
+ */
+export interface IWebDiffGenerator {
+
+    /**
+     * build the diff generator
+     */
+    buildDiffDir(): void;
+
+}
```
t1m0thyj, line 25: Creating the interface is only necessary if we use it elsewhere. Currently this one seems unused unless I have missed where we use it.

If the interface is desired, I would suggest updating the WebDiffGenerator class to implement it (`export class WebDiffGenerator implements IWebDiffGenerator`). Or if it is not needed anywhere, it may be ok to remove it?

packages/utilities/src/diff/doc/IWebDiffManager.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @interface IWebDiffManager
+ */
+export interface IWebDiffManager {
+    /**
+     * launch the diff of two file in web
+     * @param {string} content Html content string to launch in web
+     */
+    openDiffs(content: string): void;
+
+}
```
t1m0thyj, line 24: (same comment as above)

Creating the interface is only necessary if we use it elsewhere. Currently this one seems unused unless I have missed where we use it.

If the interface is desired, I would suggest updating the WebDiffManager class to implement it (`export class WebDiffManager implements IWebDiffManager`). Or if it is not needed anywhere, it may be ok to remove it?

packages/utilities/__tests__/diff/WebDiffGenerator.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import * as  fs from 'fs';
+import * as path from 'path';
+
+import WebDiffGenerator from "../../src/diff/WebDiffGenerator";
+import { ImperativeConfig } from "../../src/ImperativeConfig";
+import {IO} from '../../../io/';
+
+describe("WebDiffGenerator", () => {
+    // setting up fake cli home and web diff dir for testing
+    const cliHome: string = "packages/__tests__/fakeCliHome";
```
t1m0thyj, line 21: Instead of checking the "fakeCliHome" directory into the Git repository, should we create the directory temporarily in a `beforeAll` block before the tests run, and remove it afterwards in an `afterAll` block? This would lessen the amount of work needed to maintain this test in case the contents of the CLI home dir change in the future.

packages/utilities/src/diff/WebDiffManager.ts
```diff
@@ -0,0 +1,187 @@
...
+            />
+            <link
+              rel="stylesheet"
+              type="text/css"
+              href="https://cdn.jsdelivr.net/npm/diff2html/bundles/css/diff2html.min.css"
+            />
+            <script
+              type="text/javascript"
+              src="https://cdn.jsdelivr.net/npm/diff2html/bundles/js/diff2html-ui.min.js"
+            ></script>
```
t1m0thyj, line 148: We may want to consider bundling these CSS and JS files with Imperative, so that diffs in the browser can work offline.

I don't want to unnecessarily expand the scope of this PR and think it's a really cool feature as-is, so I'd suggest its ok to create a separate issue to track making this work offline. 🙂

t1m0thyj (changes requested):
Thanks for your contribution! Looking forward to seeing the Imperative diff utils used in Zowe CLI 🙂 

I do have a question regarding what appears to be duplicated code in this PR. It seems that the `web-diff/index.html` file is not used, and most of the `WebHelpGenerator` class which copies files over in a `web-diff` directory is also unnecessary. If the following code in `WebHelpManager`:
```
if (this.checkWebDiffDirExists()) await new WebDiffGenerator(ImperativeConfig.instance, this.webDiffDir).buildDiffDir();
```

was replaced with the following:
```
if (!this.checkWebDiffDirExists()) fs.mkdirSync(this.webDiffDir);
```

would it make it possible to delete several files (WebHelpGenerator.ts, IWebHelpGenerator.ts, index.html) from this PR?

package.json
```diff
@@ -89,10 +91,12 @@
     "yargs": "15.3.1"
   },
   "devDependencies": {
+    "@types/diff": "^5.0.2",
     "@types/find-up": "^2.1.1",
     "@types/fs-extra": "^8.0.1",
     "@types/glob": "^7.1.1",
     "@types/gulp": "^4.0.9",
+    "@types/hogan.js": "^3.0.1",
```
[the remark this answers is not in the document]
sarthakjdev: Hey @t1m0thyj , 
Thanks for reviewing the changes. 
As for now, we can absolutely remove the generator and directly create the web diff dir as suggested, but in the future, we are planning to build CSS and JS modules locally only so it will even work offline as you are even suggesting ahead. And if that will be the case we will surely need to implement the generator as it is rn. Let me know your view about this? If we want we can remove it for now too. 
And about this dependency actually whenever I remove it the ts compiler throws me an error that missing types of hogan.js that's why I kept it here.

packages/utilities/src/diff/doc/IWebDiffGenerator.ts
sarthakjdev: yeah, I will implement this interface, I missed doing so. :slightly_smiling_face:

packages/utilities/src/diff/WebDiffManager.ts
sarthakjdev: yes, that's what we thought of, but for the future scope of the feature. Will do this thing with a new issue and PR.

t1m0thyj: Sounds good 👍

package.json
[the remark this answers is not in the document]
t1m0thyj: Apologies, I had not realized the WebDiffGenerator is a work in progress and I see now that `@types/hogan.js` is required for the build to pass. I think it's fine to leave the WebDiffGenerator as-is, and can approve this PR once other comments have been addressed 🙂

packages/utilities/src/diff/WebDiffManager.ts
sarthakjdev: yes, that's what even if are thinking to do, but for the future scope of this features, as in some other issue and pr.  :slightly_smiling_face:

packages/utilities/src/diff/doc/IWebDiffManager.ts
sarthakjdev: Yes, this interface is already implemented in WebDiffManager.ts.

package.json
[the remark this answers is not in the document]
sarthakjdev: No worries, 
Will keep the web diff generator as it is then and I will improve the further required changes soon. :slightly_smiling_face:

sarthakjdev:
Done with the required changes, pls have a look .

t1m0thyj (commented):
Thanks for making the requested changes 🙂 

One more thing, I noticed a few unit tests in this PR are failing only on Linux (Ubuntu). Could you please take a look? I think the unit test for "should open the diffs in browser" may need to mock the return value of `ProcessUtils.isGuiAvailable` to pretend that there is a GUI. See this WebHelpManager test for an example of how to do so:
https://github.com/zowe/imperative/blob/64e111ae29221ed4357554c79360f1fc67577021/packages/cmd/__tests__/help/WebHelpManager.test.ts#L137

packages/utilities/src/diff/DiffUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+     * options passed into the funtions
+     * @param {string} string1
+     * @param {string} string2
+     * @param {string} options
+     * @returns {Promise<string>}
+     */
+    public static async getDiffString(string1: string, string2: string, options: IDiffOptions): Promise<string> {
+
+        if (options.outputFormat === 'terminal') {
+            let expandflag = true;
```
gejohnston, line 33: It looks like we need further test coverage.

packages/utilities/src/diff/WebDiffManager.ts
```diff
@@ -0,0 +1,187 @@
...
+            this.mInstance = new WebDiffManager();
+        }
+
+        return this.mInstance;
+    }
+
+
+    /**
+     * Launch help page for specific group/command in browser.
+     * @param {string} inContext - content of diff to be shown
```
gejohnston, line 55: The documented parameters do not match the actual parameters of the function.

packages/utilities/src/diff/WebDiffManager.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * Check if the web diff direcory base has been generated
+     * at cli home
+     */
+    private async checkWebDiffDirExists() {
+        if (fs.existsSync(this.webDiffDir)) {
+            return true;
+        }
+        return false;
```
gejohnston, line 113: Is there a reason for not creating the directory here when it does not exist?
Earlier code calls this function before trying to use webDiffDir. A few lines later it uses webDiffDir again, but this time it does not check for its existence.

sarthakjdev: Yes, I think we can remove this function and check for the condition directly itself, also few lines later it doesn't check for the existence of web diff dir again because earlier it checks for it and if it doesn't exists it creates the dir.

packages/utilities/__tests__/diff/DiffUtils.test.ts
```diff
@@ -0,0 +1,63 @@
...
+            outputFormat: 'unifiedstring'
+        };
+
+        it("should return a diff string in unified string format", async () => {
+            jest.spyOn(jestDiff , "diff").mockImplementation(jest.fn());
+            expect(await DiffUtils.getDiffString(string1, string2, diffOptions)).toMatchSnapshot();
+        });
+        it("should return a diff string in format for terminal i.e, json diff", async () => {
+            diffOptions.outputFormat = "terminal";
+            const response = await DiffUtils.getDiffString(string1, string2, diffOptions)
```
t1m0thyj, line 35: Please fix the lint warnings reported by GitHub Actions. You can also see them locally by running `npm run lint`.

sarthakjdev: Rectified the errors. :slightly_smiling_face:

t1m0thyj (approved):

gejohnston (approved):

awharn (changes requested):

awharn (changes requested):
Reviewed the code smells, requesting a few more changes

packages/utilities/src/diff/WebDiffManager.ts
```diff
@@ -0,0 +1,174 @@
...
+                    msg: errMsg
+                });
+            }
+
+            return;
+        }
+
+        if (!fs.existsSync(this.webDiffDir)) await new WebDiffGenerator(ImperativeConfig.instance, this.webDiffDir).buildDiffDir();
+
+        const htmlDiff = await html(patchDiff, {
```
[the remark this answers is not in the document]
sarthakjdev: Okay
I will correct these things.

packages/utilities/src/diff/DiffUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+            if (options.contextLinesArg >= 0) {
+                expandflag = false;
+            }
+            const jsonDiff = await diff(string1, string2, {
```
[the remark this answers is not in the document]
zFernand0: seems to be addressed now

packages/utilities/src/diff/WebDiffManager.ts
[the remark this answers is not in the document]
zFernand0: seems to be addressed now

packages/utilities/__tests__/diff/DiffUtils.test.ts
zFernand0: seems to be addressed now

zFernand0 (approved):

PR #852, opened 2022-07-13
zowe plugins enhancement

namanpatel1112:
Created new command `zowe plugins firststeps <my-plugin>`. Once a plugin is installed, a user can separately run `zowe plugins firststeps <my-plugin>` to instruct them if additional steps are needed to enable the plugin to function. If no first steps are defined for `<my-plugin>`, then the output from running the firststeps command will indicate that no additional steps are needed, and the user can start using `<my-plugin>`. First steps are defined by plugin developers in imperative.ts within the [zowe-cli-sample-plugin](https://github.com/zowe/zowe-cli-sample-plugin) repo.

Resolves https://github.com/zowe/zowe-cli/issues/1325


gejohnston (commented):
I am a little confused about the behavior. The implementation appears to implement a separate `zowe plugins firststeps` command. However, the comments before the `pluginFirstSteps?: string;` definition states the following which implies that the first steps are shown in the help.

"You can assign the first steps to the plugin that is showed in the root level help for the core CLI."

I do not object to either approach (I can see value in either one). I would find it helpful (for review purposes) if you  edited the PR's initial conversation comment to provide a thorough description of the intended behavior, maybe a sample of the command, and identify edge cases like:

- Will it show up in the help?
- Will it only be displayed when the user types a `zowe plugins firststeps` command?
- Should the firststeps command be automatically run by the `zowe plugins install` command?
- Should the `zowe plugins install` command print a message telling the user to consider running the `zowe plugins firststeps` command?
- What will the `zowe plugins firststeps` command do when a plugin does not have the pluginFirstSteps property defined?

I am not currently requesting that any of the above items be implemented. I am just asking for clarification about what the currently implemented command will do in these cases.

AmandaDErrico:
> I am a little confused about the behavior. The implementation appears to implement a separate `zowe plugins firststeps` command. However, the comments before the `pluginFirstSteps?: string;` definition states the following which implies that the first steps are shown in the help.
> 
> "You can assign the first steps to the plugin that is showed in the root level help for the core CLI."
> 
> I do not object to either approach (I can see value in either one). I would find it helpful (for review purposes) if you edited the PR's initial conversation comment to provide a thorough description of the intended behavior, maybe a sample of the command, and identify edge cases like:
> 
>     * Will it show up in the help?
> 
>     * Will it only be displayed when the user types a `zowe plugins firststeps` command?
> 
>     * Should the firststeps command be automatically run by the `zowe plugins install` command?
> 
>     * Should the `zowe plugins install` command print a message telling the user to consider running the `zowe plugins firststeps` command?
> 
>     * What will the `zowe plugins firststeps` command do when a plugin does not have the pluginFirstSteps property defined?
> 
> 
> I am not currently requesting that any of the above items be implemented. I am just asking for clarification about what the currently implemented command will do in these cases.

Thanks Gene, I've updated the PR description. I've also clarified the pluginFirstSteps documentation in IImperativeConfig.

t1m0thyj (commented):
Please fix lint warnings in `IImperativeConfig.ts`. Everything else LGTM, I'll wait to approve since we're awaiting feedback from @anaxceron on what the command name should be 🙂

anaxceron:
> Please fix lint warnings in `IImperativeConfig.ts`. Everything else LGTM, I'll wait to approve since we're awaiting feedback from @anaxceron on what the command name should be 🙂

@AmandaDErrico @t1m0thyj @gejohnston 

I think a good solution here would be  "takefirststeps." 

I think going w/ something like "getstarted" is a bit vague, whereas "takefirststeps" lets you know right away that (1) you may be dealing w/ a sequence of actions, and (2) it involves action you want to take at the initial phase of your set up, rather than something that takes place later on.

Otherwise, I was thinking "beginfirststeps" or "startfirstepts," but ultimately I thought having two words that relate to beginnings/starts/initiations came off as redundant. 

I also thought "triggerfirststeps" could work, but I wondered if it implied that the "first steps" would only happen in the back end.

I'm open to any option, but wanted to explain my preference for "takefirststeps." Hope this helps.

gejohnston:
I will suggest a slight alternative, which never occurred to me until Ana's three word suggestion of "takefirstSteps".

Since this command only displays instructions, my suggestion would be "show-first-steps". I suggest the dashes because dashes are used in most of our other multi-word parameters, so it would be consistent.

anaxceron:
> I will suggest a slight alternative, which never occurred to me until Ana's three word suggestion of "takefirstSteps".
> 
> Since this command only displays instructions, my suggestion would be "show-first-steps". I suggest the dashes because dashes are used in most of our other multi-word parameters, so it would be consistent.

@AmandaDErrico @t1m0thyj @gejohnston @zFernand0: I think this is the best option.

packages/imperative/src/doc/IImperativeConfig.ts
```diff
@@ -286,6 +286,28 @@ export interface IImperativeConfig {
...
+     * @example
+     * const config: IImperativeConfig = {
+     * ...
+     * pluginFirstSteps: "a list of first steps for your plugin"
+     * ...
+     * }
+     * @type {string}
+     * @memberof IImperativeConfig
+     */
+    pluginFirstSteps?: string;
```
zFernand0, line 309: Text input here seems ok.
I'm curious if we would eventually support a custom "handler" (similar to `pluginHealthCheck`) that gives more flexibility to plugin developers on what they can do for the users in terms of setup/configuration.

zFernand0 (approved):
Changes LGTM! 😋 

Left a question in the comments.

Other questions:
- Should we update the sample-plugin to showcase this new feature ?
- Should this be a _best practice_ for V3 conformance criteria ?

PR #854, opened 2022-07-15
Move web help dependencies into separate package.json

### LINKED ISSUE #271: Place the web-help source directory under the packages source directory

t1m0thyj:
**Question:** Do we have any concerns about using a `file:` specifier as a dev dependency?

Resolves #271

Most Imperative dependencies target Node.js, but some packages like Bootstrap and JQuery are designed for the web and we bundle them in the web help. This PR separates those web packages into a new package.json.

IMO this keeps the packages better organized - it is more obvious what the dependencies of the web help are.
Also, it makes it easier for us to audit the bundled web help dependencies, which we were not doing previously.

This PR has a "no-changelog" label because the changes are not user-facing and should only affect developers.


gejohnston (approved):
I cannot think of any problem with a file: dev dependency.

zFernand0 (approved):

PR #857, opened 2022-07-22
Fix for'Unexpected Command Error'

KevinLoesch1:
Fix for the error occuring in ImperativeExpect.toBeDefinedAndNonBlank() if a non-string is passed in. Changing Item.trim() to Item.toString().trim() corrects the problem. 

`Message:
item.trim is not a function
Stack:
TypeError: item.trim is not a function
    at Function.toBeDefinedAndNonBlank (C:\zowe\git\zowe-cli\node_modules\@zowe\imperative\lib\expect\src\ImperativeExpect.js:213:18)
    at Function.<anonymous> (C:\zowe\git\zowe-cli\packages\zosjobs\lib\GetJobs.js:106:43)
    at C:\zowe\git\zowe-cli\packages\zosjobs\lib\GetJobs.js:18:71
    at new Promise (<anonymous>)
    at __awaiter (C:\zowe\git\zowe-cli\packages\zosjobs\lib\GetJobs.js:14:12)
    at Function.getJob (C:\zowe\git\zowe-cli\packages\zosjobs\lib\GetJobs.js:104:16)
    at SpoolFileByIdHandler.<anonymous> (C:\zowe\git\zowe-cli\packages\cli\lib\zosjobs\view\spool-file-by-id\SpoolFileById.handler.js:39:63)
    at C:\zowe\git\zowe-cli\packages\cli\lib\zosjobs\view\spool-file-by-id\SpoolFileById.handler.js:18:71`


gejohnston (approved):

t1m0thyj (commented):
At first glance this PR seemed odd to me. I wondered why we are converting `item` to a string when TypeScript already declares that `item` is a string.

After seeing the context of #856, I think I understand why 🙂 It seems like the default behavior of the yargs parser is to parse any options that look like numbers as numbers ([source](https://github.com/yargs/yargs-parser#parse-numbers)). This is undesirable behavior for jobid which should be parsed as a string regardless of whether a user enters it in the format `JOB1234` or just `1234`. Is this correct?

The solution in this PR will fix the issue for this command and any command handlers that call `toBeDefinedAndNonBlank`. But I wonder if there are any other places where we may hit a similar issue if command handlers call other methods that are expecting a string and receive a number instead.

I am ok with approving this PR as-is if we believe it is the best solution. Alternatively, the behavior of yargs could be overridden (see [here](https://github.com/yargs/yargs/blob/main/docs/api.md#parserConfiguration)) to only parse options of type `number` as numbers and never parse items of type `string`. I don't know if we would consider that a bug fix, or a potentially risky breaking change?

awharn:
I agree with Timothy, perhaps it would be a good idea to disable conversion of numbers when they are specified in a string field. I don't know if that would be breaking or not, but it seems like our intended behavior. Maybe this should be a larger discussion with the team, and we introduce the solution in the PR in the interim until we make a decision?

t1m0thyj (approved):
Discussed with the team at standup. Since this is a simple fix that addresses the issue at hand, we decided to go with it for now and create an issue to document the yargs behavior 🙂

PR #865, opened 2022-08-10
Change default console log level to WARN

t1m0thyj:
This is an attempt to fix the behavior reported in https://github.com/zowe/zowe-cli/issues/511.

I believe the outcome of team discussion at standup was that we would treat this as a bug fix rather than a breaking change. See the changelog update in this PR for rationale.


gejohnston (approved):

KevinLoesch1 (approved):

zFernand0 (approved):
LGTM!
Also, thanks for cleaning up the imperative workflow 😋

awharn (approved):

PR #871, opened 2022-08-18
Improve performance of ProfileInfo API to load large team config files

t1m0thyj:
Related to zowe/vscode-extension-for-zowe#1911

**Performance Comparison**
| ProfileInfo method | Time Before (ms) | Time After (ms) |
| --- | --- | --- |
| readProfilesFromDisk | 61.0 | 44.2 |
| getDefaultProfile | 49.9 | 4.8 |
| mergeArgsForProfile | 203.4 | 12.8 |

The most significant improvement is in `mergeArgsForProfile`, since some apps like Zowe Explorer call it for every profile in the team config. The time to merge args for 100 z/OSMF profiles used to be around 20s and has been reduced by over 10x.

**Summary of Changes**
* Refactors the Config and ProfileInfo APIs so that they clone JSONC objects as little as possible. I discovered that cloning a JSONC object with `JSONC.parse(JSONC.stringify(obj))` is very slow, so it should be done sparingly except when needed to prevent accidental edits to the original object.
* Changes the parameters on some internal methods. I don't believe this is a breaking change any more than it would be for private methods, since we compile with the `stripInternal` flag. This means internal methods are only accessible by other Imperative modules and are hidden from the type definitions we publish.
* Updates the `comment-json` dependency to include a fix I made over a year ago related to comments getting lost when stringifying arrays: https://github.com/kaelzhang/node-comment-json/pull/31. The PR was merged quickly but we forgot to update to the version with the fix.
* Adds a unit test to ensure the ProfileInfo API stays reasonably performant in the future. The test fails if it takes longer than 15s to load 256 profiles. It takes under 10s on my local machine and around 12s on some GitHub Actions runners. As long as this test passes, we should be able to load 1000 profiles per minute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Improved performance of ProfileInfo API to load large team config files. [zowe/vscode-extension-for-zowe#1911](https://github.com/zowe/vscode-extension-for-zowe/issues/1911)
+
 ## `5.4.1`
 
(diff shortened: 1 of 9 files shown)


gejohnston (approved):

zFernand0 (approved):

zFernand0:
this PR will be merged with #872

PR #873, opened 2022-08-22
Revert Node lockdown in workflows

t1m0thyj:
⚠️ Do not merge until all builds pass. It seems that Ubuntu runners are still using Node 16.16 which doesn't include the fix.

Revert Node.js version lockdown now that the npm@8.11 bug has been fixed.

Also don't time the reading of profiles from disk in unit test. Since I/O performance may vary across GHA runners, this test was occasionally failing.

### DIFF
--- a/.github/workflows/imperative.yml
+++ b/.github/workflows/imperative.yml
@@ -33,7 +33,7 @@ jobs:
         fetch-depth: 0
 
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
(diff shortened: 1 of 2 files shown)


KevinLoesch1 (approved):

gejohnston (approved):

t1m0thyj:
Node.js 16.17 was released a week ago so not sure why `16.x` isn't pulling down the latest version. Created an issue here: https://github.com/actions/setup-node/issues/562

PR #876, opened 2022-08-25
Update dev deps to latest versions

t1m0thyj:
Update the following dev dependencies to maintain technical currency and reduce vulnerabilities: ESLint, Jest, Typedoc, TypeScript

* [x] Fix build errors
* [x] Fix lint errors
* [x] Fix test errors
* [x] Fix typedoc errors


package.json
```diff
@@ -143,14 +141,17 @@
...
+    "typescript": "^4.0.0",
     "uuid": "^3.3.2",
     "web-help": "file:web-help",
     "yargs-parser": "^18.1.2"
   },
+  "overrides": {
+    "glob-parent": "~5.1.2"
+  },
   "engines": {
     "node": ">=8.0.0"
```
zFernand0, line 156: Should we also update the engine to 16.0.0 ?

t1m0thyj: It would probably be good to update it 🙂 Should we use ">=14.0.0" to include all versions of Node.js that are actively supported?

zFernand0: yeah, makes sense to include 14 😋 
It should only throw warnings for people on 12 and below unless they have `engine-strict` set in their .npmrc configuration file

gejohnston (approved):
I did ask one question about spelling, which might cascade into numerous changes.

__tests__/src/packages/imperative/plugins/PluginTestContants.ts
```diff
@@ -0,0 +1,34 @@
+/*
```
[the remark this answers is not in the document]
t1m0thyj: Oops, fixed 🙂

zFernand0 (approved):
LGTM! 😋 
the code smell (`mainModule` being depracted) is fine for now.
We can always adjust in the event Node.js removes it in a future LTS version 😉

awharn (approved):

PR #877, opened 2022-08-25
Add environment variable to override plugins dir

t1m0thyj:
This is the Imperative part of the enhancement requested in https://github.com/zowe/zowe-cli/issues/1483

### DIFF
--- a/packages/imperative/src/doc/IImperativeEnvironmentalVariableSettings.ts
+++ b/packages/imperative/src/doc/IImperativeEnvironmentalVariableSettings.ts
@@ -47,4 +47,10 @@ export interface IImperativeEnvironmentalVariableSettings {
      * @type {IImperativeEnvironmentalVariableSetting}
      */
     maskOutput?: IImperativeEnvironmentalVariableSetting;
+
+    /**
+     * The directory where CLI plugins are installed.
+     * Default is `${cliHome}/plugins`.
+     */
+    pluginsDir?: IImperativeEnvironmentalVariableSetting;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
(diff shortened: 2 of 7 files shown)


packages/imperative/src/doc/IImperativeEnvironmentalVariableSettings.ts
```diff
@@ -47,4 +47,10 @@ export interface IImperativeEnvironmentalVariableSettings {
      * @type {IImperativeEnvironmentalVariableSetting}
      */
     maskOutput?: IImperativeEnvironmentalVariableSetting;
+
+    /**
+     * The directory where CLI plugins are installed.
+     * Default is `${cliHome}/plugins`.
+     */
+    pluginsDir?: IImperativeEnvironmentalVariableSetting;
```
zFernand0, line 55: Thoughts on keep using annotations like `@type` ?

It doesn't affect the generated `typedoc`, Examples:
- with types defined: [IImperativeEnvironmentalVariableSettings](https://docs.zowe.org/stable/typedoc/interfaces/_zowe_imperative.iimperativeenvironmentalvariablesettings.html)
- without types defined: [IImperativeAuthGroupConfig](https://docs.zowe.org/stable/typedoc/interfaces/_zowe_imperative.iimperativeauthgroupconfig.html)

It appears in VSCode suggestions
image

However, the type already says it's a string, and you already know the parent class/interface by hovering over the object, so `@memberOf` could also be removed.
image

zFernand0 (commented):
🥳 changes LGTM!
Left a question, but it's unrelated to the functionality 😋

KevinLoesch1 (approved):

packages/imperative/src/doc/IImperativeEnvironmentalVariableSettings.ts
gejohnston: I have never been a fan of our (often redundant) typedoc annotations. I would not suggest that we make a concentrated effort to remove them everywhere. However, if we choose to not place them in new source files, and remove them when convenient in a file that we are already working on, we could remove clutter that distracts someone who is reading the code.

gejohnston (approved):

zFernand0:
this PR will be combined with #879

PR #888, opened 2022-09-16
Fix logic to skip loading project config layers

t1m0thyj:
This is a second pass of #884 (Skip loading project config layers when project directory is an empty string). There were a few issues with the initial implementation that were discovered when testing it in Zowe Explorer:
* If project config exists in current directory, it would still be loaded even when project directory was disabled.
* Setting project directory to empty string to disable project config seemed unintuitive after further discussion. This PR changes the behavior so that project directory must be set to `false` to disable project config. This is similar to how we handle the CredentialManager override - either a string or `false`. It is not a breaking change since the previous PR has not yet been published to the `@zowe-v2-lts` tag on npmjs.org.


zFernand0 (approved):
LGTM! 😋 
Thanks for the clean implementation.
I just hope `this.cwd ?? false` doesn't feel too weird on ZE 😋

KevinLoesch1 (approved):

PR #901, opened 2022-11-14
Change the command from adduser to login

awharn:
Prepare for breaking change in NPM 9 where `npm adduser` does not login to registries anymore (replaced with `npm login`)
Note that users will need to run a separate NPM command if they are using NPM 9 to create a user on their registry.

### DIFF
--- a/packages/imperative/src/plugins/cmd/install/install.definition.ts
+++ b/packages/imperative/src/plugins/cmd/install/install.definition.ts
@@ -57,7 +57,7 @@ const loginDescription =
     "If you used this flag once for specific registry, you don't have to use it again, it uses credentials from .npmrc file.\n" +
     "\n" +
     "For more information about npm registries, see: \n" +
-    "https://docs.npmjs.com/cli/login for NPM > 9\n" +
+    "https://docs.npmjs.com/cli/login for NPM >= 9\n" +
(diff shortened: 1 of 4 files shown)


packages/imperative/src/plugins/cmd/install/install.definition.ts
```diff
@@ -53,11 +53,12 @@ const fileDescription =
...
     "The flag to add a registry user account to install from secure registry. It saves credentials " +
-    "to the .npmrc file using `npm adduser`. When this value is omitted, credentials from .npmrc file is used. " +
+    "to the .npmrc file using `npm login`. When this value is omitted, credentials from .npmrc file is used. " +
     "If you used this flag once for specific registry, you don't have to use it again, it uses credentials from .npmrc file.\n" +
     "\n" +
-    "For more information about npm registries, see: " +
-    "https://docs.npmjs.com/cli/adduser";
+    "For more information about npm registries, see: \n" +
+    "https://docs.npmjs.com/cli/login for NPM > 9\n" +
+    "https://docs.npmjs.com/cli/adduser for NPM < 9";
```
t1m0thyj, line 61: `NPM == 9` isn't included in this range. Should the first line say `NPM >= 9`?

zFernand0 (approved):

t1m0thyj (approved):

PR #909, opened 2022-11-29
Optimize slow regexes detected by SonarCloud

t1m0thyj:
When reviewing, please verify that the updated regexes should behave the same as the old ones 🙂 
* WebHelpGenerator - searches for code blocks formatted like this: `<code>$ zowe ...</code>`
* EnvQuery: searches for lines of interest in the output of `npm config list`

These changes make SonarCloud happy (fixes [S5852](https://sonarcloud.io/organizations/zowe/rules?open=typescript%3AS5852&rule_key=typescript%3AS5852)), but are not user-facing so should not require a changelog.

### DIFF
--- a/packages/cmd/src/help/WebHelpGenerator.ts
+++ b/packages/cmd/src/help/WebHelpGenerator.ts
@@ -327,7 +327,7 @@ export class WebHelpGenerator {
         htmlContent += this.renderMarkdown(markdownContent);
 
         // Add Copy buttons after command line examples
-        htmlContent = htmlContent.replace(/<code>\$\s*(.*?)<\/code>/g,
+        htmlContent = htmlContent.replace(/<code>\$\s*([^<]+)<\/code>/g,
             `<code>$1</code> <button class="btn-copy no-print" data-balloon-pos="right" data-clipboard-text="$1">Copy</button>`);
 
         // Sanitize references to user's home directory
--- a/packages/imperative/src/config/cmd/report-env/EnvQuery.ts
+++ b/packages/imperative/src/config/cmd/report-env/EnvQuery.ts
@@ -481,16 +481,9 @@ export class EnvQuery {
         npmProgress.statusMessage = "Retrieving NPM registry info";
         npmProgress.percentComplete += percentIncr;
         await EnvQuery.updateProgressBar(doesProgBarExist);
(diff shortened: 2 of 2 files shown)


KevinLoesch1 (approved):

traeok (approved):
Tested the `<code>` RegEx and its working as intended - guessing the negative set was added to prevent interpreting HTML tags within code blocks, and I verified that this works.

The other regex looks very similar to the first one, and I've tested that w/ a couple make-shift examples. Not sure if I missed any edge cases but from some quick testing everything LGTM!

gejohnston:
I misled you when we discussed a leading semicolon in the 'config report-env' output. The original regex left a leading semicolon in the output (for example the line containing "; @zowe:registry ="). The new regex does not have the semicolon.

--- original regex ---
    @broadcom:registry = "https://registry.npmjs.org/"
    ; @zowe:registry = "https://registry.npmjs.org/" ; overridden by project
    registry = "https://registry.npmjs.org/"
    "project" config from C:\ourstuff\repos\zowe-cli\.npmrc
    @zowe:registry = "https://zowe.jfrog.io/zowe/api/npm/npm-local-release/"
    node bin location = C:\pkgs\nodejs\node.exe
    cwd = C:\ourstuff\repos\zowe-cli
    HOME = C:\Users\ej608771

--- new regex ---
    @broadcom:registry = "https://registry.npmjs.org/"
    @zowe:registry = "https://registry.npmjs.org/" ; overridden by project
    registry = "https://registry.npmjs.org/"
    "project" config from C:\ourstuff\repos\zowe-cli\.npmrc
    @zowe:registry = "https://zowe.jfrog.io/zowe/api/npm/npm-local-release/"
    node bin location = C:\pkgs\nodejs\node.exe
    cwd = C:\ourstuff\repos\zowe-cli
    HOME = C:\Users\ej608771

I like the new regex output better. One could consider this a breaking change, because it alters existing output. However, I am satisfied with taking the following position:

- This is a report mostly to be viewed, not processed.
- The 'config report-env' has existed for such a short period of time that it is unlikely that anyone has written any logic to process the output. It is unlikely that anyone has even run the command yet.
- As a result, I think that we should ship this as a "non-breaking" change (or even consider it to be no change at all).

gejohnston (approved):
I left a comment with my explanation of why I approve.

PR #916, opened 2022-12-12
Fix failure to load schema when there is no profile of that type

### LINKED ISSUE #645: Failed to load schema for profile type `sample`
I create a profileInfo instance:
```
    const profInfo = new ProfileInfo("zowe", {
        requireKeytar: () => getNodeModule<NodeModule>("keytar")!
    });
```

Next attempt to read a default profile:
```
    const profAttrs = profInfo.getDefaultProfile("sample");
```

Call:
```
            const mergedArgs = profInfo.mergeArgsForProfileType(profileType, { getSecureVals: true });
```

Ensure no project config files in the current directory or global directory so that `mergeArgsForProfileType` is called against old profiles and the old schema format.  Observe an error thrown from `mergeArgsForProfileType`

```
ProfInfoErr: Failed to load schema for profile type sample
    at ProfileInfo.mergeArgsForProfile (c:\dev\node\zowe-client-sample-apps\node_modules\@zowe\imperative\packages\config\src\ProfileInfo.ts:563:19)
    at ProfileInfo.mergeArgsForProfileType (c:\dev\node\zowe-client-sample-apps\node_modules\@zowe\imperative\packages\config\src\ProfileInfo.ts:601:21)
    at Object.<anonymous> (c:\dev\node\zowe-client-sample-apps\packages\vsce\src\profiles.ts:72:41)
    at fulfilled (c:\dev\node\zowe-client-sample-apps\packages\vsce\out\profiles.js:6:58)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

t1m0thyj:
Fixes #645

### DIFF
--- a/packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
+++ b/packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
@@ -24,6 +24,7 @@ import { IProfileSchema } from "../../profiles";
 import { AbstractSession, SessConstants } from "../../rest";
 import { ConfigAutoStore } from "../src/ConfigAutoStore";
 import { ImperativeConfig } from "../../utilities/src/ImperativeConfig";
+import { ImperativeError } from "../../error";
 
 const testAppNm = "ProfInfoApp";
 const testEnvPrefix = testAppNm.toUpperCase();
@@ -277,7 +278,7 @@ describe("TeamConfig ProfileInfo tests", () => {
             const profInfo = createNewProfInfo(teamProjDir);
             await profInfo.readProfilesFromDisk();
             const desiredProfType = "tso";
-            const profAttrs = profInfo.getDefaultProfile(desiredProfType);
+            const profAttrs = profInfo.getDefaultProfile(desiredProfType) as IProfAttrs;
 
             expect(profAttrs).not.toBeNull();
             expect(profAttrs.isDefaultProfile).toBe(true);
@@ -439,7 +440,7 @@ describe("TeamConfig ProfileInfo tests", () => {
         it("should find known args in simple service profile", async () => {
             const profInfo = createNewProfInfo(teamProjDir);
             await profInfo.readProfilesFromDisk();
-            const profAttrs = profInfo.getDefaultProfile("zosmf");
+            const profAttrs = profInfo.getDefaultProfile("zosmf") as IProfAttrs;
             delete profInfo.getTeamConfig().layerActive().properties.defaults.base;
             // Since ProjectDir and HomeDir are the same (based on the ZOWE_CLI_HOME),
             // we also need to delete the base profile from that layer (even though is't just a copy)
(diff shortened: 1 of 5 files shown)


traeok (approved):

packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
```diff
@@ -277,7 +278,7 @@ describe("TeamConfig ProfileInfo tests", () => {
             const profInfo = createNewProfInfo(teamProjDir);
             await profInfo.readProfilesFromDisk();
             const desiredProfType = "tso";
-            const profAttrs = profInfo.getDefaultProfile(desiredProfType);
+            const profAttrs = profInfo.getDefaultProfile(desiredProfType) as IProfAttrs;
```
zFernand0, line 281: Just to clarify, (and we probably have discussed this already on the CLI standup) 😢 
This means that Zowe Explorer will need to do something similar when calling the `getDefaultProfile` since we are changing the function signature.

I'm afraid I can't recall were we landed on whether this is an impactful change or not 😋

packages/config/src/ProfileInfo.ts
```diff
@@ -1150,7 +1151,9 @@ export class ProfileInfo {
             }
         } else {
             // Load profile schemas from meta files in profile root dir
-            for (const { type } of this.mOldSchoolProfileCache) {
+            const profTypes = fs.existsSync(this.mOldSchoolProfileRootDir) ?
+                ProfileIO.getAllProfileDirectories(this.mOldSchoolProfileRootDir) : [];
+            for (const type of profTypes) {
```
zFernand0, line 1156: I see that this is the same technique that we use in `readProfilesFromDisk`.

I'm curious if it is necessary to do this in order to avoid a few extra `fs.existsSync` calls?
OR
if we can just add a `ensureReadFromDisk` call at the top of the `loadAllSchemas` and that way we can continue to assume that `mOldSchoolProfileCache` will have everything in it (OR will be an empty array)

🤔 It feels like calling `getAllProfileDirectories` here may not save us much.

I'd like to chat a bit about this one though 😋

zFernand0 (commented):
The code changes LGTM!
However, I'm waiting for clarification on some of my comments/questions before I approve it

packages/config/src/ProfileInfo.ts
```diff
@@ -1151,9 +1159,7 @@ export class ProfileInfo {
             }
         } else {
             // Load profile schemas from meta files in profile root dir
-            const profTypes = fs.existsSync(this.mOldSchoolProfileRootDir) ?
-                ProfileIO.getAllProfileDirectories(this.mOldSchoolProfileRootDir) : [];
-            for (const type of profTypes) {
+            for (const type of this.mOldSchoolProfileTypes) {
```
zFernand0, line 1162: Thanks for helping me understand the v1 IO operations.

packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
zFernand0: After some discussion, it seems that this could be treated as a big fix since we were already returning `null` from the `getDefaultProfile` method.

Q: Wasn't there some other method in which we could make the `profile` property optional?

zFernand0 (approved):
LGTM!
~Left a small question.~ The question only applied to ZE

packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
zFernand0: that's a ZE-only thing

PR #918, opened 2022-12-20
Kubernetes Secrets Credential Manager Integration

rudyflores:
# Integration of Kubernetes secrets as an option for Imperative when storing Zowe credentials

Allow the use of switching between storing on a Kubernetes cluster Zowe credentials as secrets or keytar on the local drive.

## How to use the new K8sCredentialManager

Inisde of `imperative.json` set your object to look like the following:
```JSON
{
  "overrides": {
    "CredentialManager": {
      "plugin": "@zowe/cli",
      "type": "k8s"
    }
  }
}
```
This will allow Imperative to know to use the k8s manager instead of the keytar manager. In order to use keytar instead use the following:
```JSON
{
  "overrides": {
    "CredentialManager": {
      "plugin": "@zowe/cli",
      "type": "keytar"
    }
  }
}
```

_Note: Old object structure was preserved, which means you are still able to utilize the old method of configuration where the override was defined as either a string or boolean_

## Testing this PR

To test this PR, as a tester you can test using the K8sCredentialManager by being logged into your Kubernetes cluster, or have your own local setup of [Minikube](https://minikube.sigs.k8s.io/docs/start/). Once being logged in, run the following steps:

1. Set your `imperative.json` as explained above and save your changes.
2. Run on Zowe CLI `zowe config init --global-config` and follow the prompts.
3. Revise on your cluster if the Kubernetes secrets were created under your namespace of your current context under your username.
4. Revise that Keytar was not used by viewing your keychain for any Zowe properties existing.


packages/security/src/K8sCredentialManager.ts
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * Retrieve data from kube config and return a {@link KubeConfig} object.
+     *
+     * @returns {k8s.CoreV1Api} the object allowing access to k8s api calls
+     *
+     * @throws {@link ImperativeError} if kube config file was not able to be opened.
+     */
+    private setupKubeConfig(): any {
```
t1m0thyj, line 187: Should we provide a means for users to define a hostname for their K8s cluster, or is it safe to assume for now that the cluster exists on the same machine where Zowe CLI is being run? I'm not a K8s expert, so please ignore this comment if it doesn't make sense 😛

t1m0thyj (approved):
Thanks for contributing the K8s credential manager 😁 
Verified that this works to store credentials in a local Minikube cluster 👍

packages/security/src/K8sCredentialManager.ts
rudyflores: The hostname does not need to be defined since the hostname is pulled from the current session when logging into the desired k8s cluster, if what you meant is the namespace, you can switch contexts in with a `kubectl config set-context --current --namespace="my-namespace-name"` and this will be the location where secrets will be stored/retrieved from without having to define it from the Zowe side.

gejohnston (changes requested):

gejohnston (approved):

traeok (approved):
LGTM! Thanks for adding this feature @rudyflores 😄

packages/security/src/K8sCredentialManager.ts
```diff
@@ -0,0 +1,229 @@
...
+        try {
+            const kc: any = new k8s.KubeConfig();
+            kc.loadFromDefault();
+
+            // Check if login was performed
+            const currentContextNamespace = kc.getContextObject(kc.getCurrentContext())?.namespace;
+            const currentUser = kc.getCurrentUser();
+            if(!currentContextNamespace || !currentUser) {
```
zFernand0, line 195: We should differentiate between not having a context and not being logged in.
Plus we should probably fallback to the cluster name instead of requiring a namespace since the `@kubernetes/client-node` package could return `undefined`.

```typescript
 // Check if login was performed
            const currentContext = kc.getContextObject(kc.getCurrentContext());
            if(!currentContext) {
                throw new Error("Current context was not found");
            }
            const currentUser = kc.getCurrentUser();
            if(!currentUser) {
                throw new Error("Current user was not found");
            }
            const k8sName = currentContext.name;
            const k8sNamespace = currentContext.namespace;
```

zFernand0 (changes requested):
I love the thorough implementation
I would like to request some changes though.

packages/security/src/K8sCredentialManager.ts
rudyflores: this sounds like a good idea, as for the `k8sName` variable, perhaps we could fallback to it if the namespace is not found, a namespace should always be required/found (even if default but not recommended) once a login is performed into a kubernetes cluster.

rudyflores:
Closing this PR after our discussion to migrate this work for a plugin instead!

PR #932, opened 2023-01-27
Fix npm not found on zowe plugins install

t1m0thyj:
Resolves zowe/zowe-cli#1615


traeok (approved):

traeok (approved):
Good call to only check if on Windows 👍

gejohnston (approved):

awharn (approved):

PR #934, opened 2023-02-02
Fix help-examples option failing on command groups

t1m0thyj:
Resolves zowe/zowe-cli#1617

### DIFF
--- a/packages/cmd/src/yargs/GroupCommandYargs.ts
+++ b/packages/cmd/src/yargs/GroupCommandYargs.ts
@@ -90,6 +90,7 @@ export class GroupCommandYargs extends AbstractCommandYargs {
                     }
                 }
                 argsForBuilder.strict();
+                CommandYargs.defineOptionsToYargs(argsForBuilder, this.definition.options);
                 return argsForBuilder;
             },
             /**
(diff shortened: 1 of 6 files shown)


KevinLoesch1 (approved):
LGTM

traeok (commented):
Do we want patch coverage for `defineOptionsToYargs`? If not, then this LGTM 🙂

awharn (approved):

t1m0thyj:
> Do we want patch coverage for the `defineOptionsToYargs` call? If not, then this LGTM 🙂

Since there are currently no tests in place for `GroupCommandYargs`, is it ok to add coverage as part of a future effort to increase coverage in Imperative?

packages/cmd/src/yargs/GroupCommandYargs.ts
```diff
@@ -90,6 +90,7 @@ export class GroupCommandYargs extends AbstractCommandYargs {
                     }
```
gejohnston, line 90: Since you are already editing this file, should you change line 88, from using the term "Bright" to "Zowe" or "Imperative"?

throw new Error(`Bright Yargs Define Error: Bright command definition type ` +
                                `${child.type} invalid.`);

gejohnston (approved):

zFernand0 (approved):

PR #943, opened 2023-02-21
Implement environment config files

awharn:
Implements the basic initial implementation of the environment configuration file.
Adds unit and integration tests for the new SDK and behavior.

This PR makes a slight modification to the Imperative initialization order. It is not believed that this change will impact any consumers of the CLI or Imperative, and all tests continue to pass.

### DIFF
--- a/packages/utilities/__tests__/EnvFileUtils.test.ts
+++ b/packages/utilities/__tests__/EnvFileUtils.test.ts
@@ -15,11 +15,15 @@ import { join } from "path";
 import { EnvFileUtils } from "../../utilities";
 
 class testError extends Error {
-    constructor (message: string) {
+    private intCode = "ENOENT";
+    constructor (message: string, code?: string) {
         super (message);
+        if (code) {
+            this.intCode = code;
+        }
     }
     get code() {
-        return "ENOENT";
+        return this.intCode;
     }
 }
 
@@ -75,7 +79,7 @@ describe("EnvFileUtils tests", () => {
         expect(error).not.toBeDefined();
     });
 
-    it("should fail to read the environment file and throw errors but close file cleanly", () => {
+    it("should fail to read the environment file and throw errors 1", () => {
         const readFileSyncSpy = jest.spyOn(fs, "readFileSync").mockReturnValueOnce(`
         {
             "TEST_VARIABLE": "TEST_VALUE_1"
@@ -96,5 +100,31 @@ describe("EnvFileUtils tests", () => {
         expect(readFileSyncSpy).toHaveBeenCalledTimes(1);
         expect(error).toBeDefined();
         expect(error.message).toContain("Failed to set up environment variables from the environment file.");
+        expect(error.message).toContain("File: ");
+        expect(error.message).toContain("Line: 4");
+        expect(error.message).toContain("Column: 12");
+    });
+
+    it("should fail to read the environment file and throw errors 2", () => {
+        const readFileSyncSpy = jest.spyOn(fs, "readFileSync").mockImplementationOnce(() => {
+            throw new testError("Test", "EPERM");
+        });
+        const setEnvironmentForAppSpy = jest.spyOn(EnvFileUtils, "setEnvironmentForApp");
+        let error;
+        try {
+            EnvFileUtils.setEnvironmentForApp("zowe");
+        } catch (err) {
+            error = err;
+        }
+        expect(setEnvironmentForAppSpy).toHaveBeenCalledTimes(1);
+        expect(setEnvironmentForAppSpy).toHaveBeenCalledWith("zowe");
+        expect(process.env.TEST_VARIABLE).not.toEqual("TEST_VALUE_1");
+        expect(process.env.ANOTHER_TEST_VARIABLE).not.toEqual("TEST_VALUE_1");
+        expect(readFileSyncSpy).toHaveBeenCalledTimes(1);
+        expect(error).toBeDefined();
+        expect(error.message).toContain("Failed to set up environment variables from the environment file.");
+        expect(error.message).toContain("File: ");
+        expect(error.message).not.toContain("Line:");
+        expect(error.message).not.toContain("Column:");
     });
 });
--- a/packages/utilities/__tests__/EnvFileUtils.test.ts
+++ b/packages/utilities/__tests__/EnvFileUtils.test.ts
@@ -1,0 +1,91 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import * as fs from "fs";
+import { homedir } from "os";
+import { join } from "path";
+import { EnvFileUtils } from "../../utilities";
+
+describe("EnvFileUtils tests", () => {
+    afterEach(() => {
+        jest.restoreAllMocks();
+        process.env.TEST_VARIABLE = "";
+        process.env.ANOTHER_TEST_VARIABLE = "";
+    });
+
+    it("should tell where a user's environment file should be for a given application", () => {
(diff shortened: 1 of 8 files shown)


packages/utilities/__tests__/EnvFileUtils.test.ts
```diff
@@ -0,0 +1,91 @@
...
+            "ANOTHER_TEST_VARIABLE": "TEST_VALUE_1"
+        };
+        `);
+        const closeSyncSpy = jest.spyOn(fs, "closeSync").mockReturnValueOnce();
+        const setEnvironmentForAppSpy = jest.spyOn(EnvFileUtils, "setEnvironmentForApp");
+        let error;
+        try {
+            EnvFileUtils.setEnvironmentForApp("zowe");
+        } catch (err) {
+            error = err;
```
traeok, line 78: If desired, we can use the `catch` block (line 77) to check the error details:

```ts
} catch (err) {
        expect(err).toBeDefined();
        expect(err.message).toContain("Failed to set up environment variables from the environment file.");
}
```

Then we won't need to define `error` beforehand - also potentially avoids Coverity and its null-pointer warnings. This really doesn't need to be changed, just figured it might avoid another Coverity complaint.

packages/utilities/src/EnvFileUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+     * Check and read in an environment file from the user home directory using the app name
+     * If the file is valid, set the environment variables
+     * If the file is not valid, display an error and continue
+     * @param {string} appName - The application name
+     * @returns {void}
+     * @throws {ImperativeError}
+     */
+    public static setEnvironmentForApp(appName: string) {
+        const expectedFileLocation = this.getEnvironmentFilePath(appName);
+        if (existsSync(expectedFileLocation)) {
```
traeok, line 34: We might be able to skip the `existsSync` check and then call `openSync` directly - I think CodeQL is mostly complaining about the time between the `existsSync` and `openSync` call.

Since we are already catching exceptions from `openSync` and we are only opening the file w/ read access, we should be able to avoid calling `existsSync` here - just an idea, might clear up the CodeQL warning.

traeok (approved):
LGTM! Thanks @awharn 

Left a comment regarding CodeQL and test code

packages/utilities/__tests__/EnvFileUtils.test.ts
awharn: Just for awareness, Coverity ignores all test files for the CLI components. I am not sure why that is not the case for the Explorer.

traeok: Oh ok, in that case you can disregard my comment as either way will accomplish the same result 👍

packages/utilities/src/EnvFileUtils.ts
awharn: The goal here was, if there is no file, then the SDK method shouldn't throw an error or attempt to do anything to the user environment because there is nothing to do. If we want the same behavior and don't want to do existsSync, then we will need to specifically check to see if the error is due to a file not being found, and ignore the error if it is that type of error, and surfacing every other type of error.

As far as I am aware, there is no error type that is able to be detected, so we would be required to check that the error code string is ENOENT, which could possibly change in the future, though that is rather unlikely. I am just a bit hesitant to be doing error detection using string comparisons.

packages/utilities/src/EnvFileUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+            try {
+                const fileHandle = openSync(expectedFileLocation, "r"); // Open file for reading
+                const fileContents = readFileSync(fileHandle).toString(); // Read the file
+                closeSync(fileHandle); // Close the file before anything goes wrong
```
t1m0thyj, line 38: Would this achieve the same thing? `const fileContents = readFileSync(expectedFileLocation).toString()`

packages/utilities/src/EnvFileUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+                const fileContentsJSON = JSON.parse(fileContents);
+                Object.keys(fileContentsJSON).forEach( key => {
+                    process.env[key] = fileContentsJSON[key];
+                });
```
t1m0thyj, line 42: Thank you for not using the spread operator to preserve case insensitivity (Path vs PATH) 😋

t1m0thyj (approved):

packages/utilities/src/EnvFileUtils.ts
awharn: It would. I just want to get the decision on existsSync out of the way before I remove it and replace it with readFileSync.

awharn: In theory I could just do a readFileSync and catch an ENOENT error.

awharn: I've gone ahead and implemented that.

packages/utilities/src/EnvFileUtils.ts
```diff
@@ -0,0 +1,58 @@
...
+        try {
+            const fileContents = readFileSync(expectedFileLocation).toString(); // Read the file in
+            const fileContentsJSON = JSON.parse(fileContents);
+            Object.keys(fileContentsJSON).forEach( key => {
+                process.env[key] = fileContentsJSON[key];
+            });
+        } catch (err) {
+            if (err.code !== "ENOENT") {
+                // Something unexpected went wrong, so return an error
+                throw new ImperativeError({msg: "Failed to set up environment variables from the environment file.\n" +
```
gejohnston, line 43: I suggest that you add the expectedFileLocation to the ImperativeError message. That might be especially important if a JSON parse error  is thrown, which probably complains about some piece of JCON text, but cannot identify what file that JSON came from.

gejohnston (changes requested):
One minor request to enhance an error message.

packages/utilities/src/EnvFileUtils.ts
awharn: I've enhanced the error message to include the full path to the file, as well as the line and column of the JSON parsing error (if applicable).

gejohnston (approved):
Thanks for going the extra mile on JSON parsing errors.

packages/utilities/src/EnvFileUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Get the expected path for the user's environment variable file
+     * @param {string} appName - The application name
+     * @param {boolean} checkCliHomeVariableFirst - Check inside of *_CLI_HOME first if it is defined
+     * @param {string} envPrefix - environment variable prefix
+     * @returns {string|undefined} - Returns the path if it exists, or undefined if it does not
+     */
+    public static getEnvironmentFilePath(appName: string, checkCliHomeVariableFirst: boolean = false, envPrefix?: string): string | undefined {
```
zFernand0, line 65: Curious if we could save some typing-definition if these methods returned `null` instead `undefined`

packages/utilities/src/EnvFileUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+     * Check and read in an environment file from the user home directory using the app name
+     * If the file is valid, set the environment variables
+     * If the file is not valid, display an error and continue
+     * @param {string} appName - The application name
+     * @param {boolean} checkCliHomeVariableFirst - Check inside of *_CLI_HOME first if it is defined
+     * @param {string} envPrefix - The environment variable prefix
+     * @returns {void}
+     * @throws {ImperativeError}
+     */
+    public static setEnvironmentForApp(appName: string, checkCliHomeVariableFirst: boolean = false, envPrefix?: string) {
```
zFernand0, line 36: should we be explicit about the return type (`... : void`)?
Also, the `:boolean` type definition might be redundant 😋

zFernand0 (approved):
My comments are very minor and I don't feel strongly about requesting changes for them.

LGTM! 😋

zFernand0 (approved):

gejohnston (approved):

t1m0thyj (approved):

traeok (approved):

Issue #945, opened 2023-02-21
Add `cause` property to ImperativeError


t1m0thyj:
Add a getter to the ImperativeError class that defines a `cause` property. It should return `this.mDetails` concatenated with `this.mAdditionalDetails`.

See MDN for details about the `cause` property on JS errors: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

LalitAswal:
#### can I work on this issue
#### if I am not wrong I just have to add
```ts
 /**
     * @type {string}
     * @memberof ImperativeError
     */
    public  get cause(): string {
    return `${this.mDetails} ${this.mAdditionalDetails}`;
  }
```
### in error/src/ImperativeError.ts file right?

LalitAswal:
> @LalitAswal Sure, feel free to 🙂 I've assigned you to the issue.
> 
> I believe we'll want to separate them with a newline character: `` return `${this.mDetails}\n${this.mAdditionalDetails}`; ``
just have doubts about this.mAdditionalDetails ( this property is not defined, should i have to define it ?)
is it  this.additionalDetails

t1m0thyj:
Good catch, `this.mDetails` is not a string and `this.mAdditionalDetails` is not defined. So I think what we really want is:
``return `${this.details}\n${this.additionalDetails}`;``

LalitAswal:
> Good catch, `this.mDetails` is not a string and `this.mAdditionalDetails` is not defined. So I think what we really want is: `` return `${this.details}\n${this.additionalDetails}`; ``

thanks :)

LalitAswal:
created PR please review it 😊

t1m0thyj:
Closing this issue since it doesn't seem possible to override the `cause` property on JS errors.

Feel free to pick up any other issues labeled as [good first issue](https://github.com/zowe/imperative/labels/good%20first%20issue) 🙂

PR #950, opened 2023-02-27
Adding cause property to ImperativeError #945

LalitAswal:
changes the function name, from cause to causeDetails

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Added `causeDetails` function to  `ImperavtiveError` class which return `details` concatenated with `additionalDetails`.
+
 ## `5.9.0`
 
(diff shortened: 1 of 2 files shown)


LalitAswal:
I have added a changelog and also removed redundant annotation from causeDetails function
please check
I am new so don't have a good idea about this, help me out if missed anything

t1m0thyj (changes requested):
Thanks for your contribution 👍

The property we want to override on ImperativeError is specifically `cause`:
image

We created the related issue (#945) because we thought that it would be possible to override this property. Now it seems like the `cause` property is restricted and cannot be overridden.

Feel free to investigate a way around this restriction. If you're able to find a way to override the `cause` property, we can proceed with this PR 🙂  Otherwise, we may have to close it and the associated issue - apologies for the confusion.

t1m0thyj:
Closing this PR since it doesn't seem possible to override the `cause` property on JS errors.

Feel free to pick up any other issues labeled as [good first issue](https://github.com/zowe/imperative/labels/good%20first%20issue) 🙂

PR #951, opened 2023-03-02
Fix environment files defect in daemon mode

awharn:
Fixes behavior where, in daemon mode, the daemon client's provided environment variables override and clobber the environment variables set from the environment file.

### DIFF
--- a/packages/cmd/__tests__/CommandProcessor.test.ts
+++ b/packages/cmd/__tests__/CommandProcessor.test.ts
@@ -24,6 +24,8 @@ import { CliUtils } from "../../utilities/src/CliUtils";
 import { WebHelpManager } from "../src/help/WebHelpManager";
 import { ImperativeConfig } from "../../utilities/src/ImperativeConfig";
 import { setupConfigToLoad } from "../../../__tests__/src/TestUtil";
+import { EnvFileUtils } from "../../utilities";
+import { join } from "path";
 
 jest.mock("../src/syntax/SyntaxValidator");
 jest.mock("../src/utils/SharedOptions");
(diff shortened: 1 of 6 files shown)


ATorrise (approved):

KevinLoesch1 (approved):

packages/cmd/__tests__/CommandProcessor.test.ts
```diff
@@ -1744,6 +1746,75 @@ describe("Command Processor", () => {
...
+            },
+            silent: true
+        };
+        EnvFileUtils.setEnvironmentForApp("zowe", false);
+        process.env.UNIT_TEST_ENV = "old";
+        try {
+            const envVarCount = Object.keys(process.env).length;
+            const commandResponse: ICommandResponse = await processor.invoke(parms);
+            expect(commandResponse).toBeDefined();
+            expect(commandResponse).toMatchSnapshot();
```
zFernand0, line 1803: I'm curious what the purpose of using snapshots here is. It looks like the command response is just an empty object?

Not requesting changes as there are other tests in this file that take snapshots for the same exact empty command response object

zFernand0 (approved):
Changes LGTM! :yum:

packages/cmd/__tests__/CommandProcessor.test.ts
awharn: I just copied it from the above test, I am not sure either. I can remove it if you prefer.

zFernand0: no need 😋

PR #952, opened 2023-03-02
being a bit clearer for how to use the ZOWE_EDITOR env variable

ATorrise:
I felt that the definition for `zowe edit config` was mildly confusing so I fleshed it out a bit more

### DIFF
--- a/packages/imperative/src/config/cmd/edit/edit.definition.ts
+++ b/packages/imperative/src/config/cmd/edit/edit.definition.ts
@@ -24,7 +24,7 @@ export const editDefinition: ICommandDefinition = {
     summary: "edit config files",
     description: `Edit an existing config file in your system's default text editor.\n\n` +
         `In a graphical environment, the application associated with JSON files will be launched. ` +
-        `In a command-line environment, vi will be launched. To override the command-line editor used, specify it in ` +
+        `In a command-line environment, vi will be launched. To override your default editor, specify the chosen program's executable location by using ` +
         `the ${ImperativeConfig.instance.loadedConfig.envVariablePrefix}_EDITOR environment variable.`,
     options: [
         {


ATorrise:
> LGTM! 😋
> 
> We could release this as a patch. That way we update snapshots on the CLI side based on this (instead of this sneaking in on the next release of imperative)

Okay - what steps do I need to take to release this as a patch?

packages/imperative/src/config/cmd/edit/edit.definition.ts
```diff
@@ -24,7 +24,7 @@ export const editDefinition: ICommandDefinition = {
     summary: "edit config files",
     description: `Edit an existing config file in your system's default text editor.\n\n` +
         `In a graphical environment, the application associated with JSON files will be launched. ` +
-        `In a command-line environment, vi will be launched. To override the command-line editor used, specify it in ` +
+        `In a command-line environment, vi will be launched. To override your default editor, specify the chosen program's executable location by using ` +
```
t1m0thyj, line 27: Thanks for revising to make the wording more clear 🙂

Instead of "the chosen program's executable location", would we want to say "the command to run the chosen program's executable"? It won't always be a full path to the executable location, for example `ZOWE_EDITOR=code` or `ZOWE_EDITOR=nano` are valid.

t1m0thyj (changes requested):
Please fix the lint warning 🙏  Also left a comment, but that's just me being picky, you don't have to address it unless you want to 😛

packages/imperative/src/config/cmd/edit/edit.definition.ts
ATorrise: can we merge both of these sentiments somehow? because yes you're right but you're also more knowledgeable than I am so that would make more sense to you than it would me

gejohnston: Just offering yet another alternative for your consideration.

Change:
    ... specify the chosen program's executable location by using ...

To:
    ... specify the command to run your chosen editor by using ...

ATorrise:
I actually think I'm going to close this branch because the feature I am working on, zowe files edit [ds, uss] <filename> --editor, will allow users to select their default editor. This will create a ZOWE_OPT_EDITOR env variable behind the scenes. The roll out of this new command and option flag may cause us to deprecate ZOWE_EDITOR and I think any changes to the default way we will enable users to set their editor should be kept in one PR.

PR #955, opened 2023-03-13
Fix plugin install error not displayed correctly

t1m0thyj:
Fixes #954

PR #958, opened 2023-03-20
Fix plugin install error not displayed correctly

zFernand0:
Same as:
- #955


packages/utilities/src/ProcessUtils.ts
```diff
@@ -80,4 +85,85 @@ export class ProcessUtils {
...
+            Logger.getImperativeLogger().info(`Opening ${filePath} in graphical editor`);
+            if (editor != null) { await require("child_process").spawn(editor, [filePath], { stdio: "inherit" }); }
+            else { this.openInDefaultApp(filePath); }
+
+        } else {
+            if (editor == null) { editor = "vi"; }
+            Logger.getImperativeLogger().info(`Opening ${filePath} in command-line editor ${editor}`);
+            await require("child_process").spawn(editor, [filePath], { stdio: "inherit" });
+        }
+    }
```
t1m0thyj, line 145: Do we want to be adding these methods to the v1-lts branch? Only the method below (`execAndCheckOutput`) is needed for the fix.

t1m0thyj (commented):
Left a comment about a number of methods that were added to ProcessUtils - not sure if we want to add any more methods than necessary, since we shouldn't be making enhancements to v1-lts.

Also, I know @awharn has identified another issue with plugin install - not sure if we want to fix that in this PR or a separate one 🙂

t1m0thyj (approved):

packages/console/src/Console.ts
```diff
@@ -16,7 +16,7 @@
  * This class supports most of the methods / accessors log4js uses.
  */
 import { IConsole } from "./doc/IConsole";
-import { TextUtils } from "../../utilities";
+import { TextUtils } from "../../utilities/src/TextUtils";
```
traeok, line 19: Did we remove `TextUtils` exports from `utilities/index.ts`? Looks like we're still doing an `export * from "TextUtils"`

packages/logger/src/Logger.ts
```diff
@@ -14,7 +14,7 @@ import { format, inspect, isNullOrUndefined } from "util";
 import { ImperativeError } from "../../error";
 import * as StackTrace from "stack-trace";
 import * as path from "path";
-import { TextUtils } from "../../utilities";
+import { TextUtils } from "../../utilities/src/TextUtils";
```
traeok, line 17: Same as previous comment (completely fine with letting this as-is, just something I noticed during review)

traeok (approved):

PR #960, opened 2023-03-22
Use cross-spawn to fix spaces in install path in v1

awharn:


t1m0thyj (approved):
LGTM, thanks @awharn!

Verified that this command works on Windows now: `zowe plugins install ".\test dir\cics-for-zowe-cli.tgz"` 👍 

The patch coverage should increase significantly if we merge #958 first 🙂

gejohnston (approved):

traeok (approved):

PR #961, opened 2023-03-22
Use cross-spawn to fix spaces in install path in v2

awharn:


t1m0thyj (approved):
LGTM, thanks @awharn!

Should we create an issue to update other places where `child_process.spawn` is used in the future? For example, `ProcessUtils.openInEditor` could fail if the file path includes a space.

gejohnston (approved):

awharn:
> LGTM, thanks @awharn!
> 
> Should we create an issue to update other places where `child_process.spawn` is used in the future? For example, `ProcessUtils.openInEditor` could fail if the file path includes a space.

It might make sense to fix that now too.

ATorrise (approved):

traeok (approved):

PR #967, opened 2023-03-30
series of changes that coincide with a recent feature of zowe cli  

ATorrise:
- enable command arguments to change `{$Prefix}_EDITOR`
- allow for terminal prompts to have text color

### DIFF
--- a/packages/cmd/src/doc/response/api/handler/IPromptOptions.ts
+++ b/packages/cmd/src/doc/response/api/handler/IPromptOptions.ts
@@ -36,4 +36,12 @@ export interface IPromptOptions {
      * @memberof IPromptOptions
      */
     maskChar?: string | null;
+
+    /**
+     * Text color to apply to your prompt
+     * ex: "yellow"
+     * @type {string}
+     * @memberof IPromptOptions
+     */
+    color?: string | null;
 }
--- a/packages/utilities/src/ProcessUtils.ts
+++ b/packages/utilities/src/ProcessUtils.ts
@@ -124,9 +124,9 @@ export class ProcessUtils {
      * the `{envVariablePrefix}_EDITOR` environment variable if specified.
      * @param filePath File path to edit
      */
-    public static async openInEditor(filePath: string) {
-        let editor;
-        if (ImperativeConfig.instance.loadedConfig.envVariablePrefix != null) {
+    public static async openInEditor(filePath: string, editorOpt?: string) {
+        let editor = editorOpt;
+        if (!editorOpt && ImperativeConfig.instance.loadedConfig.envVariablePrefix != null) {
             const editorEnvVar = `${ImperativeConfig.instance.loadedConfig.envVariablePrefix}_EDITOR`;
             if (process.env[editorEnvVar] != null) { editor = process.env[editorEnvVar]; }
         }
--- a/packages/utilities/src/CliUtils.ts
+++ b/packages/utilities/src/CliUtils.ts
@@ -592,6 +592,10 @@ export class CliUtils {
             secToWait = maxSecToWait;
         }
 
+        if (opts.color){
+            message = TextUtils.chalk[opts.color](message);
+        }
+        
         return new Promise((resolve, reject) => {
             require("read")({
                 input: process.stdin,


packages/cmd/src/doc/response/api/handler/IPromptOptions.ts
```diff
@@ -36,4 +36,12 @@ export interface IPromptOptions {
...
      */
     maskChar?: string | null;
+
+    /**
+     * Text color to apply to your prompt
+     * ex: "yellow"
+     * @type {string}
+     * @memberof IPromptOptions
+     */
+    color?: string | null;
```
t1m0thyj, line 46: This property seems to be unused, should we remove it?

packages/utilities/src/ProcessUtils.ts
```diff
@@ -126,22 +126,23 @@ export class ProcessUtils {
...
+        if (!editorOpt && ImperativeConfig.instance.loadedConfig.envVariablePrefix != null) {
             const editorEnvVar = `${ImperativeConfig.instance.loadedConfig.envVariablePrefix}_EDITOR`;
             if (process.env[editorEnvVar] != null) { editor = process.env[editorEnvVar]; }
         }
-
         if (ProcessUtils.isGuiAvailable() === GuiResult.GUI_AVAILABLE) {
             Logger.getImperativeLogger().info(`Opening ${filePath} in graphical editor`);
-            if (editor != null) { spawn.spawn(editor, [filePath], { stdio: "inherit" }); }
+            if (editor != null) {
+                sync ? spawn.sync(editor, [filePath], { stdio: "inherit" }) : spawn.spawn(editor, [filePath], { stdio: "inherit" });
```
t1m0thyj, line 138: If the same parameters are passed to both functions, this line could be simplified to:
`(sync ? spawn.sync : spawn.spawn)(editor, [filePath], { stdio: "inherit" });`

packages/utilities/src/ProcessUtils.ts
```diff
@@ -126,22 +126,23 @@ export class ProcessUtils {
...
+            }
             else { this.openInDefaultApp(filePath); }
 
         } else {
             if (editor == null) { editor = "vi"; }
             Logger.getImperativeLogger().info(`Opening ${filePath} in command-line editor ${editor}`);
-            spawn.spawn(editor, [filePath], { stdio: "inherit" });
+            sync ? spawn.sync(editor, [filePath], { stdio: "inherit" }) : spawn.spawn(editor, [filePath], { stdio: "inherit" });
```
t1m0thyj, line 145: Same comment as above - ternary operator could potentially be simplified here

packages/utilities/src/diff/doc/IDiffOptions.ts
```diff
@@ -21,7 +21,7 @@ export interface IDiffOptions {
      * @type {outputFormat}
      * @memberOf IDiffOptions
      */
-    outputFormat: outputFormat,
+    outputFormat?: outputFormat,
```
t1m0thyj, line 24: What is the expected behavior when `outputFormat` is undefined?

I think if it has a default value we should document what that is in the typedoc, and if not we should leave this property as required.

t1m0thyj (changes requested):
Thanks for enhancing the `openInEditor` method and being thorough with fixing typos 🙂 

Left a few comments, also please fix the failing unit test.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Enable command arguments to change `{$Prefix}_EDITOR`. Updating IDiffOptions
+to include names for the files that are to be compared. Changes made to support recent zowe cli work:
```
gejohnston, line 7: For consumers reading the changelog, should we change "recent zowe CLI work" to a short phrase describing the work?

packages/utilities/src/ProcessUtils.ts
```diff
@@ -126,22 +126,23 @@ export class ProcessUtils {
      * the `{envVariablePrefix}_EDITOR` environment variable if specified.
      * @param filePath File path to edit
      */
-    public static async openInEditor(filePath: string) {
-        let editor;
-        if (ImperativeConfig.instance.loadedConfig.envVariablePrefix != null) {
+    public static openInEditor(filePath: string, editorOpt?: string, sync?: boolean) {
```
gejohnston, line 129: I like that you enabled your enhancements in a non-breaking way. I think that we should add @param documentation for the new editorOpt and sync parameters.

gejohnston (changes requested):

packages/utilities/src/ProcessUtils.ts
gejohnston: In addition to the previously requested changes, I think you will have to merge master into this branch and confirm that the tests pass before you can merge this PR.

packages/utilities/src/diff/WebDiffManager.ts
```diff
@@ -155,9 +155,9 @@ export class WebDiffManager implements IWebDiffManager {
                   fileListToggle: false,
                   fileListStartVisible: false,
                   fileContentToggle: false,
-                  matching: 'lines',
+                  matching: 'none',
                   outputFormat: 'side-by-side',
-                  synchronisedScroll: true,
+                  synchronizedScroll: true,
```
awharn, line 160: I believe this needs to be `synchronisedScroll`.

awharn (changes requested):

packages/utilities/src/diff/WebDiffManager.ts
ATorrise: whoops thought i undid this

packages/utilities/src/diff/doc/IDiffOptions.ts
ATorrise: okay, default is txt

CHANGELOG.md
```diff
@@ -2,13 +2,18 @@
...
+[zowe-cli#1672](https://github.com/zowe/zowe-cli/pull/1672)
+
 ## `5.13.2`
 
 - BugFix: Reduced load time by searching for command definitions with `fast-glob` instead of `glob`.
 
 ## `5.13.1`
 
-- BugFix: Removed validation of the deprecated pluginHealthCheck property. [#980](https://github.com/zowe/imperative/issues/980)
+- BugFix: Removed validation of the deprecated pluginHealthCheck property. [#980](httpss://github.com/zowe/imperative/issues/980)
```
t1m0thyj, line 16: `httpss` -> `https`

t1m0thyj (changes requested):
Please add unit tests for the branches highlighted by Codecov as missing coverage.

Will approve once coverage checks are passing 🙂

t1m0thyj (changes requested):

packages/utilities/src/ProcessUtils.ts
```diff
@@ -124,24 +124,27 @@
...
+        if (!editorOpt && ImperativeConfig.instance.loadedConfig.envVariablePrefix != null) {
             const editorEnvVar = `${ImperativeConfig.instance.loadedConfig.envVariablePrefix}_EDITOR`;
             if (process.env[editorEnvVar] != null) { editor = process.env[editorEnvVar]; }
         }
-
         if (ProcessUtils.isGuiAvailable() === GuiResult.GUI_AVAILABLE) {
             Logger.getImperativeLogger().info(`Opening ${filePath} in graphical editor`);
-            if (editor != null) { spawn.spawn(editor, [filePath], { stdio: "inherit" }); }
+            if (editor != null) {
+                (sync ? spawn.sync : spawn.spawn)(editor, [filePath], { stdio: "inherit" });
```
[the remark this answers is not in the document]
gejohnston: Do you think that we need to do something about this code warning concerned with launching a program from an environment variable? I suspect that they are worried that some attacker puts a command like "rm -rf /*" into the editor environment variable. It is not obvious to me what we can do to confirm that the command will actually launch some editor.

gejohnston (changes requested):
I don't know if this is really a request for change, since my best suggestion to deal with the potential vulnerability of running an environment variable as a command would be to have a list of editor program names that we check for (which is not a great solution). Maybe others can identify a better action to take.

packages/utilities/src/ProcessUtils.ts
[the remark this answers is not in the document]
t1m0thyj: I believe we assessed the risk of launching programs from an `xxx_EDITOR` environment variable, and decided that it is expected behavior for this variable and consistent with how other tools (like `kubectl`) behave.

@awharn @MarkAckert I think you were involved with this discussion - could you provide more insight on this?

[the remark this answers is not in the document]
awharn: I think we decided that we are alright with specifying the command used to start the editor in the method you specified. Kubernetes have their own editor variable as well, with the same issue.  There is no good way to verify that the program that is specified is actually an editor, and unless we specifically require a particular program (at a particular location). In the worst case that it is abused, someone already has control of your user and is on your system.

t1m0thyj (approved):

t1m0thyj (changes requested):
Please fix the failing unit test 🙂

t1m0thyj (approved):
LGTM, thanks @ATorrise!

gejohnston (approved):
I left a minor comment about the title of a test. Might be nice to change, but it is not a problem.

t1m0thyj (changes requested):
Please update this file to add an export for the new interface `IDiffNameOptions`:
https://github.com/zowe/imperative/blob/master/packages/utilities/index.ts
I believe we want to have this interface exported so it can be used in Zowe CLI 😋

t1m0thyj (approved):

packages/io/__tests__/IO.test.ts
```diff
@@ -493,7 +493,7 @@ describe("IO tests", () => {
         it("should use GEdit on Linux", () => {
```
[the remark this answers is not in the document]
ATorrise: good catch!!

PR #968, opened 2023-03-31
Add --prune option to zowe config secure command

### LINKED ISSUE #547: Secure properties from non-existent files get deleted
When the `zowe config secure` command is run to save secure properties for a config file, then properties from any other config files that are non-existent are removed from the properties object stored in the vault.

Is this desired behavior? What if a config file is on a network share that sometimes is unavailable?

t1m0thyj:
Resolves #547

### DIFF
--- a/packages/imperative/src/config/cmd/secure/secure.handler.ts
+++ b/packages/imperative/src/config/cmd/secure/secure.handler.ts
@@ -75,6 +75,13 @@ export default class SecureHandler implements ICommandHandler {
             }
         }
 
+        if (params.arguments.prune) {
+            const prunedFiles = config.api.secure.rmUnusedProps();
+            if (prunedFiles.length > 0) {
+                params.response.console.log(`Deleted secure properties for ${prunedFiles.length} missing file(s)`);
+            }
+        }
+
         // Write the config layer
         await config.save();
     }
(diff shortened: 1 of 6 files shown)


packages/imperative/src/config/cmd/secure/secure.handler.ts
```diff
@@ -75,6 +75,13 @@ export default class SecureHandler implements ICommandHandler {
             }
         }
 
+        if (params.arguments.prune) {
+            const prunedFiles = config.api.secure.rmUnusedProps();
+            if (prunedFiles.length > 0) {
+                params.response.console.log(`Deleted secure properties for ${prunedFiles.length} missing file(s)`);
```
gejohnston, line 81: Since this is a destructive operation, I suggest that in addition to displaying the message to the user with the number of files that were deleted, we should also write a log message with all of the detected file names.

If we get a consumer complaint that the wrong stuff was deleted (which might a mistaken recollection of what existed) we will have a record of the files for which we deleted secure properties.

gejohnston (changes requested):
One minor suggested addition.

gejohnston (approved):

zFernand0 (approved):
LGTM! 😋

traeok (approved):
LGTM, thanks @t1m0thyj

PR #987, opened 2023-05-15
Add giveAccessOnlyToOwner function to restrict access to a file

gejohnston:
**What It Does**

The new IO.giveAccessOnlyToOwner function sets access rights on a file so that only the current user can read and write the file. It should work on both Windows and Linux-like systems.

To avoid a circular dependency from IO -> ProcessUtils -> Logger -> IO, I had to copy ProcessUtils.execAndCheckOutput to a function in a new file (ExecUtils.spawnAndGetOutput). I then had to refactor other code to use spawnAndGetOutput instead of the now deprecated execAndCheckOutput.

**How to Test**

This change is an API function. You have to write a program to test the new API. The programming is pretty simple. Add the following three lines to a program.

```
const testPermFile = "c:\\some\\temp\\directory\\perm_test.txt";
IO.createFileSync(testPermFile);
IO.giveAccessOnlyToOwner(testPermFile);
```

An `ls` command (even on Windows) will display the following:
```
ls -l
total 0
-rw------a  1 YourUserId         0 May 11 17:42 perm_test.txt
```

Additionally on Windows, the  `icacls` command will display the following:
```
icacls perm_test.txt
perm_test.txt YourDomain\YourUserId:(F)
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] added/updated automated tests
- [x] updated the changelog

**Additional Comments**

I had to leave the original ProcessUtils.execAndCheckOutput function around (Ugh!) so that we do not introduce a breaking change. I marked the function as deprecated so that it can be removed in Zowe V3.


awharn (approved):

packages/io/__tests__/IO.test.ts
```diff
@@ -17,6 +17,9 @@ import * as fs from "fs";
 import * as os from "os";
 import * as path from "path";
 import { IO } from "../../io";
+// use complete path to ExecUtils to avoid circular dependency that results from utilities/index
+import { ExecUtils } from "../../utilities/src/ExecUtils";
+import { ProcessUtils, ISystemInfo } from "../../utilities";
```
zFernand0, line 22: thanks for leaving that comment 😋

packages/utilities/src/ExecUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export class ExecUtils {
+    /**
+     * Spawn a process with arguments and throw an error if the process fails.
+     * Parameters are same as `child_process.spawnSync` (see Node.js docs).
+     * Use this method if you want the safe argument parsing of `spawnSync`
+     * combined with the smart output handling of `execSync`.
+     * @returns Contents of stdout as buffer or string
+     */
+    public static spawnAndGetOutput(command: string, args?: string[], options?: SpawnSyncOptions): Buffer | string {
```
zFernand0, line 28: Should we have the deprecated `ProcessUtils.execAndCheckOutput` function call this one?
That may avoid some duplication.

Not a huge deal though 😋

packages/io/src/IO.ts
```diff
@@ -303,6 +305,43 @@ export class IO {
...
+     * Set file access permissions so that only the current user will have access to the file.
+     * On windows, it means "full control" for the current user.
+     * On posix, it means read & write access for the current user.
+     * Obviously, the current user must have permission to change permissions
+     * on the specified file.
+     *
+     * @param  {string} fileName - file name for which we modify access.
+     * @throws An ImperativeError when the operation fails.
+     */
+    public static giveAccessOnlyToOwner(fileName: string) {
```
zFernand0, line 318: what if the file was previously an executable file?

zFernand0 (approved):
LGTM! 😋 
This is kind of a nice feature 🥳 

Question: Should we respect the executable properties of the file?
image


**Windows**
This just works 🥳 
Before: 
image

After:
image

zFernand0:
Regarding:
> Question: Should we respect the executable properties of the file?
image

We we plan to force the file to `rw`, could we rename the function to something more explicit? (e.g. `makeFileRWOnlyToOwner`)

packages/utilities/src/ExecUtils.ts
gejohnston: At first I thought that changing the automated tests would be much harder than changing the actual product, It turned out that changing the tests were pretty easy too, so I committed this change.

packages/io/src/IO.ts
gejohnston: I modified the logic to retain the executable bit, if the owner originally had it.

PR #996, opened 2023-06-26
Handle new APIML unique cookie identifier

### LINKED ISSUE #979: Handle new APIML unique cookie identifier functionality
**Description:**
As of Zowe v2.8, A user _may_ configure APIML to append a unique identifier to the token type in the cookie name (e.g., "apimlAuthenticationToken.1=[token value]"). Imperative currently expects the cookie name to be "apimlAuthenticationToken=[token value]". 

**Resolution:**
Either update client software to handle tokenType of "apimlAuthenticationToken.x" or update Imperative to ignore the unique identifier (and only set tokenType to "apimlAuthenticationToken". Alternatively, we could add a cookieUniqueID property to ISession and use that to store the text after the period if we want client sw to be able to do something with the unique identifier.

This will require some discussion.

See https://github.com/zowe/api-layer/pull/2843 for details on the changes to how APIML cookies are named.

zFernand0:
**What It Does**

- resolves #979
- added a new `add ImperativeExpect.toMatchRegExp` method
- prevented multiple `logout` operations from failing due to not having a token available 
- allowed `autoStore` for dynamic APIML token types
- ~allowed `logout` operation to remove token type and/or token value when either is not specified in the config file~
  - this cannot be done because it conflicts with the ability to logout a token different than the one stored in the vault
- added utility method: `getProfileNameFromPath` to help with nested token authentication
- added support for multiple token authentication processes in a single `config secure` command
  - added an entry to the `console.log` to inform which profile credentials are being prompted for
- prevented `auto-init` from performing two `login` operations on a single command

**How to Test**

I recommend to combine this PR with the CLI counterpart:
  - https://github.com/zowe/zowe-cli/pull/1734

Please see the above PR for testing instructions.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] added/updated automated tests
- [ ] updated the changelog

**Additional Comments**

Opened issues as a result of this PR:
 - https://github.com/zowe/imperative/issues/995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 - Enhancement: Handled unique cookie identifier in the form of dynamic token types. [#966](https://github.com/zowe/imperative/pull/996)
 - Enhancement: Added a new utility method to `ImperativeExpect` to match regular expressions. [#966](https://github.com/zowe/imperative/pull/996)
-- Enhancement: Added support for mutliple login operations in a single `config secure` command execution. [#966](https://github.com/zowe/imperative/pull/996)
+- Enhancement: Added support for multiple login operations in a single `config secure` command execution. [#966](https://github.com/zowe/imperative/pull/996)
 - BugFix: Allowed for multiple `auth logout` operations. [#966](https://github.com/zowe/imperative/pull/996)
 - BugFix: Prevented `auto-init` from sending two `login` requests to the server. [#966](https://github.com/zowe/imperative/pull/996)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Handled unique cookie identifier in the form of dynamic token types. [#966](https://github.com/zowe/imperative/pull/996)
+- Enhancement: Added a new utility method to `ImperativeExpect` to match regular expressions. [#966](https://github.com/zowe/imperative/pull/996)
+- Enhancement: Added support for mutliple login operations in a single `config secure` command execution. [#966](https://github.com/zowe/imperative/pull/996)
+- BugFix: Allowed for multiple `auth logout` operations. [#966](https://github.com/zowe/imperative/pull/996)
+- BugFix: Prevented `auto-init` from sending two `login` requests to the server. [#966](https://github.com/zowe/imperative/pull/996)
+
+
 ## `5.15.0`
 
 - Enhancement: Enabled users to display errors in a more user-friendly format with the ZOWE_V3_ERR_FORMAT environment variable. [zowe-cli#935](https://github.com/zowe/zowe-cli/issues/935)
--- a/packages/config/__tests__/Config.api.test.ts
+++ b/packages/config/__tests__/Config.api.test.ts
@@ -216,6 +216,35 @@ describe("Config API tests", () => {
                 expect(profile).toBeNull();
             });
         });
+        describe("expandPath", () => {
+            it("should expand a short proeprty path", async () => {
+                const config = await Config.load(MY_APP);
+                const profilePath = "lpar1.zosmf";
+                expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.lpar1.profiles.zosmf");
+            });
+            it.skip("should not expand a complete path", async () => {
+                const config = await Config.load(MY_APP);
+                const profilePath = "profiles.lpar1.profiles.zosmf";
+                expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.lpar1.profiles.zosmf");
+            });
+            it.skip("should expand an incomplete path", async () => {
+                const config = await Config.load(MY_APP);
+                const profilePath = "profiles.lpar1.zosmf";
+                expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.lpar1.profiles.zosmf");
+            });
+        });
+        describe("getProfileNameFromPath", () => {
+            it("should shrink profile paths", async () => {
+                const config = await Config.load(MY_APP);
+                const propertyPath = "profiles.lpar1.profiles.zosmf.properties.host";
+                expect(config.api.profiles.getProfileNameFromPath(propertyPath)).toEqual("lpar1.zosmf");
+            });
+            it("should shrink incorrect profile paths", async () => {
+                const config = await Config.load(MY_APP);
+                const propertyPath = "profiles.profiles.lpar1.profiles.profiles.zosmf.profiles.properties.host";
+                expect(config.api.profiles.getProfileNameFromPath(propertyPath)).toEqual("lpar1.zosmf");
+            });
+        });
     });
     describe("plugins", () => {
         describe("get", () => {
--- a/packages/config/__tests__/Config.api.test.ts
+++ b/packages/config/__tests__/Config.api.test.ts
@@ -222,15 +222,10 @@ describe("Config API tests", () => {
                 const profilePath = "lpar1.zosmf";
                 expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.lpar1.profiles.zosmf");
             });
-            it.skip("should not expand a complete path", async () => {
+            it("should expand a path with the keyword profiles", async () => {
                 const config = await Config.load(MY_APP);
-                const profilePath = "profiles.lpar1.profiles.zosmf";
-                expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.lpar1.profiles.zosmf");
-            });
-            it.skip("should expand an incomplete path", async () => {
-                const config = await Config.load(MY_APP);
-                const profilePath = "profiles.lpar1.zosmf";
-                expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.lpar1.profiles.zosmf");
+                const profilePath = "profiles.zosmf";
+                expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.profiles.profiles.zosmf");
             });
         });
         describe("getProfileNameFromPath", () => {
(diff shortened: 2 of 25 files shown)


packages/config/__tests__/Config.api.test.ts
```diff
@@ -216,6 +216,35 @@ describe("Config API tests", () => {
                 expect(profile).toBeNull();
             });
         });
+        describe("expandPath", () => {
+            it("should expand a short proeprty path", async () => {
+                const config = await Config.load(MY_APP);
+                const profilePath = "lpar1.zosmf";
+                expect(config.api.profiles.expandPath(profilePath)).toEqual("profiles.lpar1.profiles.zosmf");
+            });
+            it.skip("should not expand a complete path", async () => {
```
awharn, line 225: There are a couple disabled tests that should probably be re-enabled.

packages/imperative/src/config/cmd/auto-init/handlers/BaseAutoInitHandler.ts
```diff
@@ -89,6 +89,9 @@ export abstract class BaseAutoInitHandler implements ICommandHandler {
             sessCfg, params.arguments, { parms: params, doPrompting: true, serviceDescription: this.mServiceDescription },
         );
         this.mSession = new Session(sessCfgWithCreds);
+        if (this.mSession.ISession.tokenValue) {
+            this.mSession.ISession.base64EncodedAuth = this.mSession.ISession.user = this.mSession.ISession.password = this.mSession.ISession.cert = this.mSession.ISession.certKey = undefined;
```
awharn, line 93: Linter warning 😁

awharn (changes requested):

packages/config/__tests__/Config.api.test.ts
zFernand0: Since it is possible to have a `profiles.zosmf` profile, we shouldn't assume the given path is "wrong".
Same goes for the `getProfileNameFromPath` method
```bash
zFernand0 @ SSH:~/gh# zowe config list 
profiles: 
  profiles: 
    profiles: 
      zosmf: 
        properties: 
          rejectUnauthorized: false
          host:               example.com
          port:               443
          user:               USER
          password:           PASS
defaults: 
  zosmf: profiles.zosmf
autoStore: true
zFernand0 @ SSH:~/gh# zowe files list ds sys1.parmlib
SYS1.PARMLIB
SYS1.PARMLIB.D270622
SYS1.PARMLIB.POK
SYS1.PARMLIB.SHIP
```

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Handled unique cookie identifier in the form of dynamic token types. [#966](https://github.com/zowe/imperative/pull/996)
+- Enhancement: Added a new utility method to `ImperativeExpect` to match regular expressions. [#966](https://github.com/zowe/imperative/pull/996)
+- Enhancement: Added support for mutliple login operations in a single `config secure` command execution. [#966](https://github.com/zowe/imperative/pull/996)
```
t1m0thyj, line 9: `mutliple` -> `multiple`

packages/config/src/api/ConfigProfiles.ts
```diff
@@ -110,6 +110,30 @@ export class ConfigProfiles extends ConfigApi {
...
+     * Obtain the profile name (either nested or not) based on a property path.
+     *
+     * @param path The property path.
+     *
+     * @returns The corresponding profile name.
+     *
+     * @note This may be useful for supporting token authentication in a nested configuration
+     *
+     */
+    public getProfileNameFromPath(path: string): string {
```
t1m0thyj, line 124: Do you think we should deprecate `expandPath` and rename it to `getProfilePathFromName` to make both method names consistent?

packages/config/src/ConfigAutoStore.ts
```diff
@@ -106,7 +106,7 @@ export class ConfigAutoStore {
 
             if (authHandlerClass instanceof AbstractAuthHandler) {
                 const { promptParams } = authHandlerClass.getAuthHandlerApi();
-                if (profile.tokenType === promptParams.defaultTokenType) {
+                if (profile.tokenType.startsWith(promptParams.defaultTokenType)) {
```
t1m0thyj, line 109: Curious why we are replacing the equality check with `startsWith`? It makes sense for token type `apimlAuthenticationToken`, but for other token types like JWT do we still want an equality check?

I guess we're a bit limited with our options here - as we can't redefine the array of token types as regexes (`^jwtToken$`, `^apimlAuthenticationToken.*`, etc) without a breaking change.

packages/config/src/api/ConfigProfiles.ts
zFernand0: or should we name this one `shrinkPath`? 😋

packages/imperative/src/config/cmd/secure/secure.handler.ts
```diff
@@ -57,29 +57,24 @@ export default class SecureHandler implements ICommandHandler {
...
+                params.response.console.log(`Processing secure properties for profile: ${config.api.profiles.getProfileNameFromPath(propName)}`);
+                const propValue = await this.handlePromptForAuthToken(config, propName) ||
+                    await params.response.console.prompt(`Enter ${propName} - blank to skip: `, {hideText: true});
+
+                // Save the value in the config securely
+                if (propValue) {
+                    config.set(propName, propValue, { secure: true });
+                }
+            } else {
+                let propValue = await params.response.console.prompt(`Enter ${propName} - blank to skip: `, { hideText: true });
```
gejohnston, line 71: The phrase "blank to skip" has confused at least one consumer. After several failed attempts to run the secure command (due to other reasons), the consumer began to wonder if they were supposed to literally enter a blank space to skip the entry for a given secure property.

While we are making these changes, we could consider being more explicit in our prompt:

    Enter ${propName} - Press ENTER to skip

This might not be the only place that we use the phrase "blank to skip". Even so, I would choose to make an incremental (but inconsistent) improvement over maintaining the consistency of a confusing prompt.

gejohnston (approved):
I made a request related to the text used in our prompts. However, I will accept whatever judgement call you make. Thus, I am approving this PR.

packages/imperative/src/config/cmd/secure/secure.handler.ts
zFernand0: I made them all consistent 😋 
I don't believe this is a breaking change as this messages only show up in interactive mode 😋

packages/imperative/src/auth/handlers/BaseAuthHandler.ts
```diff
@@ -197,29 +196,37 @@ export abstract class BaseAuthHandler extends AbstractAuthHandler {
...
+        if (params.arguments.tokenValue != null &&
+            (params.arguments.user != null || params.arguments.password != null)) {
             params.arguments.user = undefined;
+            params.arguments.password = undefined;
         }
 
-        params.arguments.tokenType = this.mDefaultTokenType;
+        if (params.arguments.tokenType == null) {
+            params.arguments.tokenType = this.mDefaultTokenType;
+        }
```
t1m0thyj, line 208: Should we handle removing certificates from the session as well? I guess this may not be a concern, if tokens have a higher priority than certificates when both are present 😅

packages/imperative/src/config/cmd/init/init.handler.ts
```diff
@@ -186,7 +186,7 @@ export default class InitHandler implements ICommandHandler {
         }
 
         this.promptProps.push(propName);
-        const propValue: any = await this.params.response.console.prompt(`Enter ${propDesc} - blank to skip: `, { hideText: property.secure });
+        const propValue: any = await this.params.response.console.prompt(`Enter ${propDesc} - Press ENTER to skip: `, { hideText: property.secure });
```
t1m0thyj, line 189: Thanks for updating the "Press ENTER to skip" message to be consistent 🙂 Maybe we could define it as a constant somewhere? No need to change it in this PR though

awharn (approved):

packages/config/src/ConfigAutoStore.ts
```diff
@@ -80,17 +81,11 @@ export class ConfigAutoStore {
...
-        } else {
-            if (profile.basePath == null) {  // Service profiles must have basePath defined
-                return;
-            } else if (profile.tokenType == null) {  // If tokenType undefined in service profile, fall back to base profile
-                const baseProfileName = ConfigUtils.getActiveProfileName("base", opts.cmdArguments, opts.defaultBaseProfileName);
-                return this._findAuthHandlerForProfile({ ...opts, profilePath: config.api.profiles.expandPath(baseProfileName) });
-            }
+        } else if (profileType === "base" && profile.tokenType == null) {  // Base profile must have tokenType defined
+            return;
+        } else if (profile.tokenType == null) {  // If tokenType undefined in service profile, fall back to base profile
```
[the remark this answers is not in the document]
zFernand0: 🤔 not really intentional, it was more intended to prevent a confusing error with the auth-login/logout and config-secure commands.
I'll have to think that through 🤔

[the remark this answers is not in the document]
t1m0thyj: After discussing IRL, this seems ok since "basePath" may not technically be required for service profiles that connect thru the API ML.

zFernand0:
will test with ZE before moving it to ready for review again 😋

packages/config/src/ConfigAutoStore.ts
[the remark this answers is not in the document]
zFernand0: after further testing, it would seem that this could cause issues in an environment where direct-to-service profiles force a login-prompt on Zowe Explorer.

packages/imperative/src/config/cmd/init/init.handler.ts
zFernand0: moving this to a constant...

ATorrise (approved):

traeok (approved):
LGTM, thanks @zFernand0 !

PR #1001, opened 2023-07-18
ProfileInfo: Identify non-schema properties in the `mergedArgs`

ATorrise:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [X] tested my changes
- [X] added/updated automated tests
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/config/src/doc/IProfArgAttrs.ts
+++ b/packages/config/src/doc/IProfArgAttrs.ts
@@ -37,5 +37,5 @@ export interface IProfArgAttrs {
     secure?: boolean;
 
     /** Whether the argument value is defined in the schema file */
-    offSchema?: boolean;
+    inSchema?: boolean;
 }
--- a/packages/config/src/doc/IProfArgAttrs.ts
+++ b/packages/config/src/doc/IProfArgAttrs.ts
@@ -35,4 +35,7 @@ export interface IProfArgAttrs {
 
     /** Whether the argument value is stored securely */
     secure?: boolean;
+
+    /** Whether the argument value is defined in the schema file */
+    offSchema?: boolean;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- Enhancement: Added `offSchema` property for ProfileInfo to indicate if argument is a known schema argument [#899](https://github.com/zowe/imperative/issues/899)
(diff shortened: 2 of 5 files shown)


packages/config/src/doc/IProfArgAttrs.ts
```diff
@@ -35,4 +35,7 @@ export interface IProfArgAttrs {
 
     /** Whether the argument value is stored securely */
     secure?: boolean;
+
+    /** Whether the argument value is defined in the schema file */
+    offSchema?: boolean;
```
t1m0thyj, line 40: The name `offSchema` sounds weird to me - typically "off" is used in a boolean context for on/off. Would `inSchema` make sense?

t1m0thyj (commented):
This doesn't seem to be working for me locally.

Here's what I did:
1. Run `zowe config init` to create a simple config in my project directory that defines host, user, password, etc.
2. Run the following script to load the default zosmf profile and display all the loaded args:
```typescript
const { ProfileInfo } = require("@zowe/imperative");
(async () => {
    // Load connection info from default z/OSMF profile
    const profInfo = new ProfileInfo("zowe");
    const zosmfProfAttrs = profInfo.getDefaultProfile("zosmf");
    const zosmfMergedArgs = profInfo.mergeArgsForProfile(zosmfProfAttrs, { getSecureVals: true });
    console.dir(zosmfMergedArgs);
})();
```

The `offSchema` property is false for all the args. I expected it to be true for all of them since I didn't add any custom properties outside the ones defined in the schema.

packages/config/src/doc/IProfArgAttrs.ts
zFernand0: the JSDoc is backwards 😋 
The idea is to identify which properties are not defined in the schema (true), otherwise, the property `offSchema` would be `Falsy` (undefined, false, ...)

ATorrise: @t1m0thyj I agree that the name is a bit counter intuitive. @zFernand0 though the use case for this is for ZE to tell what isn't on the schema, I think it would still be alright to change the property name then search for properties not listed in the schema with `'inSchema': false` 

If that's alright with you, I'll make the change

zFernand0: sorry for the delay, and yeah, I think that makes sense 😋

traeok (approved):
I was able to confirm that the implementation works 🥳 

image

But, I agree with @t1m0thyj 's suggestion to rename the variable for clarity. Since it would just be a matter of flipping the `true` assignments to `false`, I'll approve it now, and will re-approve if my review becomes stale due to those changes 😋

gejohnston (approved):

PR #1010, opened 2023-08-28
Fix merging of profile properties in `ProfileInfo.createSession`

### LINKED ISSUE #1008: Presence of tokenValue in profile should take precedence over user and password
In the cli plugin , I am using `processWithSession` that is being passed s`ession:AbstractSession`
The session looks like this: notice **mISession**
image

in my VsCode extension I use `ProfileInfo.createSession(profile.mergedArgs.knownArgs)`; to create the session
When I hover over knownArgs:
image

and the resultant session that is created looks like this:  Notice the presence of user, password, tokenType and tokenValue

image

My profile contents 
```
"profiles": {
                "u820all": {
                    "type": "zmf",
                    "properties": {
                        "basePath": "/zmfrest",
                        "port": 9992,
                        "protocol": "https",
                        "zosmfProfile": "zosmf",
                        "tokenType": "JSESSIONID",
                        "tokenValue": "6033AA612A70926297D26E19840EC62A"
                    }
                },
```
secureArgs are not present in the system profile but instead defined in the default base profile
```
"default": {
            "type": "base",
            "properties": {
                "host": "d001.microfocus.com",
                "rejectUnauthorized": false
            },
            "secure": [
                "user",
                "password"
            ]

```
As a result, I am finding that when I issue my REST call passing session, the user and password are being used instead of the token. If I specify an invalid token this rest call still works. which would confirm that user and password are being used.

t1m0thyj:
**What It Does**
Fixes #1008 so that user and password takes precedence over tokenValue when both are defined in profile properties.

**How to Test**
1. Create a team config file that has both user/password and tokenValue defined (such as the example provided in the linked issue).
2. Call `ProfileInfo.createSession` (using a test script like the one below) and verify that only user/password are present in the session.
```javascript
const { ProfileInfo } = require("@zowe/imperative");
(async () => {
    // Load connection info from default z/OSMF profile
    const profInfo = new ProfileInfo("zowe");
    const zosmfProfAttrs = profInfo.getDefaultProfile("zosmf");
    const zosmfMergedArgs = profInfo.mergeArgsForProfile(zosmfProfAttrs, { getSecureVals: true });
    const session = ProfileInfo.createSession(zosmfMergedArgs.knownArgs);
    console.dir(session.ISession);
})();
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] added/updated automated tests
- [x] updated the changelog

**Additional Comments**
There is logic in `ConnectionPropsForSessCfg.resolveSessCfgProps` to conditionally add either user/password or tokenValue to the session. But this logic only runs when user, password, and tokenValue are passed in `cmdArgs` instead of the `sessCfg` object. The fix implemented in this PR is consistent with how Zowe CLI command handlers resolve the properties - for example the [ZosFilesBaseHandler](https://github.com/zowe/zowe-cli/blob/master/packages/cli/src/zosfiles/ZosFilesBase.handler.ts#L44-L49).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed merging of profile properties in `ProfileInfo.createSession`. [#1008](https://github.com/zowe/imperative/issues/1008)
+
 ## `5.18.0`
 
 - Enhancement: Replaced use of `node-keytar` with the new `keyring` module from `@zowe/secrets-for-zowe-sdk`. [zowe-cli#1622](https://github.com/zowe/zowe-cli/issues/1622)
--- a/packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
+++ b/packages/config/__tests__/ProfileInfo.TeamConfig.test.ts
@@ -187,6 +187,18 @@ describe("TeamConfig ProfileInfo tests", () => {
                         locType: ProfLocType.TEAM_CONFIG
                     },
                 },
+                {
+                    argName: "tokenType", dataType: "string", argValue: SessConstants.TOKEN_TYPE_JWT,
+                    argLoc: {
+                        locType: ProfLocType.TEAM_CONFIG
+                    },
+                },
+                {
+                    argName: "tokenValue", dataType: "string", argValue: "testToken",
+                    argLoc: {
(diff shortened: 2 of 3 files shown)


awharn (approved):

KevinLoesch1 (approved):

zFernand0 (approved):
changes LGTM! 😋

PR #1022, opened 2023-10-18
Fix for request steam failure

KevinLoesch1:
**What It Does**
Corrects a bug where the HTTP request does not return when downloading larger dataset or USS file to a stream. 

**How to Test**
Using the Zowe CLI, do a **'zowe files copy dsclp'** of a dataset that is larger than 16k. Without the fix it will fail with no error or warning message. With the fix applied it will complete as expected.  

**Review Checklist**
I certify that I have:
- [X ] tested my changes
- [ ] added/updated automated tests
- [X ] updated the changelog
- [ X] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fix for `AbstactRestClient` failing to return when streaming a larger file [#1805](https://github.com/zowe/zowe-cli/issues/1805)
+
 ## `5.18.2`
 
 - BugFix: Fixed normalization on stream chunk boundaries [#1815](https://github.com/zowe/zowe-cli/issues/1815)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fix for `AbstactRestClient` failing to return when streaming a larger file [#1805](https://github.com/zowe/zowe-cli/issues/1805)
```
gejohnston, line 7: Since 3 issues will be resolved, should we list all 3?

gejohnston (approved):

CHANGELOG.md
KevinLoesch1: I've updated the CHANGELOG to include all three issues.

ATorrise (approved):
👍

